PR #11, opened 2021-03-02
Automatically build API documentation

### LINKED ISSUE #10: ❓ API docs?
I was hoping to explore some API docs for the camera. Do you have those available somewhere or should I just rely on the example app for now?

mrousavy:
Automatically builds API documentation (A static website in `docs/`, will be hosted using GitHub pages) from the JSDoc comments in the TypeScript source files.

Closes #10 

## Blocker

* [ ] It doesn't pick up `export type ...`, only `export interface ...` works :(
* [ ] The default theme is ugly as hell


agrittiwari:
I was reading source code and API documentation, then read the commit/ title for this. What is it, the source code is  generating API doc for !!

PR #112, opened 2021-03-29
Remove code scanning

mrousavy:
Removes the entire barcode/QR-Code/whatever metadata code scanning. This only applies to iOS, since it wasn't implemented on Android in the first case. This is a preparation for frame processors (#2)


K-Leon:
First of - thanks for your great library!

While seeing the value in JSI Frame Processors, i'm wondering if it isn't more efficient to use the native Barcode Scan Implementation on IOS while using JSI Frame Processors for QR Detection on Android. (As far as i know the "metadata" is coming more or less 'for free'), while doing it on your own takes more performance and requires equally good algorithms.

Maybe i'm also missing a few important parts

mrousavy:
@K-Leon 

> i'm wondering if it isn't more efficient to use the native Barcode Scan Implementation on IOS while using JSI Frame Processors for QR Detection on Android.

It _might_ be more efficient to directly use the integrated barcode scan implementation from iOS regarding the algorithm alone - but all of those events have to be passed over the batched bridge at some point, so I'm not sure if that would be faster than synchronously invoking a frame processor.

Also, the "native Barcode Scan Implementation on iOS" is extremely limited in capabilities, doesn't work on Android (1), and requires additional code that is a bit messy (the `onCodeScanned` event, data marshalling through the bridge, `scannableCodes` array parsing, ...).

With Frame Processors we have a solution that can be extended through plugins by third party libraries. The performance we loose through wrapping/unwrapping data might not even be worth mentioning compared to spamming the bridge with an `onCodeScanned` event. I will have to benchmark which approach would be faster, frame processor vs iOS code scanning event. 

* (1) While a lot of codes were supported on iOS, only QR and barcode would be supported for Android (CameraX). All of the other codes here:

https://github.com/cuvent/react-native-vision-camera/blob/1558dd2f15bf33e6d342f52ef4b4179d2ad4aff0/src/Code.ts#L4-L22

wouldn't work on Android. 

I'm pretty sure frame processors are the perfect solution for all of those problems.

> As far as i know the "metadata" is coming more or less 'for free'

How so? It still has to run algorithms on the buffer internally - I don't think all of that is done on a hardware level.

K-Leon:
First of don't get me wrong! I'm extremely exited about #2 and all the work that is running into it - this is a real gamechanger. Not only for preprocessing the image, but in a broader context. (I'm watching the progress on a regular base - also your multithreading lib. It is right now one of the best places to get the hang of jsi and learn about the possibilities) Like you wrote - it is the perfect place to do the recognizing, but it is also the perfect place to offload "validating" of the scan result. (throttling, filtering, preventing sending the same barcode, giving visual feedback, on and on...)

On Android we don't need to discuss a second. I know the shitty implementations floating around and the deeply unstatisfying integration in other camera libs. (Also in CameraX....)

I'm extremely curious how this will play out. We have more or less three variants to benchmark on iOS if I haven't got anything wrong.

- Barcode recognizing in a worklet
- Feeding the "onCodeScanned" event on ios through a worklet
- The classic Bridge Implementation


> How so? It still has to run algorithms on the buffer internally - I don't think all of that is done on a hardware level.
You're absolutely right - even though i feel that it could be deeply integrated in the pipeline making it hard to beat.

If you take it that way - it is the perfect benchmark case. 

Also: If there is anything i can help you with - just let me know.

mrousavy:
@K-Leon If you happen to know how I can include C++ code from reanimated's .aar in the VisionCamera android .aar, that would solve my current blocker 😉 

Otherwise I think I can have a prototype running soon!

K-Leon:
This is really tricky.. i made a few experiments following stackoverflow suggestions without noticeable success (at least not without big changes in the other packages). But i'm sadly no c++ expert - not even remotely.

K-Leon:
I watched your awesome progress closely! It is really great work! Now I’m really curious if you managed to make a few comparisons between inbuilt barcode scanning performance and worklet powered?

mrousavy:
@K-Leon I didn't build a barcode scanner yet, but I can assure you that it is faster. I will benchmark this soon. But Frame Processors work on iOS & Android, so you can build barcode scanners yourself if you want :)

K-Leon:
(Or even more interesting utilizing built in ios barcode scanning in an worklet)

PR #113, opened 2021-03-29
Fix Audio interruptions

mrousavy:
## What

Fixes interruptions from phone calls, system alerts etc that caused the camera to completely freeze. Instead, I quickly remove the audio device so the camera is still useable, but cannot record audio during the interruption. When the interruption ends, audio will be added again.

### DIFF
--- a/ios/CameraView+AVAudioSession.swift
+++ b/ios/CameraView+AVAudioSession.swift
@@ -46,12 +46,14 @@ extension CameraView {
   /**
    Configures the CaptureSession and adds the audio device if it has not already been added yet.
    */
-  private final func addAudioInput() throws {
+  func addAudioInput() throws {
     if audioDeviceInput != nil {
       // we already added the audio device, don't add it again
       return
     }
     removeAudioInput()
+
+    ReactLogger.log(level: .info, message: "Adding audio input...")
     captureSession.beginConfiguration()
     guard let audioDevice = AVCaptureDevice.default(for: .audio) else {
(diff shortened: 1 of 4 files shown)


Futy99:
@mrousavy Are there any plans to come with proper solution instead of this workaround? Our app relies heavily on your camera library (user spends most of the time in the app recording). Now every single time when user receives phone call video is unusable because of the missing audio which is not ideal user experience. 

In case you are not planning to fix that anytime soon, is there something that could help me with our own patch implementation?

mrousavy:
Uhm, what workaround? This is the solution that is recommended by Apple - an interruption means you cannot continue using it. Once the interruption is over, everything's back to normal. I even went one step further than react-native-camera and allow you to continue to record, **just not with audio because that is not possible on iOS per design**. (they don't allow you to record phone calls)

Futy99:
Apologies I formatted this question poorly. First thing, I meant when user receives phone call notification not that he accepts the phone call. I did some research and there is `setPrefersNoInterruptionsFromSystemAlerts` function that allows to override this default behaviour (interrupting the audio session with incoming call notifications). 

If I'm not missing something we don't have this option to set preference for `setPrefersNoInterruptionsFromSystemAlerts` at the moment, right?

mrousavy:
Ah, I remember that API! Back when I worked on this a year ago I couldn't use it for some reason, but let's try again: https://github.com/mrousavy/react-native-vision-camera/pull/1278

PR #131, opened 2021-05-04
devops: Use buildcache for iOS builds

mrousavy:
from @mikehardy (https://github.com/mikehardy/buildcache-action)

### DIFF
--- a/.github/workflows/build-ios.yml
+++ b/.github/workflows/build-ios.yml
@@ -41,7 +41,7 @@ jobs:
         run: yarn install --frozen-lockfile --cwd ..
 
       - name: buildcache
-      - uses: mikehardy/buildcache-action@v1
+        uses: mikehardy/buildcache-action@v1
 
       - name: Setup Ruby (bundle)
         uses: ruby/setup-ruby@v1
--- a/.github/workflows/build-ios.yml
+++ b/.github/workflows/build-ios.yml
@@ -40,7 +40,7 @@ jobs:
       - name: Install node_modules for example/
         run: yarn install --frozen-lockfile --cwd ..
 
-      - name: buildcache
+      - name: Use buildcache
         uses: mikehardy/buildcache-action@v1
 
       - name: Setup Ruby (bundle)
--- a/.github/workflows/build-ios.yml
+++ b/.github/workflows/build-ios.yml
@@ -40,6 +40,9 @@ jobs:
       - name: Install node_modules for example/
         run: yarn install --frozen-lockfile --cwd ..
 
+      - name: buildcache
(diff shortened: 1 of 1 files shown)


mikehardy:
Hey @mrousavy :wave: :-) - you are my first integrator that isn't me so I checked this.
Looks like it's almost all cache misses right now but is working in general, i.e. buildcache is *available*
You have get Xcode to use it though, via one of the two mechanisms here that I've personally used, or via some other mechanism: https://github.com/mikehardy/buildcache-action#usage

This is one that could be done in CI only so it is zero-intrusion (alter for your project of course)
```
xcodebuild CC=clang CPLUSPLUS=clang++ LD=clang LDPLUSPLUS=clang++ -workspace demo/ios/demo.xcworkspace -scheme demo -configuration Debug -sdk iphonesimulator -derivedDataPath demo/ios/build -UseModernBuildSystem=YES
```

This is one that can be done via Podfile and affects everyone's builds of whatever you are building, but should still work even if buildcache is not installed, since clang/clang++ are put in system PATH in normal Xcode installations

```ruby
    installer.pods_project.targets.each do |target|
      target.build_configurations.each do |config|
        config.build_settings["CC"] = "clang"
        config.build_settings["LD"] = "clang"
        config.build_settings["CXX"] = "clang++"
        config.build_settings["LDPLUSPLUS"] = "clang++"
      end
    end
```

Then you should start seeing a lot more cache misses on cold cache runs, and start seeing cache hits after first run

mrousavy:
Hey @mikehardy! Thanks for the GH action. I've updated the yml to pass the compiler options, let's see if it works 😉

mrousavy:
```
Received 69272143 of 69272143 (100.0%), 76.8 MBs/sec
Cache Size: ~66 MB (69272143 B)
```
...

```
Cache status:
  Entries in cache:  988
  Cache size:        85.7 MiB (18.0%)
  Direct hits:       0
  Direct misses:     0
  Local hits:        0
  Local misses:      988
  Remote hits:       0
  Remote misses:     0
  Misses:            988
  Direct hit ratio:  0.0%
  Local hit ratio:   0.0%
  Remote hit ratio:  0.0%
  Hit ratio:         0.0%
```

looks good 🤔

mrousavy:
Before:

Screenshot 2021-05-05 at 15 27 42

After:

Screenshot 2021-05-05 at 15 27 54

mikehardy:
Excellent! That's about exactly what I expect. It's been stable somewhere right between 40% and 50% improvement on compile time, with an overhead of install/cache restore that is some constant value of around 10s total. Thanks for giving this a go with integration and let me know if you see anything strange.

Cheers!

PR #159, opened 2021-06-01
fix: Move Audio Input initialization shortly before `startRecording`

mrousavy:
## Before

Right now, opening the Camera introduces a short stutter for background music. Closing the Camera also shortly stutters. This is annoying for apps where the Camera is the main page, since users will experience music stutters everytime they check the app.

## Solution

This PR moves the audio session activation shortly before `startRecording`, so starting/stopping the Camera does not interrupt background music playback, but once the user starts to record there will be a short music stutter (not avoidable), and when he stops there will be another one. This is exactly how Snapchat does it.

## Changes

* Create separate `AVCaptureSession` which always has the audio input and output added
* Activate/Deactivate Audio Session on the fly (startRecording/stopRecording)
* Some warnings (interruptions and frame processor performance issues) will be logged to the JS console (Metro bundler). Before they were only displayed in the native Xcode console and users were very unlikely to read those.
* Added a new error union for when the audio session couldn't correctly activate
* Fix `yarn bootstrap` command
* Add `measureElapsedTime` to benchmark a few codepieces
* Only log if in DEBUG builds
* Only report slow frame processor execution times if in DEBUG builds
* Add more detailed error messages for recording errors
* Make audio optional in video recordings (aka you can now record videos while the audio device is unavailable, e.g. while in a phone call - they will just have no sound.)

## Blocker

~~I don't know how to do that, SO question: https://stackoverflow.com/q/67807394/5281431~~


mrousavy:
https://stackoverflow.com/q/67807394/5281431 :(

Issue #146, opened 2021-05-21
🐛 Selfie on Android Pixel 2 XL (OS version 11) is rotated 


jphilipstevens:
## What
Platform: Android
Tested OS version: Android 11
Devices tested: Pixel 2 XL and Pixel 3a
library version: 1.0.11

I notice using both my app and your example is that Selfie images are rotated 90 degrees counterclockwise. I checked the metadata on my app and I can see Orientation is 0 (I believe this is 0 degrees?).

at the moment I cannot move up to 2.0 as reanimaed 2,0 has debugging issues due to TurboModules.

## Reproducable sample

1. Open the example app
2. switch to selfie camera
3. capture an image

The preview of the captured image appears rotated 90 degrees counter clockwise

## Environment

* Platform(s):
  * [ ] iOS
  * [ X] Android
* react-native version: __0.64_
* react-native-vision-camera version: _1.0.11__
* react-native-reanimated version: N/A

mrousavy:
> 90 degrees counterclockwise
so it's horizontal? does it look stretched or anything?

* You said "The preview of the captured image", does this mean the actual image is not rotated, only the small thumbnail?
* Is it only the captured image itself? 
* Does the live camera Preview in the app look correct?
* Do video recordings have the correct orientation?

jphilipstevens:
Thanks for replying. Here is an example: https://imgur.com/eZZwzHd 

I used the example app to take the selfie. in my case the phone was in portrait and the couch, lamp in the house should be verticle.

- You said "The preview of the captured image", does this mean the actual image is not rotated, only the small thumbnail?
  - its the actual image
- Is it only the captured image itself?
  - I believe the image is captured in the wrong orientation
- Does the live camera Preview in the app look correct?
  -  yup, the live capture looks fine
- Do video recordings have the correct orientation?
  - I just tried the example app in this repo and the picture is flipped horizontally.

mrousavy:
> the couch, lamp in the house should be vertical.
haha, I hope so!

> and the picture is flipped horizontally
yeah that's a limitation from CameraX.

Okay so that's definitely a bug, I'll try to find out where it comes from. I wonder what rotation value the `photo` object has in this line: https://github.com/cuvent/react-native-vision-camera/blob/160ec779404be276f4953d3d760bc21d0dddb1db/android/src/main/java/com/mrousavy/camera/CameraView%2BTakePhoto.kt#L74

jphilipstevens:
[...]

PR #228, opened 2021-06-25
fix: add rotation when flipping an image

jphilipstevens:
Based on SO: https://stackoverflow.com/a/28657257/12270852

What I noticed is that the image is rotated. This worked on my pixel 2 XL. I am not sure exactly why the rotation is needed. any help would be appreciated

## What

<!--
This PR fixes an issue where selfie images appear rotated

## Changes

if you are flipping the image, add a 90-degree rotation

## Tested on

Pixel 2 XL Android 11

## Related issues

Fixes: #146

### DIFF
--- a/android/src/main/java/com/mrousavy/camera/utils/ImageProxy.save.kt
+++ b/android/src/main/java/com/mrousavy/camera/utils/ImageProxy.save.kt
@@ -45,6 +45,8 @@ fun flipImage(imageBytes: ByteArray): ByteArray {
   val bitmap = BitmapFactory.decodeByteArray(imageBytes, 0, imageBytes.size)
   val matrix = Matrix()
   matrix.preScale(-1f, 1f)
+  // Setting post rotate to 90 because image will be possibly in landscape
+  matrix.postRotate(90f);
   val newBitmap = Bitmap.createBitmap(bitmap, 0, 0, bitmap.width, bitmap.height, matrix, true)
   val stream = ByteArrayOutputStream()
   newBitmap.compress(Bitmap.CompressFormat.JPEG, 100, stream)


mrousavy:
Hey! Thanks for your PR. I'm currently not at home so I can't test the change, but it doesn't really look right to me to force-rotate it by 90 degrees as the images currently look correct on my Phone. 
With your PR they might look correct on yours, but will be incorrect on mine.
I see two solutions:
* Fully implement rotation screen rotation, there might be some adjustments needed in configureSession()
* Create a bug report at the CameraX google issue tracker, might be a bug specifically for the Pixel 2

jphilipstevens:
> Hey! Thanks for your PR. I'm currently not at home so I can't test the change, but it doesn't really look right to me to force-rotate it by 90 degrees as the images currently look correct on my Phone.
> With your PR they might look correct on yours, but will be incorrect on mine.
> I see two solutions:
> 
> * Fully implement rotation screen rotation, there might be some adjustments needed in configureSession()
> * Create a bug report at the CameraX google issue tracker, might be a bug specifically for the Pixel 2

Yup, I agree. I didn't quite know what was going on here with why the rotation was needed. I wanted to kick this or off to start hammering out a solution. I was looking at Androids docs and implemented their Orientation Listener but what I notice was orientation was 0.

My team will test out on a few devices and I'm going to try and carve out time to dig into this.

jphilipstevens:
So Something i noticed while testing:

Test case: use front-facing camera, do not run `fun flipImage`
Result: Image is mirrored as expected

Test Case: 
Results: Image is rotated 90 degrees. Below is what I changed for testing:

```
fun flipImage(imageBytes: ByteArray): ByteArray {
  val bitmap = BitmapFactory.decodeByteArray(imageBytes, 0, imageBytes.size)
  val matrix = Matrix()
  // What if we just save the image back? without any transformations?
  // matrix.preScale(-1f, 1f)
  val newBitmap = Bitmap.createBitmap(bitmap, 0, 0, bitmap.width, bitmap.height, matrix, true)
  val stream = ByteArrayOutputStream()
  newBitmap.compress(Bitmap.CompressFormat.JPEG, 100, stream)
  return stream.toByteArray()
}
```

I am not really up to speed on Android development yet, but this seems weird. Since the Constructor for a matrix creates an identity matrix i would expect the image to not be changed

coolvasanth:
Hi @jphilipstevens even I'm facing same issue on Pixel and other devices which run core android captured image is rotating. Have you found a fix for that ??

jphilipstevens:
@coolvasanth the above fix seems to work but I got here with trial and error. I have not had the chance to go deep into the fix yet.

mrousavy:
Hey, thanks for the PR but I think this has been overridden because of the Camera2 rewrite! There is a new issue to fix orientation, maybe this PR can be relevant there: #1891

PR #276, opened 2021-07-14
chore: simplify the ESLint setup

Simek:
## What

This PR simplifies the current ESLint setup in the repository. The `lint` command should yield the same results as before (19 errors thrown by `@typescript-eslint/no-unsafe-assignment` and related to `any`).

## Changes

The `@react-native-community/eslint-config` package has been updated to the latest version and most of redundant packages and related config entries have been removed.

As you can see in the `eslint-config` source, most of plugins were a direct dependency of RNC packages:
* https://github.com/facebook/react-native/blob/master/packages/eslint-config-react-native-community/package.json

Same approach applies to the configuration portion, I have removed the `plugins` or `extend` entries which were a part of other plugins loaded by RNC `eslint-config`:
* https://github.com/facebook/react-native/blob/master/packages/eslint-config-react-native-community/index.js

Hope I didn't miss anything, but feel free to point out the issue if you see any.

**EDIT:**

> You also might want to retain some of dependencies if you were using some of the newest features of the packages, because switching fully to the `eslint-config` dependencies results in small downgrades of the ESLint dependencies.

## Tested on

I have tested the changes by checking out the repository and running `yarn lint` before and after the changes.

## Related issues

None


mrousavy:
tested some simple format errors as well as some rules (missing hook dependencies) and everything looks fine - great catch @Simek, thank you!! 🖤

PR #297, opened 2021-07-23
feat: Add `console` logging support for Frame Processors

mrousavy:
## What

Until now you could not log something to the Metro JS console. The only way to log something was through the native HostFunction `_log`, which only logged to Xcode Logs/Android Logcat.
This PR changes this and adds support for using `console` to log messages to the Metro console. (e.g. `console.log`, `console.debug`, etc)

When trying to add support for this, I noticed that there were a few threading issues with the way Frame Processors are currently implemented - mainly that they use the reanimated UI Thread Scheduler for scheduling stuff, which (afaik) should only be a problem when you passed a function to a FP Plugin. But this PR fixes this by implementing a custom Scheduler.

* Adds `console` support to the VisionCamera Frame Processor runtime so you can use `console.log` inside of `useFrameProcessor`
* Creates a new Scheduler for the VisionCamera FP Thread instead of the REA UI Thread
* Android: Fixes accessing the VisionCamera FP Runtime on the wrong Thread (JS Thread instead of VisionCamera FP Thread)
* Requires a version of Reanimated which includes my fix: software-mansion/react-native-reanimated#2219

## Tested on

* iPhone 11 Pro
* Android Simulator


mrousavy:
Waiting for https://github.com/software-mansion/react-native-reanimated/pull/2219

mrousavy:
Possibly also waiting for https://github.com/software-mansion/react-native-reanimated/pull/2220

PR #308, opened 2021-07-29
feat: Expose Pixels (`planes`) to JS

mrousavy:
## What

Adds a new feature for Frame Processors: 

The `planes` property on a `Frame` object (inside `useFrameProcessor`) can be used to get the raw pixels separated in the planes straight from JS.

By using a TypedArray implementation (uint8_t) we can avoid any expensive array copies so this operation should be really cheap and exposes native arrays straight to JS. 

**This makes it possible to run any sort of algorithm using only JavaScript**, so you can write a QR code detector or a face detector or object detector or color value algorithm straight in JS without having to write any frame processor plugins.

Example:

```tsx
    'worklet'
    const planes = frame.planes
    const size = planes.length
    planes.forEach((plane) => {
      plane.pixels.forEach((pixel) => {
        console.log(`Pixel Value: ${pixel}`)
      })
    })
    // TODO: Run QR code algorithm straight from Javascript
  }, []);
```

## Challenges

* Memory management - I'm trying to figure out how to correctly clean up the buffer and release all resources after the frame processor ends, because JS doesn't immediatelly clean the variables (garbage collector) and this currently causes out of memory errors - blocked by: https://github.com/facebook/hermes/issues/564
* Android implementation - I'm not quite sure how I can expose the native buffers on Android through JNI


Kypsis:
This is a very exciting change. Too bad RN doesn't have a package that binds TFLite C API  or something to properly take advantage of this. Still a huge step towards making RN more versatile when it comes to camera.

mrousavy:
@Kypsis I agree this is a very exciting change, I hope I can finish this PR to include it in my upcoming talk at RNEU!

Also as for the Tensorflow API - you can very easily create a simple frame processor plugin for this to run any algorithm with just a few lines of code!

mrousavy:
blocked by https://github.com/facebook/hermes/issues/564

PR #357, opened 2021-08-15
fix: `Frame` has type `any` in frame processor function

yfunk:
## What
This PR fixes the parameter (`frame`) of the frame processor function in `useFrameProcessor` being `any` instead of `Frame`.

any

## Changes
Update import of `Frame` in `useFrameProcessor.ts` to be relative and be correctly resolved when using the library as a `node_module`.

## Tested on
`vision-camera-image-labeler` project in VSCode. Parameter type is `Frame` as it should be.

### DIFF
--- a/src/hooks/useFrameProcessor.ts
+++ b/src/hooks/useFrameProcessor.ts
@@ -1,7 +1,7 @@
 /* global _setGlobalConsole */
 
 import { DependencyList, useCallback } from 'react';
-import type { Frame } from 'src/Frame';
+import type { Frame } from '../Frame';
 
 type FrameProcessor = (frame: Frame) => void;
 


mrousavy:
Hey! Thanks for the PR! 🖤

Looks good, few questions before I merge:
* Do you know why it allowed me to import from `src/*` in the first place? this must be something in the tsconfig right? 
* Are there any other places where this faulty import is used? (we should know that when we fix the first point)

yfunk:
> Do you know why it allowed me to import from `src/*` in the first place? this must be something in the tsconfig right?

VSCode automatically suggests the `src/Frame` import, I think it is because`baseUrl` is `'.'` instead of `'./'`, but I can actually check later.

> Are there any other places where this faulty import is used? (we should know that when we fix the first point)

I searched the project an this is the only `'src/*'` import I found.

mrousavy:
fixed in #358

Issue #365, opened 2021-08-17
🐛 Android: build fails with: 2 files found with path 'lib/x86_64/libreactnativejni.so'


rodgomesc:
### What were you trying to do?

i'm trying to build the app for a android emulator device, but the build always fails with the error above, I figured out how to solve this, but I believe it's not the best approach. is anyone else going through this? I'm leaving it registered here if someone experiences the same problem.

 ###  workaround
  
 * open `android/app/build.gradle`

 ```js 
 
 android {
    packagingOptions {   \\ <- add this
        pickFirst '**/*.so'
    }
    
  ```

### What happened instead?

Build fails with: 2 files found with path 'lib/x86_64/libreactnativejni.so'

### Relevant log output

```shell
example git:(master) ✗ yarn android
yarn run v1.22.11
$ react-native run-android
info Running jetifier to migrate libraries to AndroidX. You can disable it using "--no-jetifier" flag.
Jetifier found 937 file(s) to forward-jetify. Using 12 workers...
info Starting JS server...
info Installing the app...
Starting a Gradle Daemon (subsequent builds will be faster)

> Configure project :react-native-vision-camera
Warning: The 'kotlin-android-extensions' Gradle plugin is deprecated. Please use this migration guide (https://goo.gle/kotlin-android-extensions-deprecation) to start working with View Binding (https://developer.android.com/topic/libraries/view-binding) and the 'kotlin-parcelize' plugin.
WARNING:: Mapping new ns http://schemas.android.com/repository/android/common/02 to old ns http://schemas.android.com/repository/android/common/01
Mapping new ns http://schemas.android.com/repository/android/common/02 to old ns http://schemas.android.com/repository/android/common/01
WARNING:: Mapping new ns http://schemas.android.com/repository/android/generic/02 to old ns http://schemas.android.com/repository/android/generic/01
Mapping new ns http://schemas.android.com/repository/android/generic/02 to old ns http://schemas.android.com/repository/android/generic/01
WARNING:: Mapping new ns http://schemas.android.com/sdk/android/repo/addon2/02 to old ns http://schemas.android.com/sdk/android/repo/addon2/01
Mapping new ns http://schemas.android.com/sdk/android/repo/addon2/02 to old ns http://schemas.android.com/sdk/android/repo/addon2/01
WARNING:: Mapping new ns http://schemas.android.com/sdk/android/repo/repository2/02 to old ns http://schemas.android.com/sdk/android/repo/repository2/01
[...]

PR #370, opened 2021-08-19
fix: Exclude META-INF from packaging options

### LINKED ISSUE #369: 🐛 Build fails for Android using Detox Commands

Tied to build the 2.5.0 build through Detox command using:

`./gradlew assembleReleaseE2E assembleAndroidTest -DtestBuildType=releaseE2E`

### Reproduceable Code

_No response_

### What happened instead?
```
> Task :react-native-vision-camera:mergeDebugAndroidTestJavaResource FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':react-native-vision-camera:mergeDebugAndroidTestJavaResource'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.Workers$ActionFacade
   > More than one file was found with OS independent path 'META-INF/MANIFEST.MF'.
```

### Relevant log output

```shell
See above
```

### Device

Default XCode CircleCI Machine

### VisionCamera Version

2.5.0

### Additional information

- [ ] I am using Expo

ribeirogabriel:
## Changes

Excluded META-INF files to avoid conflict issues

## Tested on

Android Studio 4.2.0 MacOS 11.4

## Related issues

* Fixes #369

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -112,6 +112,7 @@ android {
 
   packagingOptions {
     excludes = ["**/libc++_shared.so"]
+    exclude "META-INF/**"
   }
 
   buildTypes {
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -112,6 +112,7 @@ android {
 
   packagingOptions {
     excludes = ["**/libc++_shared.so"]
+    exclude "META-INF/*"
   }
 
   buildTypes {
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -113,6 +113,7 @@ android {
   packagingOptions {
     excludes = ["**/libc++_shared.so"]
     exclude "META-INF/**"
+    pickFirst '**/*.so'
   }
 
   buildTypes {


android/build.gradle
```diff
@@ -112,6 +112,8 @@ android {
 
   packagingOptions {
     excludes = ["**/libc++_shared.so"]
+    exclude "META-INF/**"
+    pickFirst '**/*.so'
```
mrousavy, line 116: why `**/*.so`? why not just the `META-INF/**`?

mrousavy (changes requested):

mrousavy:
thanks for your PR! Left a comment there because I don't see how that's related.. also, could you tell me how you can reproduce the build error? I've never seen it before and I am not aware of any side-effects of ignoring META-INF...

android/build.gradle
ribeirogabriel: I've added this mostly because if there are some other .so files they will conflict as well (in this case when you're integrating OpenCV for another frame processor plugin.

ribeirogabriel: > thanks for your PR! Left a comment there because I don't see how that's related.. also, could you tell me how you can reproduce the build error? I've never seen it before and I am not aware of any side-effects of ignoring META-INF...

It happened to me a couple of times when using a combination of Hermes + Detox + Vision Camera Library on CircleCI, it might not be directly related to this plugin itself (more of how gradle ended up bundling all those packages together), but it ended up working nicely 🤔 

It's more of a workaround since I wasn't able to compile successfully applying this to my `app/build.gradle`.

rodgomesc, line 116: line 116 is related to the same problem that  i found here, https://github.com/mrousavy/react-native-vision-camera/issues/365

`pickFirst` seems to be a workaround for that but not a clear solution since the build fails with duplicated `libreactnativejni.so` files

mrousavy:
Hey! Thank you for your PR! I have added this change in the latest update!

PR #412, opened 2021-09-07
feat: Make Reanimated optional (disable Frame Processors if REA v2 is not installed)

### LINKED ISSUE #411: ❓ Is react-native-reanimated now required?
### Question

Is react-native-reanimated now required?

### What I tried

In 2.4.1, I could build my iOS and Android app without having react-native-reanimated installed.

However, when I upgraded to 2.7.0 just now, both my iOS and Android builds failed due to a) a missing RNReanimated header for iOS, and b) a missing ":react-native-reanimated" project for Android.

According to the docs, the Reanimated library is only needed for frame processors, so I thought it was optional. 😢

EDIT: In fact, it seems like 2.5.0 is the first version that causes this build error. FYI I'm still running RN 0.64.1

EDIT2: I tried disabling frame processors in iOS by adding `"VISION_CAMERA_DISABLE_FRAME_PROCESSORS=1",` to my project.pbxproj, but the build error persists. (Not sure how to disable frame processors in Android...)

### VisionCamera Version

2.7.0

### Additional information

- [ ] I am using Expo

### LINKED ISSUE #660: 🐛 Documentation doesn't mention that Reanimated v2 is required

[Getting Started guide](https://mrousavy.com/react-native-vision-camera/docs/guides) hints that Reanimated v2 would only be required when Frame Processors are used. However, Reanimated v2 seems to be always required even if Frame Processors have been disabled.

image

### What happened instead?

When building the app on iOS without Reanimated v2, the building process fails with the error:

> 'RNReanimated/Scheduler.h' file not found

By looking at the issues & PRs, it seems like the Reanimated v2 is always required:  https://github.com/mrousavy/react-native-vision-camera/issues/411  https://github.com/mrousavy/react-native-vision-camera/pull/412

If that's the case, it would be great if the Getting Started page and the Troubleshoot page mentioned this.

### Device

iOS, Android

### VisionCamera Version

2.1.0

### Additional information

- [ ] I am using Expo

mrousavy:
## What

This PR adds support for using VisionCamera without Reanimated v2. A lot of people didn't migrate to v2 yet, but they should also be able to use VisionCamera - they just can't use Frame Processors.

Implemented:

* [x] iOS
* [ ] Android

## Blockers

I don't really know how to make a dependency/library optional in CMake/build.gradle....

## Related issues

* Fixes https://github.com/mrousavy/react-native-vision-camera/issues/411
* Fixes #660

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -85,7 +85,7 @@ def getExtOrIntegerDefault(name) {
 }
 
 def reanimated = rootProject.subprojects.find { it.name == 'react-native-reanimated' }
-def ENABLE_FRAME_PROCESSORS = getExtOrDefault("enableFrameProcessors") && reanimated != null
+def ENABLE_FRAME_PROCESSORS = !getExtOrDefault("disableFrameProcessors") && reanimated != null
 
 android {
   compileSdkVersion getExtOrIntegerDefault('compileSdkVersion')
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -96,15 +96,16 @@ android {
     minSdkVersion 21
     targetSdkVersion getExtOrIntegerDefault('targetSdkVersion')
 
-    externalNativeBuild {
-      cmake {
-        cppFlags "-fexceptions", "-frtti", "-std=c++1y", "-DONANDROID"
-        abiFilters 'x86', 'x86_64', 'armeabi-v7a', 'arm64-v8a'
-        arguments '-DANDROID_STL=c++_shared',
-                "-DREACT_NATIVE_VERSION=${REACT_NATIVE_VERSION}",
-                "-DENABLE_FRAME_PROCESSORS=${ENABLE_FRAME_PROCESSORS}",
-                "-DNODE_MODULES_DIR=${rootDir}/../node_modules",
-                "-DFOR_HERMES=${FOR_HERMES}"
+    if (ENABLE_FRAME_PROCESSORS) {
+      externalNativeBuild {
+        cmake {
+          cppFlags "-fexceptions", "-frtti", "-std=c++1y", "-DONANDROID"
+          abiFilters 'x86', 'x86_64', 'armeabi-v7a', 'arm64-v8a'
+          arguments '-DANDROID_STL=c++_shared',
+                  "-DREACT_NATIVE_VERSION=${REACT_NATIVE_VERSION}",
+                  "-DNODE_MODULES_DIR=${rootDir}/../node_modules",
+                  "-DFOR_HERMES=${FOR_HERMES}"
+        }
       }
     }
   }
@@ -113,9 +114,11 @@ android {
     javaMaxHeapSize "4g"
   }
 
-  externalNativeBuild {
-    cmake {
-      path "CMakeLists.txt"
+  if (ENABLE_FRAME_PROCESSORS) {
+    externalNativeBuild {
+      cmake {
+        path "CMakeLists.txt"
+      }
     }
   }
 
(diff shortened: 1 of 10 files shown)


jzxchiang1:
This is awesome!

Is anything here relevant: https://docs.gradle.org/current/userguide/feature_variants.html?

jzxchiang1:
Any luck with this?

mrousavy:
Not yet, don't understand Gradle well enough yet.

jzxchiang1:
Thanks for the update. I'll take a look as well.

mrousavy:
@jzxchiang1 did you find anything? I need two things:

1. Make this optional: https://github.com/mrousavy/react-native-vision-camera/blob/8d24e344c9252a6a500e5f63a39539e0ff24fa11/android/build.gradle#L212
    when not found, set a variable in the `build.gradle` (e.g. `REANIMATED_AVAILABLE = false`), and then optionally remove the Java code. This seems problematic to me, because the Java code actually contains an Import from `com.swmansion.reanimated.xxx`.
2. Make this optional: https://github.com/mrousavy/react-native-vision-camera/blob/8d24e344c9252a6a500e5f63a39539e0ff24fa11/android/build.gradle#L228
    when not found, set a variable in the `build.gradle` (e.g. `REANIMATED_AVAILABLE = false`) which I can then pass to `externalNativeBuild` (to be read and used in `CMakeLists.txt`)

Any help appreciated. I can take it from there.

mrousavy:
This _might_ be helpful: https://stackoverflow.com/a/16823279

jzxchiang1:
The good news is that react-native-reanimated will actually support v8 quite soon. I just got it working after a lot of help from the maintainer of react-native-v8, who was awesome enough to share a couple of patches.

So I think this might not be urgent anymore, now that react-native-reanimated should support JSC, Hermes, and v8.

angelos3lex:
@mrousavy Just in case we can't yet update to reanimated-v2 (in our case because we are using `react-native-draggable-flatlist` which is not yet compatible with v2 (https://github.com/computerjazz/react-native-draggable-flatlist/pull/335)), and we need the `react-native-vision-camera` but without frame-processors, Is there any workaround?

mrousavy:
the workaround is to delete all of the C++ files and CMake setup 😅

angelos3lex:
Do you think that making frame processors a separate repo from the vision camera and have it as an add-on (which would entirely need animated-V2), would be a solution?

mrousavy:
No, because that is way more complex than simply allowing Frame Processors to be disabled if REA is not installed

nelsonprsousa:
> @mrousavy Just in case we can't yet update to reanimated-v2 (in our case because we are using `react-native-draggable-flatlist` which is not yet compatible with v2 ([computerjazz/react-native-draggable-flatlist#335](https://github.com/computerjazz/react-native-draggable-flatlist/pull/335))), and we need the `react-native-vision-camera` but without frame-processors, Is there any workaround?

I believe react-native-draggable-flatlist is compatible with REA v2. It just doesn't uses the new REA v2 API, but the old one still works with V2.

@computerjazz knows better

computerjazz:
@nelsonprsousa yep that's correct. react-native-draggable-flatlist uses the v1 api because performance is better with v1 when rendering an arbitrarily-long list where each item has complex animated value interactions (largely thanks to the `proc` node). This is only an implementation detail, the package works just fine with reanimated 2, and I actually have reanimated 2 as a peer dep just to allow me to transition to the new api whenever the perf issues are fixed.

Would love to get your take on this @mrousavy but I won't hijack this thread :). I know some performance fixes have landed in reanimated recently but last I checked the v2 api still resulted in low fps.

mrousavy:
@jzxchiang1 can you test if it works for you now?

mrousavy:
It works! Just tested both on Android and iOS, Reanimated is no longer required! And if you don't use Frame Processors, compile time is faster! :)

jzxchiang1:
Sorry just saw this. Thank you! I ended up getting reanimated to work either way, but this is helpful as well.

PR #415, opened 2021-09-08
Avoid division by zero

icc-guerrero:
## What

    This PR fixes a bug in FrameProcessor (iOS). Variable averageExecutionTimeSeconds is wrongly initialised if the collector is run before obtaining any performanceSample.

## Tested on

    * iPhone 8


mrousavy:
Hey! Thanks for your PR! 🖤 
Instead of providing false data (1), how about we avoid running the performance sample collection alltogether? I patched together a quick PR here: #416, does that work for you?

PR #416, opened 2021-09-08
fix: Fix divison by zero in Performance Sample collector

mrousavy:


icc-guerrero:
Works also fine for me and looks better 👍 
Thanks for the prompt action!

PR #419, opened 2021-09-11
refactor: suppress AppState.removeListener warning by deprecation

developerantoniosousa:
## What
Suppress AppState.removeListener warning by deprecation

## Changes
The way of removing a subscription has changed for AppState event listeners. Basically, it's an update for using a new way to handle unsubscription. AppState.AddEventListener returns a remove function subscription, which replaces the removeListener method

## Tested on
Samsung A10, Android 11

## Related issues
I couldn't find one.

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -35,7 +35,7 @@
     "@react-native-community/eslint-config": "^3.0.0",
     "@react-native-community/eslint-plugin": "^1.1.0",
     "@types/react": "^17.0.19",
-    "@types/react-native": "^0.64.13",
+    "@types/react-native": "^0.65.0",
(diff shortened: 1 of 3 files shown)


mrousavy:
Hi! Thanks for the PR! 🖤 

I think the react-native types have to be upgraded, since TS fails with this error message:

```
src/hooks/useIsForeground.ts(14,20): error TS2339: Property 'remove' does not exist on type 'void'.
14
```

Also there's a `;` missing at the end of the line, that's why the Lint Action failed :)

developerantoniosousa:
Hi @mrousavy. Is there something else missing about this PR?

mrousavy:
Hi @developerantoniosousa! Thank you for your PR!

PR #491, opened 2021-10-05
chore: Migrate from react-native-navigation to react-navigation

mrousavy:


KingAmo:
why migrate to react-navigation? is react-navigation better than RNN ? or just react-navigation has more user?

mrousavy:
@KingAmo nothing is better than the other, there are different trade-offs. I know react-native-navigation very well because I was a contributor for some time, but those are the reasons I chose react-navigation instead here:

* Simplicity (less lines of code, navigation prop instead of componentId)
* No bugs with "no-animation modals" (RNN had a few, those were already fixed but kept returning after upgrades)
* Fewer native dependencies so the native code/app builds faster.
* It's just an example app - we don't _need_ to watch performance as much as in an actual production app.

PR #559, opened 2021-10-27
feat: Add `pixelFormat` field

thomas-coldwell:
## What

This PR adds support to include the pixel format code as part of the camera device format object on iOS. This is useful if you need to select a specific format for frame processors such as is required when using MLKit: https://developers.google.com/ml-kit/reference/ios/mlkitvision/api/reference/Classes/MLKVisionImage#parameters_1

## Changes

Adds an extra field on the `device.format` called `mediaSubType` that will hold the pixel format e.g. 420v, 420f

## Tested on

iPad 8, iOS 14.4.2

### DIFF
--- a/ios/CameraViewManager.swift
+++ b/ios/CameraViewManager.swift
@@ -101,13 +101,7 @@ final class CameraViewManager: RCTViewManager {
           "supportsLowLightBoost": $0.isLowLightBoostSupported,
           "supportsFocus": $0.isFocusPointOfInterestSupported,
           "formats": $0.formats.map { format -> [String: Any] in
-            var dict = format.toDictionary()
-            // Get the pixel format (mediaSubType), remove apostrophes from start and end
-            // of value
-            if #available(iOS 13.0, *) {
-              dict["mediaSubType"] = String(format.formatDescription.mediaSubType.description.dropFirst(1).dropLast(1))
-            }
-            return dict
+            format.toDictionary()
           },
         ]
       }
--- a/ios/CameraViewManager.swift
+++ b/ios/CameraViewManager.swift
@@ -101,7 +101,13 @@ final class CameraViewManager: RCTViewManager {
           "supportsLowLightBoost": $0.isLowLightBoostSupported,
           "supportsFocus": $0.isFocusPointOfInterestSupported,
           "formats": $0.formats.map { format -> [String: Any] in
-            format.toDictionary()
+            var dict = format.toDictionary()
+            // Get the pixel format (mediaSubType), remove apostrophes from start and end
+            // of value
+            if #available(iOS 13.0, *) {
+              dict["mediaSubType"] = String(format.formatDescription.mediaSubType.description.dropFirst(1).dropLast(1))
+            }
+            return dict
           },
         ]
       }
--- a/ios/Extensions/FourCharCode+toString.swift
+++ b/ios/Extensions/FourCharCode+toString.swift
@@ -1,0 +1,17 @@
+//
+//  FourCharCode+toString.swift
+//  VisionCamera
+//
+//  Created by Thomas Coldwell on 28/10/2021.
+//  Based off this SO answer: https://stackoverflow.com/a/25625744
+//
+
+extension FourCharCode {
+    func toString() -> String {
+        var s: String = String (UnicodeScalar((self >> 24) & 255)!)
+        s.append(String(UnicodeScalar((self >> 16) & 255)!))
+        s.append(String(UnicodeScalar((self >> 8) & 255)!))
+        s.append(String(UnicodeScalar(self & 255)!))
+        return (s)
+    }
+}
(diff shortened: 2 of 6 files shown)


ios/CameraViewManager.swift
```diff
@@ -101,7 +101,13 @@ final class CameraViewManager: RCTViewManager {
...
           "supportsFocus": $0.isFocusPointOfInterestSupported,
           "formats": $0.formats.map { format -> [String: Any] in
-            format.toDictionary()
+            var dict = format.toDictionary()
+            // Get the pixel format (mediaSubType), remove apostrophes from start and end
+            // of value
+            if #available(iOS 13.0, *) {
+              dict["mediaSubType"] = String(format.formatDescription.mediaSubType.description.dropFirst(1).dropLast(1))
+            }
+            return dict
```
mrousavy: This entire thing should be moved to `AVCaptureDevice.+toDictionary`:

https://github.com/mrousavy/react-native-vision-camera/blob/7206a794a9e1a2da4def7fdf4d8a70c9eb7f43ff/ios/Extensions/AVCaptureDevice.Format%2BtoDictionary.swift#L50

src/CameraDevice.ts
```diff
@@ -171,6 +171,10 @@ export interface CameraDeviceFormat {
    * All supported video stabilization modes
    */
   videoStabilizationModes: VideoStabilizationMode[];
+  /**
+   * Specifies the pixel format on iOS. e.g. 420v, 420f
+   */
+  mediaSubType?: string;
```
mrousavy: I would prefer to make it non-optional. Every Camera has a format, that includes iPhones before iOS 13.0, and Android. So on those devices, find out what their default value is and use that. (Put in the `else` of the `if #available` on iOS, and into `CameraViewModule.getAvailableCameraDevices` on Android)

mrousavy (changes requested):
Thanks for the PR Thomas! DM me on Discord if you need help or have any questions

mrousavy:
Maybe also rename `mediaSubType` -> `pixelFormat`?

mrousavy:
Oh and lastly - use the C-style method. That one is also available before iOS 13.:

```swift
let mediaSubType = CMFormatDescriptionGetMediaSubType(format.formatDescription)
```

src/CameraDevice.ts
```diff
@@ -171,6 +171,10 @@ export interface CameraDeviceFormat {
    * All supported video stabilization modes
    */
   videoStabilizationModes: VideoStabilizationMode[];
+  /**
+   * Specifies the pixel format on iOS. e.g. 420v, 420f
+   */
+  pixelFormat: string;
```
mrousavy: Do we know all possible values of `pixelFormat`? I believe Formats have `420f`, `420v` and since the latest iOS version also `420x` (compressed), so let's not use `string`, but rather a TS union.

1. Create `PixelFormat.ts`:

```ts
/**
 * Represents the pixel format of a `Frame`.
 * * `420v`: 420 YpCbCr 8 Bi-Planar Video Range
 * * `420f`: 420 YpCbCr 8 Bi-Planar Full Range
 * * `x420`: 420 YpCbCr 10 Bi-Planar Video Range
 */
export type PixelFormat = '420f' | '420v' | 'x420'
```

And in `CameraDevice.ts`:

```ts
import type { PixelFormat } from './PixelFormat'
…
pixelFormat: PixelFormat;
```

ios/Extensions/FourCharCode+toString.swift
```diff
@@ -0,0 +1,17 @@
...
+//  Based off this SO answer: https://stackoverflow.com/a/25625744
+//
+
+extension FourCharCode {
+    func toString() -> String {
+        var s: String = String (UnicodeScalar((self >> 24) & 255)!)
+        s.append(String(UnicodeScalar((self >> 16) & 255)!))
+        s.append(String(UnicodeScalar((self >> 8) & 255)!))
+        s.append(String(UnicodeScalar(self & 255)!))
+        return (s)
```
mrousavy, line 15: isn't that reverted? so `f024` instead of `420f`? I found [this code](https://gist.github.com/patrickjuchli/d1b07f97e0ea1da5db09#file-fourcharcode-string-swift-L42-L49) that does the same thing in reverse order

mrousavy (changes requested):

ios/Extensions/FourCharCode+toString.swift
mrousavy: ah nevermind that's not reverse order 🤦‍♂️

thomas-coldwell:
Thanks for the feedback Marc - will do the implementation on the Android side now and update the pixelFormat type union with a specific list of what's available on Android too 😃

mrousavy:
1. Looks like [here's some information about pixel formats on Android](https://developer.android.com/reference/android/graphics/ImageFormat#YCBCR_P010)
2. You need to run `yarn check-all` to format Swift code

thomas-coldwell:
Will make sure to run it - it altered some other swift files (that I presume were unlinted/unformated) when I ran it before and didn't want those changes to clog up the PR 😅

Bit of a noob when it comes to these different formats, but are the Image Formats there the same as Pixel Formats or is it a case of you can tell what the Pixel Format will be from the Image Format?

PR #2, opened 2021-02-20
JSI Frame Processors (iOS)

mrousavy:
## Update (3.5.2021)

**Frame Processors are ready for iOS** 🎉. Here's what actually happened:

* I had to create a custom `AVCaptureVideoDataOutput` delegate which uses a custom `AVAssetWriter` to write the video files (turned out quite complex!), since you cannot use the `AVCaptureMovieFileOutput` and a `AVCaptureVideoDataOutput` (for frame processing) delegate at the same time.
* I had to create a custom `AVCaptureAudioDataOutput` delegate which uses a custom `AVAssetWriter` for the same reason as above
* I had to figure out how Swift <-> (Objective-)C++ communication worked
* I had to figure out how Reanimated implemented worklets.
* I have created numerous PRs at the react-native-reanimated repo to fix a few threading issues, increase performance of worklets, and touched almost all files to extract the "workletization API" to make it re-usable for other libraries (VisionCamera) ([#1768](https://github.com/software-mansion/react-native-reanimated/pull/1768), [#1859](https://github.com/software-mansion/react-native-reanimated/pull/1859), [#1860](https://github.com/software-mansion/react-native-reanimated/pull/1860), [#1861](https://github.com/software-mansion/react-native-reanimated/pull/1861), [#1863](https://github.com/software-mansion/react-native-reanimated/pull/1863), [#1883](https://github.com/software-mansion/react-native-reanimated/pull/1883), [#1984](https://github.com/software-mansion/react-native-reanimated/pull/1984))
* I created [react-native-multithreading](https://github.com/mrousavy/react-native-multithreading) to experiment with the workletization API I extracted from react-native-reanimated in [#1861](https://github.com/software-mansion/react-native-reanimated/pull/1861)
* I spawned a separate JS-Runtime which uses the Reanimated workletization API to workletize frame processors and runs them.
* I added a plugin API which uses Objective-C macros to "inject" functions into the VisionCamera JS-Runtime. 
* I added the possibility to use return values and parameters for the Plugins, which requires conversion between Objective-C types and JS types (`jsi::Value`)
* I added Swift support for the Plugin API
* I tweaked reanimated's babel plugin to allow passing custom globals (= frame processor plugins), PR yet to be merged.
* I spent a few nights debugging threading issues (`EXC_BAD_ACCESS` errors), and submitted a PR to reanimated to fix those.
* I wrote comprehensive documentation for frame processors, the plugin API and a step by step guide on how you can create your own plugins.

[...]

PR #645, opened 2021-12-12
feat: [ios] video codec option

### LINKED ISSUE #485: 🐛 iPhone 11 doesn't record video with H.264 codec

I tried to record video with H.264 codec

### Reproduceable Code

```tsx
<Camera
    ref={cameraRef}
    device={devices.front}
    video={true}
    audio={true}
    isActive={true}
    preset="iframe-960x540"
    fps={25}
    onInitialized={() => setCameraReady(true)}
/>
```

### What happened instead?

When I tried to record a video with H.264 codec, I got HEVC video codec.

### Relevant log output

_No response_

### Device

iPhone 11 (IOS 15.0.1)

### VisionCamera Version

2.8.0.

### Additional information

- [ ] I am using Expo

paulrostorp:
## What

For IOS only, adds an optional `videoCodec` parameter to the `startRecording` options to specify the codec to record with. Requires IOS 11 (see commit a09201c2bc88eb6c11417046824cac88d9338069 vs 17b960603298e19d45e10933fe6dbf1da6ee00a6 .... please advise if not suitable)

## Changes
- handle user provided video codec in `CameraView+RecordVideo.swift`
- adds an optional `videoCodec` parameter to the `RecordVideoOptions` interface
- adds a `getAvailableVideoCodecs` method which takes a `fileType` parameter (see: https://stackoverflow.com/questions/46843849/why-is-videodataoutputs-availablevideocodectypes-empty/58017253)

## Tested on
- iphone 7, ios 12
- ipad air, ios 14.4

## Related issues

resolves #485

### DIFF
--- a/ios/CameraView+RecordVideo.swift
+++ b/ios/CameraView+RecordVideo.swift
@@ -107,17 +107,20 @@ extension CameraView: AVCaptureVideoDataOutputSampleBufferDelegate, AVCaptureAud
         callback.reject(error: .capture(.createRecorderError(message: nil)), cause: error)
         return
       }
-
+      
+      var videoCodec:AVVideoCodecType?
+      if let codecString = options["videoCodec"] as? String {
+        videoCodec = AVVideoCodecType.init(withString: codecString)
+      }
+      
       // Init Video
-      guard var videoSettings = videoOutput.recommendedVideoSettingsForAssetWriter(writingTo: fileType),
+      guard let videoSettings = videoCodec != nil ?
+              videoOutput.recommendedVideoSettings(forVideoCodecType: videoCodec!, assetWriterOutputFileType: fileType) :
+                videoOutput.recommendedVideoSettingsForAssetWriter(writingTo: fileType),
             !videoSettings.isEmpty else {
-        callback.reject(error: .capture(.createRecorderError(message: "Failed to get video settings!")))
-        return
-      }
-
-      if let videoCodec = options["videoCodec"] as? String {
-        videoSettings["AVVideoCodecKey"] = AVVideoCodecType.init(withString: videoCodec)
-      }
+              callback.reject(error: .capture(.createRecorderError(message: "Failed to get video settings!")))
+              return
+            }
 
       // get pixel format (420f, 420v, x420)
       let pixelFormat = CMFormatDescriptionGetMediaSubType(videoInput.device.activeFormat.formatDescription)
--- a/ios/CameraView+RecordVideo.swift
+++ b/ios/CameraView+RecordVideo.swift
@@ -114,10 +114,8 @@ extension CameraView: AVCaptureVideoDataOutputSampleBufferDelegate, AVCaptureAud
       }
 
       // Init Video
-      guard let videoSettings = videoCodec != nil ?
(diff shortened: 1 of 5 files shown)


ios/CameraView+RecordVideo.swift
```diff
@@ -108,12 +108,20 @@ extension CameraView: AVCaptureVideoDataOutputSampleBufferDelegate, AVCaptureAud
...
+      if let codecString = options["videoCodec"] as? String {
+        videoCodec = AVVideoCodecType(withString: codecString)
+      }
+
       // Init Video
-      guard let videoSettings = videoOutput.recommendedVideoSettingsForAssetWriter(writingTo: fileType),
-            !videoSettings.isEmpty else {
+      guard let videoSettings = videoCodec != nil ?
+        videoOutput.recommendedVideoSettings(forVideoCodecType: videoCodec!, assetWriterOutputFileType: fileType) :
+        videoOutput.recommendedVideoSettingsForAssetWriter(writingTo: fileType),
```
mrousavy: This is too much complexity, try to lift this out of that guard to keep it easy to read.

src/VideoFile.ts
```diff
@@ -21,6 +32,13 @@ export interface RecordVideoOptions {
    * Called when the recording has been successfully saved to file.
    */
   onRecordingFinished: (video: VideoFile) => void;
+  /**
+   * IOS only
+   * Specify the video codec to use. To get a list of available video codecs use the `getAvailableVideoCodecs()` function.
```
mrousavy:
```suggestion
   * Set the video codec to record in. Different video codecs affect video quality and video size.
   * To get a list of all available video codecs use the `getAvailableVideoCodecs()` function.
```

mrousavy (changes requested):
Hey Paul, thank you for your PR! 🖤 

VisionCamera's philosophy is to avoid platform-specifics and provide a uniform abstraction that doesn't require the user to understand the native underlying Camera architectures to use VisionCamera. 
Can you add the Video Codec option to Android as well? The latest CameraX version ships with a separate Video Recording library, maybe that allows Video Codec? Or via Camera2 interop.

Also I left two small comments, otherwise it looks good :)

src/VideoFile.ts
paulrostorp: This description is wrong, that method isn't actually available, I made a copy-paste error when copying from https://github.com/mrousavy/react-native-vision-camera/pull/2/files#diff-432ccbc3244e65723bf8031a79bbad1191b792c59affdbd2ea93ef6e6b1138d4L6

paulrostorp:
@mrousavy Thanks for your review, I've made some changes according to it, incl. adding a `getAvailableVideoCodecs` method.

I, definitely, agree with providing cross-platform features, however, after spending a few hours looking into it, it doesn't seem that the `Camera2interop` or the `Recorder` from the latest cameraX expose functionality to specify the codec or to retrieve available codecs.

The only thing I am able to find is using the `MediaRecorder`(https://developer.android.com/reference/android/media/MediaRecorder) but that doesn't seem suitable...

My native android skills are almost non-existent, so I'm quite unsure about this, but to support these findings, I couldn't find any other React Native (or flutter) packages that support choosing a codec on android.

mrousavy:
Nice job Paul! Thank you for your PR. 🖤  Let's ship it 🚀

mrousavy (approved):

PR #661, opened 2021-12-18
Update ZOOMING.mdx

hirbod:
Change example link from fixed commit to master (this prevents copying old/non-working code)


mrousavy:
thanks!

PR #710, opened 2022-01-03
fix: version_code issue on Android 29

Stukz:
## What

I'm currently using the library on a project with React Native 0.63.3 and Android Version 29 and with the last fix, I'm not able to run it due to the condition against Android 30

this PR adds the option to install on React Native with VERSION_CODES = 29 without having to upgrade

## Changes

* This PR changes `Build.VERSION.SDK_INT >= Build.VERSION_CODES.R` to support Build.VERSION_CODES.Q without having to upgrade React Native/Android Version

## Tested on
* Pixel 2 API 30


Stukz:
I was doing a little bit more testing and this isn't working as expected, not sure if `RequiresApi` or `TargetApi` can fix this otherwise I'll have to use a patch and remove the lines where Build.VERSION.SDK_INT >= Build.VERSION_CODES.R it's used

PR #717, opened 2022-01-05
Fix Setup document.

YutaSugimura:
## What
Fixed Setup document.
The documentation for setting targetSdkVersion to 31 or higher is only in [Troubleshooting](https://mrousavy.com/react-native-vision-camera/docs/guides/troubleshooting#android), so add it to the setup documentation.

## Changes
Added necessary information for android setup.

### DIFF
--- a/docs/docs/guides/SETUP.mdx
+++ b/docs/docs/guides/SETUP.mdx
@@ -81,6 +81,30 @@ Open your project's `AndroidManifest.xml` and add the following lines inside the
 
 <!-- optionally, if you want to record audio: -->
 <uses-permission android:name="android.permission.RECORD_AUDIO" />
+
+<application>
+  ...
+  <activity
+    ...
+    android:exported="true" <- ADD
+  >
+    ...
+  </activity>
+</application>
+```
+
+Change compileSdkVersion and targetSdkVersion in android/build.gradle to 31.
+
+```gradle
+  buildscript {
+    ext {
+        ...
+        compileSdkVersion = 31
+        targetSdkVersion = 31
+        ...
+    }
+    ...
+  }
 ```
 
 </TabItem>


hirbod:
For me it was enough to just bump `compileSdkVersion` to 31, `targetSdkVersion` can still be 30.
We should also mention that there is a nice working `expo-config-plugin` for the EAS/Expo Folks to bump the version.

The troubleshooting should also mention that if you add kotlin-gradle-plugin dependencies, it has to be minimum `1.5.30`.
I had another library pulling 1.4.10 and it was breaking as well.

```
        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:1.5.30"
```

CanRau:
Not sure I understand your comment correctly @hirbod but the provided `expo-config-plugin` (I believe that's what you're referring to?) doesn't bump `compileSdkVersion` or anything like that.

As I'm pretty new to React Native and Expo I was looking for a way to follow the setup instructions including `expo prebuild` and then being able to `expo run:android` but as it seems I have to manually follow the troubleshooting steps to bumb versions for iOS & Android.

Couldn't/shouldn't the `expo-config-plugin` take care of this?
I started my project using `expo init` running expo `5.3.0`
For now it's okay for me to "eject" and bump versions manually though I feel like it should be possible to stay in a little more managed flow with the config plugin right?

Should I create a new issue for that or is there already one?

hirbod:
@CanRau my first PR did that, but Marc asked to remove it and not handle it by the shipped expo-plugin.


```ts
const {
    withProjectBuildGradle
} = require('@expo/config-plugins')

const setCompileSdkVersion = (buildGradle, sdkVersion) => {
    const regexpCompileSdkVersion = /\bcompileSdkVersion\s*=\s*(\d+)/
    const match = buildGradle.match(regexpCompileSdkVersion)

    if (match) {
        const existingVersion = parseInt(match[1], 10)

        if (existingVersion < sdkVersion) {
            buildGradle = buildGradle.replace(
                /\bcompileSdkVersion\s*=\s*\d+/,
                `compileSdkVersion = ${sdkVersion}`
            )
        } else {
            console.warn(`minSdkVersion is already >= ${sdkVersion}, skipping`);
        }
    }

    return buildGradle
}

module.exports = (config, sdkVersion) => {
    return withProjectBuildGradle(config, (config) => {
        if (config.modResults.language === 'groovy') {
            config.modResults.contents = setCompileSdkVersion(
                config.modResults.contents,
                sdkVersion
            )
        } else {
            throw new Error(
                "Can't set minSdkVersion in the project build.gradle, because it's not groovy"
            )
        }
        return config
    })
}
```

Save the file somewhere (I have a directory `plugins`) as withCompileSdkVersion.js and add the contents.
Update your config and add it to your plugins section.

```
    ['./plugins/withCompileSdkVersion', 31],
```

eduardoborges (approved):

mrousavy:
Hey thanks, but thanks to the CAemra2 rewrite this is no longer required!

PR #735, opened 2022-01-11
Use ViewGroupManager instead of SimpleViewManager

NilsKrause:
## What

* This PR adds support for ListViews as child components of the Camera component.

## Changes

* This PR just changes the `SimpleViewManager` used in `android/src/main/java/com/mrousavy/camera/CameraViewManager.kt` to `ViewGroupManager`.

## Tested on

* Pixel 4a, Android 12

## Related issues

* None that I know of

### DIFF
--- a/android/src/main/java/com/mrousavy/camera/CameraViewManager.kt
+++ b/android/src/main/java/com/mrousavy/camera/CameraViewManager.kt
@@ -3,12 +3,12 @@ package com.mrousavy.camera
 import com.facebook.react.bridge.ReactApplicationContext
 import com.facebook.react.bridge.ReadableMap
 import com.facebook.react.common.MapBuilder
-import com.facebook.react.uimanager.SimpleViewManager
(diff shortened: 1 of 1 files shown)


mrousavy:
Hey, thank you for your PR!

Why would you want to add children to the Camera component though?

NilsKrause:
I personally have no direct use case for it at the moment, but I was experimenting a bit with the component and saw that the component accepts children - but it just crashed when I added a FlatList. So I thought, when children are supported it would only make sense to also allow Lists.

Angelk90:
@NilsKrause : What android device did you try on?

NilsKrause:
@Angelk90 as described in the first post I tested the behaviour on a Google Pixel 4a with Android 12.

Angelk90:
@NilsKrause : Tested on the emulator?

NilsKrause:
Emulator and actual device

PR #744, opened 2022-01-15
fix(types): duration is not optional anymore

hirbod:
Duration is not optional anymore and not limited to iOS. This PR fixes the types.

### DIFF
--- a/src/VideoFile.ts
+++ b/src/VideoFile.ts
@@ -50,10 +50,6 @@ export interface RecordVideoOptions {
 export interface VideoFile extends TemporaryFile {
   /**
    * Represents the duration of the video, in seconds.
-   *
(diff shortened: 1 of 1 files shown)


hirbod:
@mrousavy it also needs the `size` type, right? It's neither in the interface extension and not in TemporaryFile type. Can we assume that this value is ALWAYS present? You know the native codebase better. If true, I would also add size

mrousavy:
Hey thanks for the PR!
Yes, size is also always present

hirbod:
What does size represent exactly? On iOS, I don't have `size` in my onMediaCaptured. Android does have a value

mrousavy:
Ah then we need a separate PR to add it for iOS native

PR #745, opened 2022-01-15
Add depth data processing for iOS (TrueDepth and dual camera)

thomas-coldwell:
## What

- This PR adds a `AVCaptureDepthDataOutput` which enables the depth data to be passed to frame processor plugins. 
- It will also introduce a new prop `enableDepthData` to select whether depth data should be included when using one of the depth enabled cameras e.g. `builtInDualCamera` or `builtInTrueDepth`

## Changes

- Adds a `AVCaptureDepthDataOutput`
- Adds a `AVCaptureDataOutputSynchronizer` to sync up the video and depth data
- Adds the corresponding delegate class for the synchronizer and handles the processing in the method `dataOutputSynchronizer` rather than `captureOutput` as it is not a single output
- Adds a depth property of type `CVPixelBuffer?` to the `Frame` class

## Tested on

- iPhone 13 Pro, 15.1.1


thomas-coldwell:
I've edited a bunch of stuff just to test out the depth data in frame processors that I'll remove in a new commit as well as linting + formatting btw. Not ready to merge yet 😅

mrousavy:
Shouldn't `depth` also be a `Frame`?

So the TypeScript type should look like this:

```ts
// Already existing Frame type
type Frame = {
  width: number
  height: number
  ...
  depth: Omit<Frame, 'depth'>
}
```

no? Because the depth frame can actually have a different width/height than the RGB/YUV frame. 

### Edit

Technically that's already how it works, but in the JSI HostObject we just need to return a new Frame HostObject when the user is accessing `.depth`, I can do that too.

thomas-coldwell:
Yeah it will need to be its own frame with `depth` as a property of that. I haven't updated that end of it yet - purely checking if the `depth` buffer comes through on the frame processor at the minute, but will need to add that as a property of it.

thomas-coldwell:
The last commit just adds `depthDataFormat` as a prop now for the Camera component so that the user can specify which format they want given they have selected a format that contains their desired depth data format. 

Also the depth can be toggled on/off with `enableDepthData` which simply switches it to just configure the video output as before.

thomas-coldwell:
Took a stab at it and seems to all be working - omitted `isValid` and `close()` from the depth frame too as these only make sense in the CMSampleBuffer context, but the depth is just the raw CVPixelBuffer. 

I think the last thing to look at is just making sure the `AVCaptureOutputSynchronizer` is properly syncing the video and depth outputs. Once this is good we can do some tests in a frame processor just to check its getting a stable depth value (at least from pixel to pixel, if not absolute physical accuracy in meters) and then I think its about there 😃  If there's anything else you spot that I've missed or we want to pass into the frame processor callback let me know

ios/CameraQueues.swift
```diff
@@ -24,6 +24,13 @@ public class CameraQueues: NSObject {
                                                      autoreleaseFrequency: .inherit,
                                                      target: nil)
```
mrousavy: do we still need that queue then?

ios/CameraView+AVCaptureSession.swift
```diff
@@ -125,16 +127,44 @@ extension CameraView {
...
+        captureSession.addOutput(videoOutput!)
+        depthOutput = AVCaptureDepthDataOutput()
+        // Add a depth data output
+        if captureSession.canAddOutput(depthOutput!) {
+          captureSession.addOutput(depthOutput!)
+          depthOutput!.isFilteringEnabled = true
+          if let connection = depthOutput!.connection(with: .depthData) {
+            connection.isEnabled = true
+          } else {
+            print("No AVCaptureConnection")
```
mrousavy: Shouldn't that also raise the error?

ios/CameraView+AVCaptureSession.swift
```diff
@@ -125,16 +127,44 @@ extension CameraView {
...
+          }
+        } else {
+          invokeOnError(.parameter(.unsupportedOutput(outputDescriptor: "depth-output")))
+          captureSession.commitConfiguration()
```
mrousavy: That should probably be removed

ios/CameraView+AVCaptureSession.swift
```diff
@@ -147,7 +177,7 @@ extension CameraView {
   // pragma MARK: Configure Device
 
   /**
-   Configures the Video Device with the given FPS, HDR and ColorSpace.
+   Configures the Video Device with the given FPS, HDR, ColorSpace and Depth Data Format.
```
mrousavy, line 179: nice! ;)

ios/CameraView+AVCaptureSession.swift
```diff
@@ -203,6 +233,30 @@ extension CameraView {
...
         device.activeColorSpace = avColorSpace
       }
+      if #available(iOS 13.0, *) {
+        if enableDepthData?.boolValue == true && !device.activeFormat.supportedDepthDataFormats.isEmpty {
+          if let depthDataFormat = depthDataFormat as String? {
+            // Check if the selected depth format is supported as part of the activeFormat
+            let avCaptureDepthDataFormat = depthDataFormat
+            if device.activeFormat.supportedDepthDataFormats.map({
+              CMFormatDescriptionGetMediaSubType($0.formatDescription).toString()
+            }).contains(avCaptureDepthDataFormat) {
```
mrousavy: instead of two loops you could probably just create a single `find(where:)` loop?

ios/CameraView+AVCaptureSession.swift
```diff
@@ -203,6 +233,30 @@ extension CameraView {
...
+              })!
+              device.activeDepthDataFormat = selectedDepthFormat
+            } else {
+              invokeOnError(.format(.invalidDepth))
+              return
+            }
+          }
+        }
+      } else {
+        // Fallback on earlier versions
```
mrousavy: log that info to console here (with `ReactLogger`)

ios/CameraView+RecordVideo.swift
```diff
@@ -189,7 +192,43 @@ extension CameraView: AVCaptureVideoDataOutputSampleBufferDelegate, AVCaptureAud
...
+    // Read all outputs
+    guard let syncedVideoData: AVCaptureSynchronizedSampleBufferData =
+      synchronizedDataCollection.synchronizedData(for: videoOutput!) as? AVCaptureSynchronizedSampleBufferData else {
+      // only work on synced pairs
+      ReactLogger.log(level: .info, message: "Video data out of sync")
+      return
+    }
+    guard let syncedDepthData: AVCaptureSynchronizedDepthData =
+      synchronizedDataCollection.synchronizedData(for: depthOutput!) as? AVCaptureSynchronizedDepthData else {
+      ReactLogger.log(level: .info, message: "Depth data out of sync")
```
mrousavy: Shouldn't that be `.error` or at least `.warning`? Or even `invokeOnError`?

ios/CameraView+RecordVideo.swift
```diff
@@ -189,7 +192,43 @@ extension CameraView: AVCaptureVideoDataOutputSampleBufferDelegate, AVCaptureAud
     }
   }
 
+  public final func dataOutputSynchronizer(_: AVCaptureDataOutputSynchronizer, didOutput synchronizedDataCollection: AVCaptureSynchronizedDataCollection) {
+    // Read all outputs
+    guard let syncedVideoData: AVCaptureSynchronizedSampleBufferData =
+      synchronizedDataCollection.synchronizedData(for: videoOutput!) as? AVCaptureSynchronizedSampleBufferData else {
+      // only work on synced pairs
+      ReactLogger.log(level: .info, message: "Video data out of sync")
```
mrousavy: Shouldn't that be `.error` or at least `.warning`? Or even `invokeOnError`?

ios/CameraView+RecordVideo.swift
```diff
@@ -189,7 +192,43 @@ extension CameraView: AVCaptureVideoDataOutputSampleBufferDelegate, AVCaptureAud
...
+      return
+    }
+    guard let syncedDepthData: AVCaptureSynchronizedDepthData =
+      synchronizedDataCollection.synchronizedData(for: depthOutput!) as? AVCaptureSynchronizedDepthData else {
+      ReactLogger.log(level: .info, message: "Depth data out of sync")
+      return
+    }
+
+    if syncedDepthData.depthDataWasDropped || syncedVideoData.sampleBufferWasDropped {
+      return
```
mrousavy, line 210: Also maybe log something or do `invokeOnError`?

ios/CameraView+RecordVideo.swift
```diff
@@ -189,7 +192,43 @@ extension CameraView: AVCaptureVideoDataOutputSampleBufferDelegate, AVCaptureAud
...
+      if !isRunningFrameProcessor {
+        // we're not in the middle of executing the Frame Processor, so prepare for next call.
+        CameraQueues.frameProcessorQueue.async {
+          self.isRunningFrameProcessor = true
+          let frame = Frame(bufferAndDepth: sampleBuffer, depth: depthPixelBuffer, orientation: self.bufferOrientation)
+          frameProcessor(frame)
+          self.isRunningFrameProcessor = false
+        }
+      }
+    }
```
mrousavy, line 227: We could probably create a method that wraps all of this logic so we only call the call constructor once, because this code is mostly duplicated.

E.g. something like:

```swift
if isFrameProcessorReady {
  callFrameProcessor(withFrame: Frame(..))
}
```

But I'll do that refactor later. Leave it as is now

ios/CameraView+RecordVideo.swift
```diff
@@ -189,7 +192,43 @@ extension CameraView: AVCaptureVideoDataOutputSampleBufferDelegate, AVCaptureAud
...
+  }
+
   public final func captureOutput(_ captureOutput: AVCaptureOutput, didOutput sampleBuffer: CMSampleBuffer, from _: AVCaptureConnection) {
+    ReactLogger.log(level: .info, message: "Capture output: \(captureOutput.debugDescription)")
```
mrousavy: should be removed

ios/CameraView.swift
```diff
@@ -241,7 +246,7 @@ public final class CameraView: UIView {
         }
 
         if shouldUpdateOrientation {
-          self.updateOrientation()
+//          self.updateOrientation()
```
mrousavy: Are there problems with orientation?

ios/CameraView.swift
```diff
@@ -314,7 +319,7 @@ public final class CameraView: UIView {
 
   @objc
   func onOrientationChanged() {
-    updateOrientation()
+//    updateOrientation()
```
mrousavy: Are there problems with orientation?

ios/Extensions/AVCaptureDevice.Format+matchesFilter.swift
```diff
@@ -98,6 +98,12 @@ extension AVCaptureDevice.Format {
       }
     }
 
+    if let depthData = filter.value(forKey: "supportsDepthData") as? NSNumber {
+      if depthData.boolValue == true && supportedDepthDataFormats.isEmpty {
```
mrousavy, line 102: Can't this conditional go both ways?

ios/Frame Processor/Frame.h
```diff
@@ -16,7 +16,10 @@
 
 - (instancetype) initWithBuffer:(CMSampleBufferRef)buffer orientation:(UIImageOrientation)orientation;
 
+- (instancetype) initWithBufferAndDepth:(CMSampleBufferRef)buffer depth:(CVPixelBufferRef)depth orientation:(UIImageOrientation)orientation;
+
 @property (nonatomic, readonly) CMSampleBufferRef buffer;
+@property (nonatomic, readonly) CVPixelBufferRef depth;
```
mrousavy: pls add `nullable` modifier

ios/Frame Processor/Frame.h
```diff
@@ -16,7 +16,10 @@
 
 - (instancetype) initWithBuffer:(CMSampleBufferRef)buffer orientation:(UIImageOrientation)orientation;
 
+- (instancetype) initWithBufferAndDepth:(CMSampleBufferRef)buffer depth:(CVPixelBufferRef)depth orientation:(UIImageOrientation)orientation;
+
 @property (nonatomic, readonly) CMSampleBufferRef buffer;
```
mrousavy: pls add `nonnull` modifier

ios/Frame Processor/Frame.h
```diff
@@ -16,7 +16,10 @@
...
+@property (nonatomic, readonly) CVPixelBufferRef depth;
 @property (nonatomic, readonly) UIImageOrientation orientation;
```
mrousavy, line 23: pls add `nonnull` modifier

ios/Frame Processor/FrameHostObject.mm
```diff
@@ -78,6 +79,31 @@
     auto planesCount = CVPixelBufferGetPlaneCount(imageBuffer);
     return jsi::Value((double) planesCount);
   }
+  if (name == "depth") {
+      if (this->frame.depth != nil) {
```
mrousavy: Early return is better here.

```c++
if (this->frame.depth == nil) {
  return jsi::Value::undefined();
}
```

ios/Frame Processor/FrameHostObject.mm
```diff
@@ -78,6 +79,31 @@
...
+          auto width = CVPixelBufferGetWidth(frame.depth);
+          auto height = CVPixelBufferGetHeight(frame.depth);
+          auto bytesPerRow = CVPixelBufferGetBytesPerRow(frame.depth);
+          auto planesCount = CVPixelBufferGetPlaneCount(frame.depth);
+          depthFrame.setProperty(runtime, "toString", jsi::Function::createFromHostFunction(runtime, jsi::PropNameID::forUtf8(runtime, "toString"), 0, toString));
+          depthFrame.setProperty(runtime, "width", jsi::Value((double) width));
+          depthFrame.setProperty(runtime, "height", jsi::Value((double) height));
+          depthFrame.setProperty(runtime, "bytesPerRow", jsi::Value((double) bytesPerRow));
+          depthFrame.setProperty(runtime, "planesCount", jsi::Value((double) planesCount));
+          return depthFrame;
```
mrousavy: hmm, do you think it makes sense to have a separate `HostObject` for that? Then it's lazy and easier to extend.

src/Frame.ts
```diff
@@ -46,4 +46,8 @@ export interface Frame {
    * ```
    */
   close(): void;
+  /**
+   * The depth frame which is attached to the video frame
+   */
+  depth?: Omit<Frame, 'depth' | 'isValid' | 'close'>;
```
mrousavy: Use `Pick` instead of `Omit`.

src/CameraDevice.ts
```diff
@@ -160,6 +176,10 @@ export interface CameraDeviceFormat {
    * Specifies whether this format supports HDR mode for photo capture
    */
   supportsPhotoHDR: boolean;
+  /**
+   * Specifies whether this format supports depth data capture
+   */
+  supportedDepthDataFormats: DepthDataFormat[];
```
mrousavy, line 182: Please make the Android func return `[]` (aka no depth data formats are supported). Change is required in `CameraViewModule.kt`

ios/CameraQueues.swift
thomas-coldwell: Will need to check as part of the issue with the synchronizer - it can probably be removed as we are still getting the `Depth data out of sync` issue with it on this separate queue anyways 😅

ios/CameraView+AVCaptureSession.swift
thomas-coldwell: Why? Don't we need to return early and throw an error if the use has requested depth data and the depth output can't be attached to the capture session?

ios/CameraView.swift
thomas-coldwell: Couldn't build the app - kept saying this method was undefined somewhere - not sure what was going on with it so commented it all out for the time being

thomas-coldwell: Actually nvm - seems to be fine now, must have been XCode having a bad day 🤦‍♂️

ios/Frame Processor/FrameHostObject.mm
thomas-coldwell: Yeah if we can - wasn't sure how to set it up separately and by lazy do you mean lazy access of each of the properties - what's the advantage of this in JSI, I was a little confused with how properties are accessed in the sense of `setProperty` and getting it from the object vs. `if (name == "propertyName")`?

ios/CameraView+RecordVideo.swift
thomas-coldwell: Yeah that would be brill - will leave for now :)

ios/CameraView+AVCaptureSession.swift
mrousavy: Yeah but the commit should be removed. I am committing the changes above anyways (`defer` block)

ios/Frame Processor/FrameHostObject.mm
mrousavy: Well in this case it doesn't matter since you only get simple number properties, but if we had a `ArrayBuffer` of the pixels then it gets more complicated

ios/CameraQueues.swift
thomas-coldwell: Queue removed - not required for keeping them in sync and produced the same error with/without it

ios/CameraView+TakePhoto.swift
```diff
@@ -78,6 +78,10 @@ extension CameraView {
       if #available(iOS 12.0, *) {
         photoSettings.isPortraitEffectsMatteDeliveryEnabled = photoOutput.isPortraitEffectsMatteDeliveryEnabled
       }
+        
+      if self.enableDepthData?.boolValue == true {
+        photoSettings.embedsDepthDataInPhoto = true;
```
mrousavy: Are you sure that this works? From reading the documentation of that prop, it says that it is enabled by default..

Also I'd prefer the direct assignment:

```swift
photoSettings.embedsDepthDataInPhoto = photoSettings.isDepthDataDeliveryEnabled
```

thomas-coldwell: It should be yeah - I'll remove this and set it up that it just takes the available `depthData` CVPixelBuffer of the captured image and writes this to an additional file for now. I think this will help keep it more modular and easier to access than having to try and decode the depth data out of the saved JPEG directly or through metadata which seems really odd and convoluted in the first place for Apple to embed it like that

OlleThunbergDevoteam:
What's the status of this PR? 

Is it functional?

mrousavy:
This PR is amazing work, but unfortunately got a ton of conflicts because of V3. 

Let's keep this open for reference if I implement depth at some point. The synchronizer is amazing!

I wonder if this is possible on Android at all?

android/src/main/java/com/mrousavy/camera/CameraViewModule.kt
```diff
@@ -283,6 +283,7 @@ class CameraViewModule(reactContext: ReactApplicationContext) : ReactContextBase
               format.putString("autoFocusSystem", "none") // TODO: Revisit getAvailableCameraDevices (autoFocusSystem) (CameraCharacteristics.CONTROL_AF_AVAILABLE_MODES or CameraCharacteristics.LENS_INFO_FOCUS_DISTANCE_CALIBRATION)
               format.putArray("videoStabilizationModes", videoStabilizationModes)
               format.putString("pixelFormat", pixelFormat)
+              format.putArray("supportedDepthDataFormats", Arguments.createArray())
```
renanmav, line 286: What this does for Android?

mrousavy: Nothing, it's an empty array meaning there are no supported depth data formats on Android. I honestly don't know if Android cameras have depth sensors?

mrousavy:
Hey Tom thanks for your PR!

Exciting news: this is now obsolete, as VisionCamera V5 now supports depth streaming! 🥳 

(Unfortunately not on Android, but that's a CameraX limitation)

PR #772, opened 2022-01-27
Add Dynamsoft Barcode Reader plugin

xulihang:
## What

add new plugin for barcode reading using Dynamsoft Barcode Reader

## Tested on

* iPhone SE iOS 15.1
* Sharp S2 Android 8


mrousavy:
Very cool, thank you!

PR #816, opened 2022-02-09
feat: Add `autoRotateAnalysisImage` prop

thomas-coldwell:
## What

- This PR adds an additional android-only prop called `autoRotateAnalysisImage` that if set rotates each frame passed to the frame processor callback to the upright (0 degree) orientation, see https://developer.android.com/reference/androidx/camera/core/ImageAnalysis.Builder#setOutputImageRotationEnabled(boolean).
- This adds an additional overhead per frame to perform the rotation
- This is required when performing any image analysis on Android as the image proxy rotation can vary from device to device based on the orientation of the camera sensor

## Changes

- Adds `autoRotateAnalysisImage` boolean prop

## Tested on

Tested on a OnePlus 6

### DIFF
--- a/android/src/main/java/com/mrousavy/camera/CameraView.kt
+++ b/android/src/main/java/com/mrousavy/camera/CameraView.kt
@@ -72,7 +72,7 @@ class CameraView(context: Context, private val frameProcessorThread: ExecutorSer
     const val TAG = "CameraView"
     const val TAG_PERF = "CameraView.performance"
 
-    private val propsThatRequireSessionReconfiguration = arrayListOf("cameraId", "format", "fps", "hdr", "lowLightBoost", "photo", "video", "enableFrameProcessor")
+    private val propsThatRequireSessionReconfiguration = arrayListOf("cameraId", "format", "fps", "hdr", "lowLightBoost", "autoRotateAnalysisImage", "photo", "video", "enableFrameProcessor")
     private val arrayListOfZoom = arrayListOf("zoom")
   }
 
@@ -93,6 +93,7 @@ class CameraView(context: Context, private val frameProcessorThread: ExecutorSer
   var hdr: Boolean? = null // nullable bool
   var colorSpace: String? = null
   var lowLightBoost: Boolean? = null // nullable bool
+  var autoRotateAnalysisImage: Boolean? = null // nullable bool
   // other props
   var isActive = false
   var torch = "off"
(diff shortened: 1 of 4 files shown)


mrousavy:
If it is _required when performing any image analysis on Android_ we should just enable it by default and don't make it configurable, no? Same behaviour as on iOS then?

thomas-coldwell:
Oh sorry I meant to say _most_ image analysis as it does add an extra over head for each frame that gets processed and you might not need it if you don't need to scan across the image or read from pixels at a particular X,Y.

But maybe it's better if it is just the exact same API as iOS? I just don't want it to add an additional 30ms per frame for frame processor plugins where it's not needed - what do you think?

mrousavy:
> or read from pixels at a particular X,Y.

That would be wrong too since it's not oriented correctly, no?

Can you give me some examples where the user might **not** want this enabled?

mrousavy:
> I just don't want it to add an additional 30ms per frame for frame processor plugins where it's not needed - what do you think?

As always, Android is making a lot of trouble. Why do their Camera APIs suck so much? Now I understand all of the "poor android camera quality" memes

thomas-coldwell:
Cases like this where MLKit has a class for creating an input image that seems to do it internally https://github.com/rodgomesc/vision-camera-face-detector/blob/master/android/src/main/java/com/visioncamerafacedetector/VisionCameraFaceDetectorPlugin.java#L140
or on ColorWaver where sampling pixels isn't required and adding this would just reduce performance https://github.com/mrousavy/Colorwaver/blob/main/android/app/src/main/java/com/mrousavy/colorwaver/PaletteFrameProcessorPlugin.java

mrousavy:
Ah yes that makes sense lol. Hmm, can we do this operation ourselves? What if we just add it to the native Frame type so the frame processor plugin can call it on demand?

thomas-coldwell:
Do you mean something along the lines of `var uprightFrame = frame.getUprightFrame()` in the frame processor plugin to return the corrected frame if required?

mrousavy:
Yeah, this will require a breaking change though because we no longer pass `ImageProxy`, but rather a custom `Frame` instance. At least this will now be in sync with iOS?

thomas-coldwell:
Yeah I think it makes sense that there is a Frame instance on android so it's the same API on both. I'll get that into the PR 🙌

xulihang:
If we only need to get the coordinates after analysis, we may not need to rotate the image. We can just rotate the coordinates. But I haven't figured out how to do this exactly.

xulihang:
> We can just rotate the coordinates.

I've figured out how to rotate the coordinates.

https://github.com/xulihang/vision-camera-dynamsoft-barcode-reader/blob/de144377440c35e270d8128ba47986efdbe0218a/android/src/main/java/com/visioncameradynamsoftbarcodereader/VisionCameraDBRPlugin.java#L160

mrousavy:
hey sorry tom for not merging this back then I probably missed this

Issue #770, opened 2022-01-26
❓ Android resolution and FPS setting


seolhyeonyang:
### Question

I'm a React-Native beginner.
I want to adjust the resolution and FPS on Android.
Currently, when you shoot a video, it comes out at 1920x1080, 30fps.
I want to raise it to 3820 x 2160, 60fps.

```
            <Camera
                ref={camera}
                style={{flex:1}}
                device={phonedata}
                isActive={true}
                video={true}
                audio={true}
                fps={60}
                videoWidth={3840}
                videoHeight={2160}
            />
```
It's my code.

It's information from 'Camera.getAvailableCameraDevices()'.

[...]

Issue #281, opened 2021-07-17
🐛 Android: incorrect resolution/size photo output.


julianklumpers:
### What were you trying to do?

On android, I get a max photo output of 1080x1920 resolution. While the camera is capable of taking pictures of 3000x4000. It looks like the camera is zoomed in somehow, but i have zoom on 0.

I tried the expo-camera/react-native-camera library and their output size is correct.

### What happened instead?

Photo output is max 1080x1920 while the camera is capable of 3000x4000

### Devices

Samsung A41
Samsung A40
Samsung Galaxy S9
Samsung Galaxy S10
Sony xperia XZ1

### VisionCamera Version

2.5.0

### Additional information

- [ ] I am using Expo
- [X] I have read the [Troubleshooting Guide](https://mrousavy.github.io/react-native-vision-camera/docs/guides/troubleshooting)
- [X] I agree to follow this project's [Code of Conduct](https://github.com/mrousavy/react-native-vision-camera/blob/main/CODE_OF_CONDUCT.md)
- [X] I searched for [similar issues in this repository](https://github.com/mrousavy/react-native-vision-camera/issues) and found none.

mrousavy:
Please share the code of the camera component. Are you using `format={...}`? What `device={...}` are you using? expo-camera and react-native-camera both do not support custom/multiple devices selection nor format selection

julianklumpers:
Hi Marc,

Thank you for your reply.
The devices I tested it on is a Samsung A41 and a Samsung S9+. IOS is all fine.

**1. Did you use format={...}**
Yes, I did use format={...} no luck either. 

First i try'd the `useCameraFormat`. That gave me a resolution of 1080x1920.
Then i try'd a custom format. I used the sort format function from the documentation to get the highest resolution
https://mrousavy.github.io/react-native-vision-camera/docs/guides/formats#sort
output: 
[...]

PR #833, opened 2022-02-17
fix: switch width and height of target resolution if portrait for Android

xulihang:
## What

This PR fixes the incorrect resolution setting for Android.

## Changes

switch height and width of target resolution if portrait for Android

## Tested on

Sharp S2, Android 8

## Related issues

https://github.com/mrousavy/react-native-vision-camera/issues/770
https://github.com/mrousavy/react-native-vision-camera/pull/325
https://github.com/mrousavy/react-native-vision-camera/issues/281

### DIFF
--- a/android/src/main/java/com/mrousavy/camera/CameraView.kt
+++ b/android/src/main/java/com/mrousavy/camera/CameraView.kt
@@ -410,9 +410,19 @@ class CameraView(context: Context, private val frameProcessorThread: ExecutorSer
         // User has selected a custom format={}. Use that
         val format = DeviceFormat(format!!)
         Log.i(TAG, "Using custom format - photo: ${format.photoSize}, video: ${format.videoSize} @ $fps FPS")
-        previewBuilder.setTargetResolution(format.videoSize)
+        
+        val targetPreviewResolution:Size
+        if (getDisplay().rotation == Surface.ROTATION_0) {
+          val newWidth = format.videoSize.height
+          val newHeight = format.videoSize.width
+          targetPreviewResolution = Size(newWidth, newHeight)
+        } else {
+          targetPreviewResolution = Size(format.videoSize.width, format.videoSize.height)
+        }
+
+        previewBuilder.setTargetResolution(targetPreviewResolution)
         imageCaptureBuilder.setTargetResolution(format.photoSize)
-        imageAnalysisBuilder.setTargetResolution(format.videoSize)
+        imageAnalysisBuilder.setTargetResolution(targetPreviewResolution)
 
         // TODO: Ability to select resolution exactly depending on format? Just like on iOS...
         when (min(format.videoSize.height, format.videoSize.width)) {
--- a/android/src/main/java/com/mrousavy/camera/CameraView.kt
+++ b/android/src/main/java/com/mrousavy/camera/CameraView.kt
@@ -411,19 +411,15 @@ class CameraView(context: Context, private val frameProcessorThread: ExecutorSer
         // User has selected a custom format={}. Use that
         val format = DeviceFormat(format!!)
         Log.i(TAG, "Using custom format - photo: ${format.photoSize}, video: ${format.videoSize} @ $fps FPS")
-        
-        val targetPreviewResolution:Size
-        if (getDisplay().rotation == Surface.ROTATION_0) {
-          val newWidth = format.videoSize.height
-          val newHeight = format.videoSize.width
-          targetPreviewResolution = Size(newWidth, newHeight)
-        } else {
-          targetPreviewResolution = Size(format.videoSize.width, format.videoSize.height)
+        val videoSize:Size
+        if (context.resources.configuration.orientation == Configuration.ORIENTATION_PORTRAIT) {
+          videoSize = Size(format.videoSize.height, format.videoSize.width)
+        }else {
+          videoSize = format.videoSize
         }
-
-        previewBuilder.setTargetResolution(targetPreviewResolution)
+        previewBuilder.setTargetResolution(videoSize)
         imageCaptureBuilder.setTargetResolution(format.photoSize)
-        imageAnalysisBuilder.setTargetResolution(targetPreviewResolution)
+        imageAnalysisBuilder.setTargetResolution(videoSize)
 
         // TODO: Ability to select resolution exactly depending on format? Just like on iOS...
         when (min(format.videoSize.height, format.videoSize.width)) {
--- a/android/src/main/java/com/mrousavy/camera/CameraView.kt
+++ b/android/src/main/java/com/mrousavy/camera/CameraView.kt
@@ -8,6 +8,7 @@ import android.content.res.Configuration
 import android.hardware.camera2.*
 import android.util.Log
 import android.util.Range
+import android.util.Size
 import android.view.*
 import android.view.View.OnTouchListener
 import android.widget.FrameLayout


mrousavy:
What happens when the user rotates his phone after the Camera has been initialized?

xulihang:
In my case, the camera still works in the specified resolution.

xulihang:
In the docs, it says that the resolution Size should be expressed in the coordinate frame after rotating the supported sizes by the target rotation: https://developer.android.com/reference/kotlin/androidx/camera/core/ImageAnalysis.Builder#setTargetResolution(android.util.Size)

ldstein:
FWIW I've been using a hack which shares similarities with what @xulihang proposes.

https://gist.github.com/ldstein/bf8b88fabc27617dca891aa3be58f9d7

Seemed to get me out of trouble for the limited devices I have on hand.

I noticed in this PR, `Surface.ROTATION_0` is being used to determine if the width and height should be reversed. This works for devices whose natural orientation is potrait but may have the opposite effect for devices which are landscape natural orientation (I say "may" because I have no such devices on hand to test).

An alternative would be to reference [context.resources.configuration.orientation](https://gist.github.com/ldstein/bf8b88fabc27617dca891aa3be58f9d7#file-react-native-vision-camera-2-12-0-patch-L21-L23) as this will inform us explicitly if the device is currently in Landscape or Portrait.

> What happens when the user rotates his phone after the Camera has been initialized?

As @xulihang mentions, it appears to work. Maybe under the hood it's already flipping it around when `targetRotation` is set during `updateOrientation()`?

I do wonder however if `imageAnalysis.targetRotation` and `preview.targetRotation` should both be set to `inputRotation` to keep them 1:1, but maybe that's a different topic.

xulihang:
> This works for devices whose natural orientation is potrait but may have the opposite effect for devices which are landscape natural orientation

In the docs: https://developer.android.com/training/camera2/camera-preview-large-screens#camera-orientation, it says that the camera sensor’s natural orientation is landscape.

ldstein:
In the above link it states:

> The [Android Compatibility Definition](https://source.android.com/compatibility/cdd) specifies that a camera image sensor “MUST be oriented so that the long dimension of the camera aligns with the screen’s long dimension.”

If you have an Android tablet which is naturally landscape orientated, this logic I think will incorrectly set the wrong resolution:

```kt
// Asume:
// Device Rotation: 0
// Device Screen Size: 1024x768
// Camera Sensor Orientation: Landscape (same as screen)
// format.videoSize.width:1024
// format.videoSize.height: 768

if (getDisplay().rotation == Surface.ROTATION_0) {
          val newWidth = format.videoSize.height // 768
          val newHeight = format.videoSize.width // 1024
          targetPreviewResolution = Size(newWidth, newHeight) // resolution is incorrectly set to portrait
```

I think that's why [my original hack](https://github.com/mrousavy/react-native-vision-camera/issues/281#issuecomment-920690010) fixed the resolution for phones, but did not work on @viljark['s TB-X505L tablet](https://github.com/mrousavy/react-native-vision-camera/issues/281#issuecomment-993952219).

`format.videoSize` appears to always be reported in landscape, so I think we only need to set a portrait targetResolution when the device is in portrait:

```kt
// Asume:
// Device Rotation: 0
// Device Screen Size: 1024x768
// Camera Sensor Orientation: Landscape (same as screen)
// format.videoSize.width:1024
// format.videoSize.height: 768

val videoSize = when (context.resources.configuration.orientation) {
    Configuration.ORIENTATION_PORTRAIT -> Size(format.videoSize.height, format.videoSize.width) // resolution is set to portrait
    else -> format.videoSize // resolution is set to landscape
}
```

Having said all this, I'm no expert, so happy to be proven wrong :)

xulihang:
Using orientation seems a better solution. I also find a page on detecting the natural orientation: http://www.java2s.com/example/android/user-interface/get-device-natural-orientation.html

xulihang:
I've updated the code to use orientation.

android/src/main/java/com/mrousavy/camera/CameraView.kt
```diff
@@ -410,9 +411,15 @@ class CameraView(context: Context, private val frameProcessorThread: ExecutorSer
         // User has selected a custom format={}. Use that
         val format = DeviceFormat(format!!)
         Log.i(TAG, "Using custom format - photo: ${format.photoSize}, video: ${format.videoSize} @ $fps FPS")
-        previewBuilder.setTargetResolution(format.videoSize)
+        val videoSize:Size
+        if (context.resources.configuration.orientation == Configuration.ORIENTATION_PORTRAIT) {
```
mrousavy, line 415: Hey, thanks again for the PR.

Unfortunately I am still not quite sure if I understand this correctly, doesn't this only check the "locked" Orientation in AndroidManifest? What if the user is holding his phone in portrait, then going to landscape? And then going back to portrait? Is it still the correct resolution?

xulihang: If I read the docs correctly, `context.resources.configuration.orientation` returns the device's current orientation.

mrousavy: Okay, but can we use `outputOrientation` instead of `context.resources.configuration.orientation` here? I already detect orientation in this class

ldstein: `outputRotation` is referencing `context.displayRotation`. That tells us how many degrees the device is rotated from it's natural position but I believe we cannot assume zero rotation is portrait.

`context.resources.configuration.orientation` differs as it tells us explicitly whether the device is currently being viewed in portrait or landscape.

Similarly, there maybe an issue here in the outputRotation getter. Portrait would only equal Surface.ROTATION_0 for naturally-portrait devices:

```kotlin
  private val outputRotation: Int
    get() {
      if (orientation != null) {
        // user is overriding output orientation
        return when (orientation!!) {
          "portrait" -> Surface.ROTATION_0
          "landscapeRight" -> Surface.ROTATION_90
          "portraitUpsideDown" -> Surface.ROTATION_180
          "landscapeLeft" -> Surface.ROTATION_270
 ```
 
 Also not 100% sure on this, but maybe `imageAnalysis.targetRotation` should be the same as `preview.targetRotation` ? Keeps the analysis results 1:1 with the preview (when drawing overlays, for example).
 
I can see the use of the 'orientation' prop to affect photo / video output, but maybe should be independent to modifying the frame processor rotation.

xulihang: I cannot find `outputOrientation`. There is `orientation` which is a `string` and `outputRotation` which is an integer.

ldstein: Sorry, typo on my end. I fixed up my earlier post. I meant `outputRotation`

mrousavy: Okay guys thanks for the insights. I will test this some time this week and then merge the PR if everything works in every use-case :)

Would be helpful if anyone could help me with testing, with various devices and various orientations (e.g. start with landscape or portrait, then rotate)

siddharthkr: @mrousavy @xulihang What are your views on this PR, is it ready to get merged or ...?

bekatd:
Camera formats and everything related to it are are completely wrong. 

- width and heights are swapped,
- unable to set desired formats,
- also `photoWidth` and `photoHeight` of `useCameraFormat` hook are totally wrong, since the width and height of frame in frameprocessor plugin is for example 640x640 while current cameraformat is 2000x4000 :| 
- also camera feed strangely zooms on different formats
- also if you log all available formats and look at the format returned by default, you will discover that it can't be found in the available formats :|
- also on every different device everything works differently and wrongly :|

how all of this is even possible :|

mrousavy:
Hey, thanks for your PR and sorry for the long radio silence here! width and height are swapped because Camera sensors are in 90deg orientation. Orientation is now tracked in this issue: https://github.com/mrousavy/react-native-vision-camera/issues/1891

PR #842, opened 2022-02-21
Add OCR frame processor plugin

aarongrider:
Replace existing, unimplemented frame processor plugin with new one.


mrousavy:
Hey thanks for the PR! Let's not remove @rodgomesc's plugin but instead only add your's directly below that

rodgomesc:
hi i think we can remove my plugin, i still don't have time to implement this

rodgomesc:
sorry for that guys, i’ll keep maintaining my other plugins :)

mrousavy:
Cool, then LGTM!

PR #958, opened 2022-03-31
Fix Android focus not using correct focus point

### LINKED ISSUE #758: 🐛 Focus on tap

Hello,

I am trying to simulate focus on tap, but the location where I tapped is newer focussed correctly.

Fresh project - ("react-native": "0.66.4", "react-native-vision-camera": "^2.12.0", ...android device)
Locked portrait orientation,
Camera component has marginTop:60, marginLeft:60 to better check of tap coordinates.
I am using coordinates relative to Camera component rectangle as mentiond here https://mrousavy.com/react-native-vision-camera/docs/guides/focusing/

I made demo project where this problem is ilustrated.
https://github.com/TasigurIshtar/RNVisionCameraTest

Thanks for your time.

### Reproduceable Code

```tsx
const touchII = async (event: NativeTouchEvent) => {
    let point: Point = {
      x: Math.round(event.pageX - camLocation.x),
      y: Math.round(event.pageY - camLocation.y),
    };
    await cameraRef?.current?.focus(point)     
  };

.....
 <Camera
      ref={cameraRef}
      style={{
        marginTop: 60,
        marginLeft: 60,
        width: window.width - 60,
        height: window.height - 60,
      }}
      onLayout={event => {
        const layout = event.nativeEvent.layout;
        setCamLocation({x: layout.x, y: layout.y});
      }}
      device={device}
      isActive={true}
      onTouchEnd={x => device.supportsFocus && touchII(x.nativeEvent)}
    />
```

### What happened instead?

after tapping to the top of the bottle .focus({"x": 53, "y": 393})  -> the bottle doesn't seems to be focused.  _(when I was tapping more on the right site around the grayish mat, occasionaly the top of the bottle was super sharp. )_

image

### Relevant log output

_No response_

### Device

Samsung Galaxy S10

### VisionCamera Version

2.12.0

### Additional information

- [ ] I am using Expo

Menardi:
## What

This PR fixes the `focus` function on Android not focusing on the given point (#758). The issue is that when calculating the focus point with `SurfaceOrientedMeteringPointFactory`, it does not factor in any scaling or cropping that has been done to the camera's view (thanks to [this StackOverflow answer](https://stackoverflow.com/a/60585382) for pointing me in the right direction!). I think it's possible this issue is not obvious unless you're resizing the camera view significantly in some way.

## Changes

Rather than manually calculate the focus point, we can get the PreviewView to do it for us. This fixes the issue because the PreviewView factors in any scaling or resizing of the view on the screen. The only potential issue is that this needs to run on the UI thread (which is what the `withContext` is doing), but I've tested it with frame processors enabled and disabled, and have found no issues in either case.

## Tested on

Tested on Pixel 3a running Android 12

## Related issues

Fixes #758

### DIFF
--- a/android/src/main/java/com/mrousavy/camera/CameraView+Focus.kt
+++ b/android/src/main/java/com/mrousavy/camera/CameraView+Focus.kt
@@ -1,9 +1,9 @@
 package com.mrousavy.camera
 
 import androidx.camera.core.FocusMeteringAction
-import androidx.camera.core.SurfaceOrientedMeteringPointFactory
 import com.facebook.react.bridge.ReadableMap
 import kotlinx.coroutines.guava.await
+import kotlinx.coroutines.withContext
 import java.util.concurrent.TimeUnit
 
 suspend fun CameraView.focus(pointMap: ReadableMap) {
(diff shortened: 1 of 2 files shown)


mrousavy:
Hey, thanks so much for your PR! I think everything looks good, LGTM!

Menardi:
Thanks for the quick review! 😄

PR #991, opened 2022-04-12
fix: Adds Setting compileSdkVersion to 31 in ExpoPlugin

thorbenprimke:
## What

In order for a managed Expo project to build successfully with Expo EAS,
the min compile sdk version value needs to be set to 31.
Default value in Expo 44 is 30.

## Changes

Enhances the config plugin to set the min compile sdk value.

## Tested on
1) 
- Patched in changes logically
- Ran `expo prebuild -p android`
- Verified that value is set as expected

2) 
- Patched in changes logically
- Kicked of Expo EAS build
- Build succeeded

### DIFF
--- a/src/expo-plugin/withVisionCamera.ts
+++ b/src/expo-plugin/withVisionCamera.ts
@@ -31,6 +31,11 @@ const withCamera: ConfigPlugin<Props> = (config, props = {}) => {
     config = withDisableFrameProcessorsIOS(config);
   }
 
+  config = AndroidConfig.Version.withBuildScriptExtMinimumVersion(config, {
+    name: 'compileSdkVersion',
+    minVersion: 31,
+  });
+
   return withPlugins(config, [[AndroidConfig.Permissions.withPermissions, androidPermissions]]);
 };
 


mrousavy:
Hi! Thanks for your PR!

I'm a bit worried that this is out of scope for my library though.

1. What happens if the user set the minCompileSdkVersion to `32`? Will this override it and set it back to `31`?
2. What happens if another library sets it to `32` and my lib sets it to `31`?
3. What happens if the user does not have `31` yet, uses my plugin so it sets it to `31`, and then the build fails because since `31` you need to have `android:exported="true"` on your `<activity>` in `AndroidManifest.xml`?

I'm kinda more biased towards the users making sure they are on 31 themselves and using a separate plugin to achieve that, after all in a few months when every Expo app is using `31` this is completely useless anyways.

wdyt?

hirbod:
When I added the Frame Processor exclusion plugin, I had that part built in, you asked me to remove it. I think your point was valid and still is. This should not be the scope of this library. But I think we should link a Gist file for a minSdk bumper plugin inside the docs.

thorbenprimke:
@mrousavy, @hirbod - Thanks for the context and feedback. 

Yes, I thought it would be a good add to the config plugin because the 31 requirement stems from the `androidx.camera` dependencies. 

In regards to your questions:

> 1) What happens if the user set the minCompileSdkVersion to 32? Will this override it and set it back to 31?
> 2) What happens if another library sets it to 32 and my lib sets it to 31?

[It is only applied if the current value is smaller.](https://github.com/expo/expo-cli/blob/main/packages/config-plugins/src/android/Version.ts#L22) No change if another config plugin set it to 32.

> 3) What happens if the user does not have 31 yet, uses my plugin so it sets it to 31, and then the build fails because since 31 you need to have android:exported="true" on your <activity> in AndroidManifest.xml?

Ah. I didn't know about this detail - I guess I hadn't run into it. I checked the [template file in Expo](https://github.com/expo/expo/blob/main/template-files/android/AndroidManifest.xml) and all the activities do have it set to true, so I don't think this should cause a problem since the config plugin applies to managed Expo projects only (or at the time one ejects/runs prebuild).

---

Yes, I suppose this will not be required anymore once the next version of Expo drops. In the meantime though, it could save someone trouble when integrating vision camera and building a managed Expo project with EAS.

Perhaps if you prefer not to include this in the config plugin, a note could be added to the Getting Started guide under Managed Expo.

mrousavy:
Yea I think I'll add it to the TROUBLESHOOTING guide, after all I don't want Getting Started to be messy. Any added steps or details there could confuse the user, I mean some READMEs are just way too complicated (e.g. react-native-vector-icons, I still don't understand how to correctly install it)

thorbenprimke:
Sounds good. Thanks for the review and thoughts.

PR #995, opened 2022-04-14
fix: AVAssetWriter status is 1 crash

### LINKED ISSUE #930: 🐛 [ios] Crash report - Cannot call method when status is 1

My user was recording multiple videos. So a bunch of startRecording / stopRecording calls.
From what I can tell, this issue is related to older iPhone devices. My user is using an iPhone SE 2nd Generation with iOS 15.4

The App crashes, heats up a lot and also screen dimming is happening.
It is kinda hard to tell what is happening, I just have those crash reports. I never had a crash on my iPhone 13 Pro Max, guess it has more CPU, GPU, RAM etc and can handle frequent start/stop triggers.

### Reproduceable Code

_No response_

### What happened instead?

The app crashes with `Cannot call method when status is 1`

Bildschirmfoto 2022-03-24 um 13 00 27

### Relevant log output

```shell
NSInternalInconsistencyException
*** -[AVAssetWriter finishWritingWithCompletionHandler:] Cannot call method when status is 1
```

### Device

iPhone SE 2nd Generation (15.4)

### VisionCamera Version

2.13.0

### Additional information

- [ ] I am using Expo

hirbod:
## What

`markAsFinished` has to be called within the finishWritingWithCompletionHandler.
Moving that 2 lines allowed me to spam start / stop for hundreds of time without a single crash.

Since I am not a native developer, I can only assume that what I've done is correct. I had no regressions and the issue, which was 100% reproducible every single time on older devices due to performance I guess, has not happened again after applying this little change.

## Changes

Moved markedAsFinished into finishWriting.
This article was helpful to track the issue:
https://code-examples.net/en/q/e14f33

## Tested on

- iPhone 7, iOS 12.1
- iPhone 12 Pro Max, 15.4.1
- iPhone 13 Pro Max, 15.4.1

## Related issues

Fixes #930

### DIFF
--- a/ios/RecordingSession.swift
+++ b/ios/RecordingSession.swift
@@ -199,9 +199,9 @@ class RecordingSession {
                           userInfo: [NSLocalizedDescriptionKey: "Stopped Recording Session too early, no frames have been recorded!"])
       completionHandler(.failed, error)
     } else if assetWriter.status == .writing {
-      bufferAdaptor?.assetWriterInput.markAsFinished()
-      audioWriter?.markAsFinished()
       assetWriter.finishWriting {
+        self.bufferAdaptor?.assetWriterInput.markAsFinished()
+        self.audioWriter?.markAsFinished()
         self.completionHandler(self.assetWriter.status, self.assetWriter.error)
       }
     } else {


mrousavy:
Does it still stop the video at the correct point?

mrousavy:
Cool, I think this is a good change. Thanks @hirbod 🖤, LGTM

hirbod:
@mrousavy I was so happy... but unfortunately the crash happened again with this fix. Looks like less frequent, but still possible. 

You already had concerns about the correct stop position, it might be better to revert this, since I tried this into the blue based on an article. There is basically 0 documentation about this error 

So some informations I could find:

Enums:

 AVAssetWriterStatusUnknown = 0,
 AVAssetWriterStatusWriting = 1,
 AVAssetWriterStatusCompleted = 2,
 AVAssetWriterStatusFailed = 3,
 AVAssetWriterStatusCancelled = 4
 
So if I understand that correctly, status 1 means that we still trying to write something while status is 1?
This explains why this only happens when I spam start/stop

hirbod:
So some informations I could find:

Enums:

 AVAssetWriterStatusUnknown = 0,
 AVAssetWriterStatusWriting = 1,
 AVAssetWriterStatusCompleted = 2,
 AVAssetWriterStatusFailed = 3,
 AVAssetWriterStatusCancelled = 4
 
So if I understand that correctly, status 1 means that we still trying to write something while status is 1?
This explains why this only happens when I spam start/stop

hirbod:
@mrousavy I think I could isolate the issue a bit more:

```
2022-04-15 12:13:02.933455+0200 app[840:267779] [native] VisionCamera.finish(): Finishing Recording with AssetWriter status "writing"...
2022-04-15 12:13:02.940310+0200 app[840:267779] [native] VisionCamera.finish(): Finishing Recording with AssetWriter status "writing"...
2022-04-15 12:13:02.942954+0200 app[840:267779] DevLauncher tries to handle uncaught exception: *** -[AVAssetWriter finishWritingWithCompletionHandler:] Cannot call method when status is 1
```

The crash only happens, when finish get called twice (which shouldn't but somehow can happen). 
So spamming fast enough can get you into this situation.


Edit: I think I have a solid fix ready now.

PR #996, opened 2022-04-15
fix: AVWriter status 1 crash - follow up

hirbod:
This is a follow up for #995 and #930

My last PR did not fix the issue and I found the main cause. 

I realized this by watching the logs:

```
2022-04-15 12:13:02.933455+0200 app[840:267779] [native] VisionCamera.finish(): Finishing Recording with AssetWriter status "writing"...
2022-04-15 12:13:02.940310+0200 app[840:267779] [native] VisionCamera.finish(): Finishing Recording with AssetWriter status "writing"...
2022-04-15 12:13:02.942954+0200 app[840:267779] DevLauncher tries to handle uncaught exception: *** -[AVAssetWriter finishWritingWithCompletionHandler:] Cannot call method when status is 1
```
When you spam start/stop, you might end up calling `finish()` twice. I don't know exactly how it can happen, but it does happen. 

So the real solution would be to prevent calling finish twice in the first place, but this little boolean check is sufficient to prevent the crash. This time, I've tested it more intensively. 

After spamming start/stop to generate nearly 400 video fragments (you can do really cool effects with something like that), I ended up with that amount of entries

```
2022-04-15 12:45:34.766416+0200 app[990:282220] [native] VisionCamera.finish(): Detected two finish() calls on the same AVWriter single-use object, returning to prevent crash
2022-04-15 12:45:36.489299+0200 app[990:282130] [native] VisionCamera.finish(): Detected two finish() calls on the same AVWriter single-use object, returning to prevent crash
2022-04-15 12:45:40.158120+0200 app[990:282817] [native] VisionCamera.finish(): Detected two finish() calls on the same AVWriter single-use object, returning to prevent crash
2022-04-15 12:45:43.825698+0200 app[990:282528] [native] VisionCamera.finish(): Detected two finish() calls on the same AVWriter single-use object, returning to prevent crash
2022-04-15 12:45:52.602799+0200 app[990:283148] [native] VisionCamera.finish(): Detected two finish() calls on the same AVWriter single-use object, returning to prevent crash
2022-04-15 12:45:53.110238+0200 app[990:282130] [native] VisionCamera.finish(): Detected two finish() calls on the same AVWriter single-use object, returning to prevent crash
2022-04-15 12:45:57.600504+0200 app[990:282130] [native] VisionCamera.finish(): Detected two finish() calls on the same AVWriter single-use object, returning to prevent crash
2022-04-15 12:45:58.759749+0200 app[990:282528] [native] VisionCamera.finish(): Detected two finish() calls on the same AVWriter single-use object, returning to prevent crash
2022-04-15 12:47:17.176342+0200 app[990:282168] [native] VisionCamera.finish(): Detected two finish() calls on the same AVWriter single-use object, returning to prevent crash
2022-04-15 12:47:40.303175+0200 app[990:283911] [native] VisionCamera.finish(): Detected two finish() calls on the same AVWriter single-use object, returning to prevent crash
```

So I effectively prevent the app from crashing. I know from beta testers, that this happened to them  without even trying too hard.

### DIFF
--- a/ios/RecordingSession.swift
+++ b/ios/RecordingSession.swift
@@ -197,7 +197,7 @@ class RecordingSession {
       ReactLogger.log(level: .warning, message: "Tried calling finish() twice while AssetWriter is still writing!")
       return
     }
-    if !isFinishing {
+    if !hasWrittenFirstVideoFrame {
       let error = NSError(domain: "capture/aborted",
                           code: 1,
                           userInfo: [NSLocalizedDescriptionKey: "Stopped Recording Session too early, no frames have been recorded!"])
--- a/ios/RecordingSession.swift
+++ b/ios/RecordingSession.swift
@@ -193,12 +193,12 @@ class RecordingSession {
    */
   func finish() {
     ReactLogger.log(level: .info, message: "Finishing Recording with AssetWriter status \"\(assetWriter.status.descriptor)\"...")
-    
+
     if isFinishing {
       ReactLogger.log(level: .warning, message: "Tried calling finish() twice while AssetWriter is still writing!")
       return
     }
-    
+
     if !hasWrittenFirstVideoFrame {
       let error = NSError(domain: "capture/aborted",
                           code: 1,
--- a/ios/RecordingSession.swift
+++ b/ios/RecordingSession.swift
@@ -33,7 +33,7 @@ class RecordingSession {
   private var initialTimestamp: CMTime?
   private var latestTimestamp: CMTime?
   private var hasWrittenFirstVideoFrame = false
-  private var hasRunningWritingAttempt = false;
+  private var hasRunningWritingAttempt = false
 
   var url: URL {
     return assetWriter.outputURL
@@ -208,7 +208,7 @@ class RecordingSession {
       bufferAdaptor?.assetWriterInput.markAsFinished()
       audioWriter?.markAsFinished()
       assetWriter.finishWriting {
-        self.hasRunningWritingAttempt = false;
+        self.hasRunningWritingAttempt = false
         self.completionHandler(self, self.assetWriter.status, self.assetWriter.error)
       }
     } else {
--- a/ios/RecordingSession.swift
+++ b/ios/RecordingSession.swift
@@ -207,7 +207,7 @@ class RecordingSession {
       bufferAdaptor?.assetWriterInput.markAsFinished()
       audioWriter?.markAsFinished()
       assetWriter.finishWriting {
-        self.hasRunningWritingAttempt = false
+        self.isFinishing = false
         self.completionHandler(self, self.assetWriter.status, self.assetWriter.error)
       }
     } else {
--- a/ios/RecordingSession.swift
+++ b/ios/RecordingSession.swift
@@ -193,10 +193,12 @@ class RecordingSession {
    */
   func finish() {
     ReactLogger.log(level: .info, message: "Finishing Recording with AssetWriter status \"\(assetWriter.status.descriptor)\"...")
+    
     if isFinishing {
       ReactLogger.log(level: .warning, message: "Tried calling finish() twice while AssetWriter is still writing!")
       return
     }
+    
     if !hasWrittenFirstVideoFrame {
       let error = NSError(domain: "capture/aborted",
                           code: 1,
--- a/ios/RecordingSession.swift
+++ b/ios/RecordingSession.swift
@@ -33,6 +33,7 @@ class RecordingSession {
   private var initialTimestamp: CMTime?
   private var latestTimestamp: CMTime?
   private var hasWrittenFirstVideoFrame = false
+  private var hasRunningWritingAttempt = false;
 
   var url: URL {
     return assetWriter.outputURL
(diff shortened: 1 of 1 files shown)


mrousavy:
You are never setting it to true.

hirbod:
I do, line 208

hirbod:
@mrousavy  good catch. I actually did set it to `true` but only in my local build.
I did not push push from a repo but used the Github editor and forgot to add it. I fixed it!

hirbod:
Do you think my fix is ok or would you like to go another route?

mrousavy:
After reading through this a couple of times I think I want to go another route.

The problem here is that an additional binary state value can be out of sync because it's not volatile and we're dealing with concurrency here. There might be cases where writing is happening but we are not aware of it and therefore can't stop it anymore.

The ideal solution to this problem would be to make the Queue serial (I think it should be already?) and have the result here be blocking. We can delegate this blocking up to JS so JS knows when it's recording frames and avoid calling `stop` if we are currently recording. I think this would be the ideal way, I'll investigate this more with Fabric/TurboModules migration.

Thank you for your PR though @hirbod, this is definitely a complicated issue.

hirbod:
Will migrating force us to enable Fabric? This would be a blocker, since I am pretty sure that the vast majority will not be able to enable Fabric within the next 1-2 years. 

I was not able to detect any side effects with this fix, and it was only reproducible on older devices. Devices with more CPU/GPU etc never printed this `log` line.

mrousavy:
> since I am pretty sure that the vast majority will not be able to enable Fabric in the next 1-2 years.

disagree

> I was not able to detect any side effects with this fix

well obviously this takes a lot of time and testing to find edge cases. The code, as is, is able to reproduce states that are out of sync. Maybe we can use an atomic boolean

mrousavy:
btw. - the CI is failing anyways - you need to read the CONTRIBUTING guidelines and make sure you ran the Swift formatter

ios/RecordingSession.swift
```diff
@@ -33,6 +33,7 @@ class RecordingSession {
   private var initialTimestamp: CMTime?
   private var latestTimestamp: CMTime?
   private var hasWrittenFirstVideoFrame = false
+  private var hasRunningWritingAttempt = false
```
mrousavy: Rename to `isFinishing`

ios/RecordingSession.swift
```diff
@@ -193,15 +194,22 @@ class RecordingSession {
   func finish() {
     ReactLogger.log(level: .info, message: "Finishing Recording with AssetWriter status \"\(assetWriter.status.descriptor)\"...")
 
+    if(hasRunningWritingAttempt) {
```
mrousavy: fix lint errors, read contributing guidelines.

ios/RecordingSession.swift
```diff
@@ -193,15 +194,22 @@ class RecordingSession {
...
+      ReactLogger.log(level: .info, message: "Detected two finish() calls on the same AVWriter single-use object, returning to prevent crash")
+      return
+    }
+
```
mrousavy: fix lint errors, read contributing guidelines.

ios/RecordingSession.swift
```diff
@@ -193,15 +194,22 @@ class RecordingSession {
   func finish() {
     ReactLogger.log(level: .info, message: "Finishing Recording with AssetWriter status \"\(assetWriter.status.descriptor)\"...")
 
+    if(hasRunningWritingAttempt) {
+      ReactLogger.log(level: .info, message: "Detected two finish() calls on the same AVWriter single-use object, returning to prevent crash")
```
mrousavy: Make error message say `"Tried calling `finish()` twice while AssetWriter is still writing!"` and make it a `.warning` or `.error`

hirbod:
Ok, done. And your linter is happy now :)
Sorry for NOT reading the contribution guide.

ios/RecordingSession.swift
```diff
@@ -192,16 +193,21 @@ class RecordingSession {
    */
   func finish() {
     ReactLogger.log(level: .info, message: "Finishing Recording with AssetWriter status \"\(assetWriter.status.descriptor)\"...")
-
+    if isFinishing {
+      ReactLogger.log(level: .warning, message: "Tried calling finish() twice while AssetWriter is still writing!")
+      return
+    }
```
mrousavy, line 200: can you add a new line before and after the `if`? Give it room to breathe

mrousavy (changes requested):

mrousavy:
sorry just a minor codestyle nitpick

ios/RecordingSession.swift
hirbod: There was whitespace before, but the linter complained :D

hirbod:
You see.. I added whitespace and the job is failing now :D

ios/RecordingSession.swift
hirbod: ```Trailing Whitespace Violation: Lines should not have trailing whitespace. (trailing_whitespace)```

mrousavy: What happens if you run `yarn check-ios` in the root? I think it is complaining about the line having one whitespace character (`\n `) instead of just the new-line (`\n`).

hirbod: Yeah - that was the reason. Pretty restrictive

hirbod:
Ok, linting passes. I hope thats fine now

hirbod:
@mrousavy anything holding back?

mrousavy:
Thank you for your PR!

Ideally I'd love to not have a separate boolean managed by my lib, but instead check that on the `assetWriter` instance directly. But apparently the `.status` prop does not expose if it's currently finishing, and the Thread is not blocking either, which is weird but I guess that's the only solution 🤷

Issue #973, opened 2022-04-07
🐛 Video pausing does not work correctly on iOS


DragonSpirit:
### What were you trying to do?

Just call pauseRecording and resumeRecording() after start of recording

### Reproduceable Code

```tsx
It could be reproduced on example app
```

### What happened instead?

The video gets stuck on the last frame between pause and resume, without shortening the gap between them. Audio does not work correctly either.
Android works like a charm.

### Device

iPhone 11, iOS 15.4.1

### VisionCamera Version

2.13.1

### Additional information

- [ ] I am using Expo
- [X] I have read the [Troubleshooting Guide](https://mrousavy.github.io/react-native-vision-camera/docs/guides/troubleshooting)
- [X] I agree to follow this project's [Code of Conduct](https://github.com/mrousavy/react-native-vision-camera/blob/main/CODE_OF_CONDUCT.md)
- [X] I searched for [similar issues in this repository](https://github.com/mrousavy/react-native-vision-camera/issues) and found none.

itershukov:
Have the same issue on iPhone 7

itershukov:
Is there any news?

mrousavy:
Closing as this is a stale issue - this might have been fixed with the full rewrite in VisionCamera V3 (🥳) - if not, please create a new issue.

PR #1051, opened 2022-05-17
fix: Pause on iOS not working correctly

DragonSpirit:
## What

This PR fixes a gap while video on pause

## Changes

This PR stores the difference between the last timestamp and the timestamp when the video was paused. Subsequently, the diff is taken to add new frames.

## Tested on
iPhone 11, iOS 15.4

## Related issues

#973 

## Additional

It works correctly in case of video, but I still have an issue with sound. Any help would be appreciated.

### DIFF
--- a/ios/CameraView+RecordVideo.swift
+++ b/ios/CameraView+RecordVideo.swift
@@ -152,7 +152,9 @@ extension CameraView: AVCaptureVideoDataOutputSampleBufferDelegate, AVCaptureAud
   func stopRecording(promise: Promise) {
     cameraQueue.async {
       self.isRecording = false
-
+      self.isPausing = false
+      self.totalDiff = CMTime.zero
+    
       withPromise(promise) {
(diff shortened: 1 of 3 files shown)


mrousavy:
Phew, this looks pretty complicated. Wouldn't it be possible to just add a single new variable that keeps track of the latest recorded timestamp when pause was initiated? Maybe we can abstract this into a separate object to avoid having everything in the CameraView...

DragonSpirit:
In the case of storing the last recorded timestamp, we may encounter the problem of multiple pauses in the same video.
I dunno a simpler way when we use a timestamp to add in a buffer.

kevyg03:
Any progress on this PR? Would be a huge help.

mrousavy:
Hey thanks for your PR and sorry for the long radio silence here! I think this approach is a bit messy, I wonder why AVAssetWriter doesn't properly work with the timestamp difference when paused. Maybe we should use another master clock?

PR #1075, opened 2022-06-03
fix: Asset Writer Video-Audio Sync

thomas-coldwell:
## What

This PR fixes an issue with the AV Asset Writer video and audio becoming desynchronised. Currently the asset writing session was started with a source time of just when it was called rather than the presentation time of the first video frame as should be the case described in Apple's docs: https://developer.apple.com/documentation/avfoundation/avassetwriter/1389908-startsession

## Changes

- Adds a new flag for checking if the asset writer session has started `hasStartedWritingSession`
- Calls the `assetWriter.startSession()` when the first video frame gets appended instead of being called directly after `assetWriter.startWriting()`

## Tested on

- iPhone 13 Pro

### DIFF
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -324,7 +324,7 @@ PODS:
     - React
   - RNVectorIcons (8.1.0):
     - React-Core
-  - VisionCamera (2.13.0):
+  - VisionCamera (2.13.3):
     - React
     - React-callinvoker
     - React-Core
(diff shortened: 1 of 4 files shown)


hirbod:
I just came here to report that I've encountered my videos being async to audio. You're my saviour! I just pulled your changes in and will see if audio-video sync issues will be fixed afterwards

hirbod:
I can confirm that the desync issue is gone!

PR #1082, opened 2022-06-06
feat: Support RAW photo capture on iOS

### LINKED ISSUE #73: ✨ RAW Photo Capture
Currently you can only take .jpg photos. There are use-cases where you'd also want to take photos in the camera's RAW format for additional post-processing.

This should be configurable by setting a `raw: true` parameter in the `takePhoto(...)` func

* [`AVCapturePhotoSettings::init(rawPixelFormatType:)`](https://developer.apple.com/documentation/avfoundation/avcapturephotosettings/1648662-init)

This would use Apple ProRAW on iOS, and RAW YUV on Android.

WWDC resource: https://developer.apple.com/videos/play/wwdc2021/10160/

ladypinzon25:
## What

This PR adds support for capturing photos with RAW format on iOS.

## Changes

* (iOS) Adds `enableRawCapture` boolean prop available in iOS.
* (iOS) Finds a `rawFormat` for the `AVCapturePhotoSettings`
* (iOS) Adds a delegate file called `RAWCaptureDelegate` which is responsible for:
Making a unique DNG file URL.
Saving the photo in the phone gallery. I wanted to save the photo to the phone with the `CameraRoll.save()` function in the react native example, but the **CameraRoll library** doesn’t support saving `dng` files to the phone. Currently, there is a PR for saving `dng` files to the CameraRoll library: https://github.com/react-native-cameraroll/react-native-cameraroll/pull/223/files
Return the photo data: `path`, `width`, `height`, `isRawPhoto`, `metadata`, `thumbnail`
* (iOS) Adds a raw embedded thumbnail photo format to the `photoSettings`

## Tested on

* iPhone 13 pro max, iOS 15.4.1

## Related issues
* Fixes #73

### DIFF
--- a/ios/CameraView+AVCaptureSession.swift
+++ b/ios/CameraView+AVCaptureSession.swift
@@ -61,6 +61,10 @@ extension CameraView {
       }
     }
 
+    if enableRawCapture?.boolValue == true {
+      captureSession.sessionPreset = .photo
+    }
+
     // pragma MARK: Capture Session Inputs
     // Video Input
     do {
@@ -115,6 +119,9 @@ extension CameraView {
         return
       }
       captureSession.addOutput(photoOutput!)
+      if #available(iOS 14.3, *), enableRawCapture?.boolValue == true {
+        photoOutput!.isAppleProRAWEnabled = photoOutput!.isAppleProRAWSupported
+      }
       if videoDeviceInput!.device.position == .front {
         photoOutput!.mirror()
       }
@@ -233,21 +240,40 @@ extension CameraView {
       return
     }
 
-    // get matching format
-    let matchingFormats = device.formats.filter { $0.matchesFilter(filter) }.sorted { $0.isBetterThan($1) }
-    guard let format = matchingFormats.first else {
-      invokeOnError(.format(.invalidFormat))
-      return
-    }
+    if #available(iOS 14.3, *), enableRawCapture?.boolValue == true {
+      guard let format = device.formats.first(where: { $0.isHighestPhotoQualitySupported }) else {
+        // handle failure to find a format that supports highest quality stills
+        invokeOnError(.format(.invalidFormat))
+        return
+      }
 
-    do {
-      try device.lockForConfiguration()
-      device.activeFormat = format
-      device.unlockForConfiguration()
-      ReactLogger.log(level: .info, message: "Format successfully configured!")
-    } catch let error as NSError {
-      invokeOnError(.device(.configureError), cause: error)
-      return
+      do {
+        try device.lockForConfiguration()
+        device.activeFormat = format
+        device.unlockForConfiguration()
+        ReactLogger.log(level: .info, message: "Format successfully configured!")
+      } catch let error as NSError {
+        invokeOnError(.device(.configureError), cause: error)
+        return
+      }
+    } else {
+      // get matching format
+      let matchingFormats = device.formats.filter { $0.matchesFilter(filter) }.sorted { $0.isBetterThan($1) }
+
+      guard let format = matchingFormats.first else {
+        invokeOnError(.format(.invalidFormat))
+        return
+      }
+
+      do {
+        try device.lockForConfiguration()
+        device.activeFormat = format
+        device.unlockForConfiguration()
+        ReactLogger.log(level: .info, message: "Format successfully configured!")
+      } catch let error as NSError {
+        invokeOnError(.device(.configureError), cause: error)
+        return
+      }
     }
   }
 
--- a/ios/CameraView+TakePhoto.swift
+++ b/ios/CameraView+TakePhoto.swift
@@ -54,14 +54,49 @@ extension CameraView {
       }
 
       // Create photo settings
-      let photoSettings = AVCapturePhotoSettings(format: format)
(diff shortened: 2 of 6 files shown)


ios/CameraView+AVCaptureSession.swift
```diff
@@ -233,21 +240,40 @@ extension CameraView {
...
-      return
+      do {
+        try device.lockForConfiguration()
+        device.activeFormat = format
+        device.unlockForConfiguration()
+        ReactLogger.log(level: .info, message: "Format successfully configured!")
+      } catch let error as NSError {
+        invokeOnError(.device(.configureError), cause: error)
+        return
+      }
```
mrousavy, line 258: Why do we need to repeat ourselves here? Why does `enableRawCapture` follow a different code path? Can't we just use the same code (with the `matchingFormats` sorter func and all that)?

ios/CameraView+TakePhoto.swift
```diff
@@ -54,29 +54,64 @@ extension CameraView {
...
+          fatalError("No RAW format found.")
+        }
+
+        let processedFormat = [AVVideoCodecKey: AVVideoCodecType.hevc]
+
+        photoSettings = AVCapturePhotoSettings(rawPixelFormatType: rawFormat,
+                                               processedFormat: processedFormat)
+      } else {
+        photoSettings = AVCapturePhotoSettings(format: format)
+      }
```
mrousavy, line 75: Could we maybe wrap that in a separate class? I don't want to cloud up this file with such nested and very specific code. 
Additionally, `photoSettings` is now optional and we're force unwrapping it down there, so this is prone to null unwraps.

ios/CameraView+TakePhoto.swift
```diff
@@ -54,29 +54,64 @@ extension CameraView {
...
+          return
+        }
+
+        let dimensions = videoDeviceInput.device.activeFormat.highResolutionStillImageDimensions
+
+        photoSettings!.rawEmbeddedThumbnailPhotoFormat = [
+          AVVideoCodecKey: thumbnailPhotoCodecType,
+          AVVideoWidthKey: dimensions.width,
+          AVVideoHeightKey: dimensions.height,
+        ]
```
mrousavy, line 99: Same as above - this could probably go into a `AVCapturePhotoSettings` wrapper class instead of here, no?

ios/CameraView+TakePhoto.swift
```diff
@@ -85,28 +120,39 @@ extension CameraView {
...
+        self.captureDelegates[photoSettings!.uniqueID] = delegate
+        // Remove the delegate reference when it finishes its processing.
+        delegate.didFinish = {
+          self.captureDelegates[photoSettings!.uniqueID] = nil
+        }
+        // Tell the output to capture the photo.
+        photoOutput.capturePhoto(with: photoSettings!, delegate: delegate)
+      } else {
+        photoOutput.capturePhoto(with: photoSettings!, delegate: PhotoCaptureDelegate(promise: promise))
+      }
```
mrousavy, line 153: This as well - we should probably extend (or subclass) `PhotoCaptureDelegate` to also be able to take RAW photos instead of adding this nested branch here. Or we make it a protocol and have two implementations, one for RAW photos one for "normal" photos.

ios/CameraView.swift
```diff
@@ -97,6 +98,7 @@ public final class CameraView: UIView {
   internal var lastFrameProcessorCall = DispatchTime.now()
   // CameraView+TakePhoto
   internal var photoCaptureDelegates: [PhotoCaptureDelegate] = []
+  internal var captureDelegates = [Int64: RAWCaptureDelegate]()
```
mrousavy, line 101: Why is this a map and the one above a simple array?

ios/RAWCaptureDelegate.swift
```diff
@@ -0,0 +1,107 @@
...
+          options.shouldMoveFile = true
+          creationRequest.addResource(with: .photo,
+                                      fileURL: rawFileURL,
+                                      options: options)
+        } completionHandler: { _, _ in
+          // Process the Photos library error.
+          print("Error while saving the RAW (DNG) file.")
+        }
+      }
+    }
```
mrousavy, line 105: Wait - what? Why are we writing to the Photo library? The other save photo code in VisionCamera simply writes to a temp file where we don't need permission, I'm sure this will also work for RAW photos, no?

src/CameraProps.ts
```diff
@@ -156,6 +156,11 @@ export interface CameraProps extends ViewProps {
    * @default false
    */
   enableHighQualityPhotos?: boolean;
+  /**
+   * Enable the Camera to take photos with RAW format.
+   * @platform iOS 10+
+   */
+  enableRawCapture?: boolean;
```
mrousavy, line 163: I think instead of a boolean it makes more sense to have a union type here so the user knows it's an "either this or that".

Something like

`captureType: 'raw' | 'photo'`

Maybe a better name for 'photo'.

mrousavy (changes requested):
Hi! Thanks so much for your PR, this is really cool. I have some nitpicks about code stuff, let me know what you think about this

mrousavy:
Closing due to inactivity.

Issue #887, opened 2022-03-10
🐛 [iOS] Extremely Slow Camera initialization with Reanimated 2.4.1


Cavallando:
### What were you trying to do?

We recently bumped our `react-native-reanimated` version to 2.4.1 and surprisingly it caused very very slow initialization (~15 seconds on average) times for the camera. I believe it was due to a CameraRuntime error being thrown but the thrown error didn't have much info.

This was hard to produce on device connected to a debugger but once we had an archived release the problem was extremely apparent.

Downgrading reanimated back down to 2.3.1 immediately resolved this issue.

### What happened instead?

I would expect the reanimated version to not affect the performance of this library.

### Device

iPhone 12, iPhone 11, iPhone 11 Pro Max, iPhone 13

### VisionCamera Version

2.11.2

### Additional information

- [ ] I am using Expo
- [X] I have read the [Troubleshooting Guide](https://mrousavy.github.io/react-native-vision-camera/docs/guides/troubleshooting)
- [X] I agree to follow this project's [Code of Conduct](https://github.com/mrousavy/react-native-vision-camera/blob/main/CODE_OF_CONDUCT.md)
- [X] I searched for [similar issues in this repository](https://github.com/mrousavy/react-native-vision-camera/issues) and found none.

mrousavy:
Hmm, I am running that REA version as well. Try clearing your buildcache and metro bundler cache and try again

Cavallando:
@mrousavy I really only saw this issue once when running in Debug mode connected to a physical device, it wasn't until we shipped a TestFlight that this issue was very apparent. Once we downgraded back to REA 2.3.1 and shipped another TestFlight the issue was gone.

hirbod:
@Cavallando I can always reproduce this when I have XCode open and try to open the camera while XCode is attached to the app (eg showing logs etc).

Running the camera without XCode Debug is fast. With Xcode it freezes for like 10-15 seconds until the camera gets active.
I have this issue with 2.3.1 as well though.

hirbod:
I have a stacktrace (without a crash but in the moment it does freeze) @mrousavy 

```log
=================================================================
Main Thread Checker: UI API called on a background thread: -[UIApplication statusBarOrientation]
PID: 393, TID: 33374, Thread name: (none), Queue name: mrousavy/VisionCamera.main, QoS: 33
Backtrace:
[...]

PR #1091, opened 2022-06-13
Accessing outputOrientation needs to be done in main thread

fobin:
## What

This PR fixes UI being accessed from non main thread on iOS. 

## Changes

When self.outputOrientation is accessed it jumps to main queue for it. 

## Tested on

iPhone 8, iOS 14.4.1

## Related issues

* Relates to #887

### DIFF
--- a/ios/CameraView+Orientation.swift
+++ b/ios/CameraView+Orientation.swift
@@ -52,7 +52,10 @@ extension CameraView {
               connection.automaticallyAdjustsVideoMirroring = false
               connection.isVideoMirrored = isMirrored
             }
-            connection.setInterfaceOrientation(self.outputOrientation)
+            // Need to jump back to main thread to update the orientation
+            DispatchQueue.main.async {
+              connection.setInterfaceOrientation(self.outputOrientation)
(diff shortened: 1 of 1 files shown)


mrousavy:
Hey! Thanks for the PR, but now we have three thread hops in that function. I wonder if we need to run the funcs on the `cameraQueue` here, did you try removing that and see if it actually is blocking? https://github.com/mrousavy/react-native-vision-camera/blob/0f4490d9a49b503aba8760decb632812e5354b28/ios/CameraView%2BOrientation.swift#L47-L48

fobin:
Is just took comment in the code as fact. 😀  Not so familiar with AVCaptureSession. But I tried without the `cameraQueue` just throwing it all in main thread and it works so it might be better solution. Want me to update the PR with that?

mrousavy:
Ah I think I found a better approach: https://github.com/mrousavy/react-native-vision-camera/pull/1094 that way it doesn't block nor does it access it on the wrong thread

fobin:
Yup! 👍

PR #1109, opened 2022-06-27
fix: Use `facebook::jsi` instead of `jsi` to fix build error in  react native 0.69

hrupesh:
## What

This PR fixes the Build error in React Native 0.69 for iOS

## Changes

Replaced `jsi` with `facebook::jsi`

## Tested on

* iPhone 11, iOS 15.3

### DIFF
--- a/ios/React Utils/JSConsoleHelper.mm
+++ b/ios/React Utils/JSConsoleHelper.mm
@@ -37,18 +37,18 @@
     return nil;
   }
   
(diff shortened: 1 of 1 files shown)


unfernandito:
Hello, nice PR's, i think all we need an update with this code

mm-webx (approved):

mrousavy:
Thank you for the PR!

caferyukseloglu:
Hi this should control the rn version for rn 0.68.2, all users who are getting the build error should downgrade for now.

Issue #1112, opened 2022-06-29
🐛Recorded video duration is returning 0.


usaidather:
### What were you trying to do?

After a latest update, when you record video, duration is 0 in iOS it was working fine on previous updates.

### What happened instead?

return {"duration": 0, "path": "file:///private/var/mobile/Containers/Data/Application/1BBA1F87-4BF5-4185-B9B6-AFB496EDE7EF/tmp/ReactNative/D03DD64C-8DED-49B0-8F34-B7BC69C93CC4.mov"}

### Device

iPhone 12

### VisionCamera Version

2.13.5

### Additional information

- [ ] I am using Expo
- [X] I have read the [Troubleshooting Guide](https://mrousavy.github.io/react-native-vision-camera/docs/guides/troubleshooting)
- [X] I agree to follow this project's [Code of Conduct](https://github.com/mrousavy/react-native-vision-camera/blob/main/CODE_OF_CONDUCT.md)
- [X] I searched for [similar issues in this repository](https://github.com/mrousavy/react-native-vision-camera/issues) and found none.

mrousavy:
On which version did it work? 2.13.4?

thomas-coldwell:
My bad! I removed a line that set the `initialTimestamp` so it always zero'd out when I was making some changes. Should be fixed in the above PR now :)

sergeymild:
Hi, on ios in callback `onRecordingFinished`  video.duration is always zero library version `^2.13.5`

PR #1114, opened 2022-06-30
fix: Set initial timestamp on first frame

thomas-coldwell:
## What

Sets the `initialTimestamp` when the first frame gets received in the recording session. This fixes the issue referenced in https://github.com/mrousavy/react-native-vision-camera/issues/1112 with the duration on iOS being reported incorrectly as zero.

## Tested on

iPhone 13 Pro

## Related issues

https://github.com/mrousavy/react-native-vision-camera/issues/1112


mrousavy:
Ace, thanks for the PR Thomas! ❤️

PR #1201, opened 2022-08-18
feat: implement jpeg compression quality

### LINKED ISSUE #1059: ❓ Resizing images
### Question

Whats the recommended approach of scaling the images down to optimize storage space etc before uploading to a backend service? We are taking pictures to confirm the placement of packages, but we dont need to have 2MB sized pictures for that.

I've tried playing with presets and formats, but I'm not really getting the results I would expect.

### What I tried

<Camera
    preset={"low"} // 'medium', 'photo' etc

### VisionCamera Version

2.13.0

### Additional information

- [ ] I am using Expo

jbrodriguez:
## What

This PR adds support for setting jpeg image compression quality

## Changes

- This PR adds a jpegQuality prop to Camera
- This PR adds a jpegQuality protp to takePhoto

## Tested on

- Huawei Y5
- iPhone 11

## Related issues
* Fixes #1059

### DIFF
--- a/android/src/main/java/com/mrousavy/camera/CameraView.kt
+++ b/android/src/main/java/com/mrousavy/camera/CameraView.kt
@@ -72,7 +72,7 @@ class CameraView(context: Context, private val frameProcessorThread: ExecutorSer
     const val TAG = "CameraView"
     const val TAG_PERF = "CameraView.performance"
 
-    private val propsThatRequireSessionReconfiguration = arrayListOf("cameraId", "format", "fps", "hdr", "lowLightBoost", "photo", "video", "enableFrameProcessor")
+    private val propsThatRequireSessionReconfiguration = arrayListOf("cameraId", "format", "fps", "hdr", "lowLightBoost", "photo", "video", "enableFrameProcessor", "jpegQuality")
     private val arrayListOfZoom = arrayListOf("zoom")
   }
 
@@ -110,6 +110,7 @@ class CameraView(context: Context, private val frameProcessorThread: ExecutorSer
       lastFrameProcessorPerformanceEvaluation = System.currentTimeMillis()
       frameProcessorPerformanceDataCollector.clear()
     }
+  var jpegQuality = 0
 
   // private properties
   private var isMounted = false
@@ -451,6 +452,9 @@ class CameraView(context: Context, private val frameProcessorThread: ExecutorSer
         }
       }
 
+      if (jpegQuality >= 1 && jpegQuality <= 100) {
+        imageCaptureBuilder.setJpegQuality(jpegQuality)
+      }
 
       // Unbind use cases before rebinding
       videoCapture = null
--- a/ios/CameraView+TakePhoto.swift
+++ b/ios/CameraView+TakePhoto.swift
@@ -47,6 +47,15 @@ extension CameraView {
         }
         if photoOutput.availablePhotoCodecTypes.contains(photoCodec) {
           format = [AVVideoCodecKey: photoCodec]
+          if (photoCodec == AVVideoCodecType.jpeg) {
+            if let jpegQuality = options["jpegQuality"] as? Float {
+              guard jpegQuality >= 0.0 && jpegQuality <= 1.0 else {
+                promise.reject(error: .parameter(.invalid(unionName: "JpegQuality", receivedValue: jpegQuality)))
+                return
+              }
+              format![AVVideoQualityKey] = jpegQuality
+            }
+          }
         } else {
           promise.reject(error: .capture(.invalidPhotoCodec))
           return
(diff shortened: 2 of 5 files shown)


ios/CameraView+TakePhoto.swift
```diff
@@ -47,6 +47,15 @@ extension CameraView {
         }
         if photoOutput.availablePhotoCodecTypes.contains(photoCodec) {
           format = [AVVideoCodecKey: photoCodec]
+          if (photoCodec == AVVideoCodecType.jpeg) {
+            if let jpegQuality = options["jpegQuality"] as? Float {
+              guard jpegQuality >= 0.0 && jpegQuality <= 1.0 else {
+                promise.reject(error: .parameter(.invalid(unionName: "JpegQuality", receivedValue: jpegQuality)))
+                return
+              }
+              format![AVVideoQualityKey] = jpegQuality
```
mrousavy, line 56: Does that actually work? Will it have lower quality when you pass something in in the resulting photo? Seems weird that they don't have a nicer API for this, no?

android/src/main/java/com/mrousavy/camera/CameraView.kt
```diff
@@ -110,6 +110,7 @@ class CameraView(context: Context, private val frameProcessorThread: ExecutorSer
       lastFrameProcessorPerformanceEvaluation = System.currentTimeMillis()
       frameProcessorPerformanceDataCollector.clear()
     }
+  var jpegQuality = 0
```
mrousavy, line 113: Indentation is wrong here, no?

src/CameraProps.ts
```diff
@@ -161,6 +161,15 @@ export interface CameraProps extends ViewProps {
...
 
+  /**
+   * Specifies the compression quality of the photo. The value ranges from `1` (lowest quality) to `100` (highest quality).
+   * 
+   * @platform Android
+   * @default 0 - not set, in which case it uses [`95`](https://developer.android.com/reference/kotlin/androidx/camera/core/ImageCapture.Builder#setjpegquality)
+   * 
+   */
+  jpegQuality?: number;
+
```
mrousavy, line 172: I kinda don't like how we have two separate props for this. One is for the `<Camera>`, one is for the options in `takePhoto(...)`. Either we do both to `<Camera>` and just use that for iOS as well, or we find a way to get the one on Android into `takePhoto(...)` too (better in my opinion)

Do you think it's possible we can move `jpegQuality` to `takePhoto(...)` for both platforms?

mrousavy (changes requested):
Hi! Thanks so much for your PR, this is really good stuff. One thing that bothers me are that we have two separate props, one for `<Camera>` one for `takePhoto(...)` options. I'd like to have both in `takePhoto(...)` if possible since that's the most logical. Is there a way we can configure Android "on the fly" like that?

ios/CameraView+TakePhoto.swift
jbrodriguez: yes, it works !

there's another api [jpegData(compressionQuality:)](https://developer.apple.com/documentation/uikit/uiimage/1624115-jpegdata), but you need to have an image already, something that we don't have at this step of the flow

android/src/main/java/com/mrousavy/camera/CameraView.kt
jbrodriguez: will fix it

src/CameraProps.ts
jbrodriguez: yes, having separate props is very painful to me as well.

the big issue is that on android, the jpeg quality setting is part of the [`ImageCapture.Builder` class](https://developer.android.com/reference/androidx/camera/core/ImageCapture.Builder#setJpegQuality(int)) and this is set up while initializing the camera, way before we do a takePicture.

on the other hand, on ios, the only (seemingly) way to set jpeg quality is when taking the picture, at least from the research i've done so far.

also, not sure how the previous react-native-camera implemented it, but i think you're using a different camera class on android right ? so it wouldn't help to explore that

anderpaz:
Do you have a forecast for this pr?

ios/CameraView+TakePhoto.swift
mrousavy: Oh wow I did not know that - I think then we have a way to do it on iOS as well!

mrousavy: Oh well actually yea I don't really like that this has to be configured with the session instead of just `takePhoto` options... We need to find another solution that works like on Android

jbrodriguez: yes, it's quirky :(

not sure there's another solution, although i haven't looked at it in quite a while

mrousavy: Yea I think if we convert to CGImage first then we can compress with quality, but imo it's an unnecessary round-trip.

I hope there's gonna be an API for that soon, otherwise that's just how it's gonna be 🤷‍♂️

mrousavy:

Exciting news: I believe this is now obsolete, as VisionCamera V5 now supports this! 🥳 
I'm closing this PR, but please let me know if this is still not working as intented in V5 and we can open a new issue for discussion, or possibly a PR already with a fix.

PR #1255, opened 2022-09-24
fix: CameraDevices type can't be imported

efefurkankarakaya:
## What

This PR provides CameraDevices able to be used as type.

ex: ` const devices: CameraDevices = useCameraDevices();`

## Changes

Just added an `export` keyword to the related line.

## Tested on

Pixel 5 XL, Android 10


mrousavy:
Thanks!

efefurkankarakaya:
You're welcome.

> Thanks!

Issue #1044, opened 2022-05-13
❓ shutter sound not playing on Android


albert-pjj:
### Question

The camera shutter sound plays in ios, but the shutter sound does not play in Android.

### What I tried

```
<Camera
        ref={camera}
        device={device}
        format={formats[0]}
        isActive={isActive}
       photo={true}
       orientation="portrait"
/>

const result = await camera.current?.takePhoto({
        enableAutoStabilization: true,
        skipMetadata: Platform.OS === 'android',
        qualityPrioritization: 'speed',
        flash: 'off',
});
```

```
buildToolsVersion = "31.0.0"
minSdkVersion = 23
compileSdkVersion = 31
targetSdkVersion = 31
ndkVersion = "21.4.7075529"
```

react-native: 0.68.0

### VisionCamera Version

2.13.3

### Additional information

- [ ] I am using Expo
- [X] I have read the [Troubleshooting Guide](https://mrousavy.github.io/react-native-vision-camera/docs/guides/troubleshooting)
- [X] I agree to follow this project's [Code of Conduct](https://github.com/mrousavy/react-native-vision-camera/blob/main/CODE_OF_CONDUCT.md)
- [X] I searched for similar questions [in the issues page](https://github.com/mrousavy/react-native-vision-camera/issues?q=is%3Aopen+is%3Aissue+label%3A%22❓+question%22) as well as [in the discussions page](https://github.com/mrousavy/react-native-vision-camera/discussions) and found none.

mrousavy:
I don't think there is any default shutter sound effect on Android?

kesteer:
@mrousavy 
You are correct in that sense. I think the issue comes from people who are moving over from RN Camera to RNVC, as IIRC RN Camera had a shutter sound implemented for Android. 

A disclaimer in the documentation might be a good idea. Since people testing on both types of devices will probably get thrown off by it?

Or, I was considering making a PR that would add this feature since there is technically a default sound for this in the Android OS if you'd be open to that.
https://developer.android.com/reference/android/media/MediaActionSound#SHUTTER_CLICK

mrousavy:
from their docs:

> The sound used by Camera.takePicture to indicate still image capture.

Why doesn't it play then? I think this would be a cool option for VisionCamera. It's configurable both on iOS and Android

kesteer:
> Why doesn't it play then? I think this would be a cool option for VisionCamera. It's configurable both on iOS and Android

[...]

PR #1072, opened 2022-05-31
feat: Allow control of shutter sound on iOS

wassgha:
## What

Adds functionality to control whether the shutter sound plays or not on iOS

## Changes

* Adds capture setting `disableShutterSound` to prevent shutter sound from playing on iOS
* Uses it on the example project

## Tested on

* iPhone 13 Pro, iOS 15.5

## Related issues
* Fixes #311

PR #1277, opened 2022-10-12
feat: Add shutter sound control

### LINKED ISSUE #311: ✨ How do I enable/disable the shutter sound
### Question

How do I enable or disable the shutter sound when taking a photo?

For some reason, when testing on my physical iPhone device, the sound plays when flash is off but doesn't play when flash is on.

Is this customizable? Thanks!

### What I tried

_No response_

### VisionCamera Version

2.4.1

### Additional information

- [ ] I am using Expo

imjihun:
## What
 * This PR Adds shutter sound control

## Changes
 * This PR changes Add disableShutterSound option to takePhoto function (default: false)

## Tested on
 * iPhone 12 Pro, iOS 15.4.1
 * Android 10 API 29 (Galaxy Note9)

## Related issues
 * Fixes #311, #1044

### DIFF
--- a/android/src/main/java/com/mrousavy/camera/CameraView+TakePhoto.kt
+++ b/android/src/main/java/com/mrousavy/camera/CameraView+TakePhoto.kt
@@ -2,6 +2,7 @@ package com.mrousavy.camera
 
 import android.annotation.SuppressLint
 import android.hardware.camera2.*
+import android.media.MediaActionSound
 import android.util.Log
 import androidx.camera.camera2.interop.Camera2CameraInfo
 import androidx.camera.core.ImageCapture
@@ -61,6 +62,7 @@ suspend fun CameraView.takePhoto(options: ReadableMap): WritableMap = coroutineS
     // TODO enableAutoDistortionCorrection
   }
   val skipMetadata = if (options.hasKey("skipMetadata")) options.getBoolean("skipMetadata") else false
+  val disableShutterSound = if (options.hasKey("disableShutterSound")) options.getBoolean("disableShutterSound") else false
 
   val camera2Info = Camera2CameraInfo.from(camera!!.cameraInfo)
   val lensFacing = camera2Info.getCameraCharacteristic(CameraCharacteristics.LENS_FACING)
@@ -70,6 +72,9 @@ suspend fun CameraView.takePhoto(options: ReadableMap): WritableMap = coroutineS
       Log.d(CameraView.TAG, "Taking picture...")
       val startCapture = System.nanoTime()
       val pic = imageCapture!!.takePicture(takePhotoExecutor)
+      if(!disableShutterSound) {
+        takePhotoSound!!.play(MediaActionSound.SHUTTER_CLICK)
+      }
       val endCapture = System.nanoTime()
       Log.i(CameraView.TAG_PERF, "Finished image capture in ${(endCapture - startCapture) / 1_000_000}ms")
       pic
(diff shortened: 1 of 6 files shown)


mrousavy:
Hey, thanks for your PR!

As mentioned in the previous PR (https://github.com/mrousavy/react-native-vision-camera/pull/1072), I'd rather have a prop called `enableShutterSound` that is `true` by default, rather than using reverse logic here. Makes more sense to me.

Also, I think this can be implemented on Android as well, so would be great if we could add that and don't have that limited for iOS only?

ios/CameraView+TakePhoto.swift
```diff
@@ -103,7 +103,13 @@ extension CameraView {
         photoSettings.isAutoContentAwareDistortionCorrectionEnabled = enableAutoDistortionCorrection
       }
 
-      photoOutput.capturePhoto(with: photoSettings, delegate: PhotoCaptureDelegate(promise: promise))
+      // disableShutterSound
+      var disableShutterSound = false
+      if let _disableShutterSound = options["disableShutterSound"] as? Bool{
+        disableShutterSound = _disableShutterSound
+      }
```
mrousavy: This probably looks a bit cleaner as a one-liner? not sure though
```suggestion
      let disableShutterSound = (options["disableShutterSound"] as? Bool) ?? false
```

imjihun:
@mrousavy Thanks for the suggestion!

android/src/main/java/com/mrousavy/camera/CameraView+TakePhoto.kt
```diff
@@ -70,6 +72,9 @@ suspend fun CameraView.takePhoto(options: ReadableMap): WritableMap = coroutineS
       Log.d(CameraView.TAG, "Taking picture...")
       val startCapture = System.nanoTime()
       val pic = imageCapture!!.takePicture(takePhotoExecutor)
+      if(!disableShutterSound) {
+        takePhotoSound!!.play(MediaActionSound.SHUTTER_CLICK)
+      }
```
mrousavy, line 77:
```suggestion
      if (!disableShutterSound) {
        takePhotoSound.play(MediaActionSound.SHUTTER_CLICK)
      }
```

mrousavy:
Hey @imjihun! Thanks for your edit, on Android it looks good. 

But as I said in my first review, I'd want the thing called `enableShutterSound` instead of `disableShutterSound`. Then it is `false` by default, and the user can enable it if he wants. This makes more sense to me than the reverse logic. What do you think?

jeffBoom:
> Hey @imjihun! Thanks for your edit, on Android it looks good.
> 
> But as I said in my first review, I'd want the thing called `enableShutterSound` instead of `disableShutterSound`. Then it is `false` by default, and the user can enable it if he wants. This makes more sense to me than the reverse logic. What do you think?

Personally, I think having the shutter sound on should be the default behavior. I have a need to disable the shutter sound but I would imagine it's not that common?

mrousavy:
I think we should have it disabled by default, that's the default behaviour on both native iOS and native Android :)

jeffBoom:
> I think we should have it disabled by default, that's the default behaviour on both native iOS and native Android :)

Ah my bad I didn't know that. Makes sense.

copterisimo:
@jeffBoom Closely following your PR, looks well done!

ueda-tetsuya:
When will this PR be merged into main?

jkaufman:
Thank you for the PR, I'm glad to see this. Are there plans to merge, @mrousavy? Anything needed from the community to get it over the finish line?

mrousavy:
Hey thanks for your PR but this is fixed in V3!

PR #1364, opened 2022-11-30
feat: Render Frame offscreen

mrousavy:
## What

The Camera can push frames at any refresh rate, whether that's 30, 60, 90, 120 or even 240 FPS - but the screen refreshes at a different refresh rate (synced to DisplayLink), e.g. 30, 60 or 120 FPS.

Since those are not sync, we often have situations where we render the Camera Frame (incl. FP call) to the UI Drawable more than once, which is undefined behaviour.

The solution to that would be to render the Camera Frame to an off-screen Metal Texture, and only render that in-memory Metal Texture to the screen on every UI refresh.


mrousavy:
LGTM!

PR #1428, opened 2023-01-24
fix: filter out non-integer cameraIds

KodyFintak:
## What

When using this library with our android device connected to an external display with a camera (I think) the app crashes with a NullPointerException.

```sh
Failed to parse camera Id /dev/video5 to integer
java.lang.NullPointerException
at com.mrousavy.camera.utils.CameraCharacteristicsUtilsKt.getDeviceTypes(CameraCharacteristicsUtils.kt:30)
at com.mrousavy.camera.CameraViewModule$getAvailableCameraDevices$1.invokeSuspend(CameraViewModule.kt:193)
at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)
```

Looking at the CameraManager, the cameraId that it is getting is `/dev/video5`.

Our very bad fix just filters out cameraIds that our non-integers (as the code is parsing that)

## Tested on
Android API 29 tablet + external display


mrousavy:
Hi! Thanks for your PR, but this is not the right way to fix this. I just looked at the stacktrace, looks like `CameraCharacteristicsUtils.kt` is expecting `CameraCharacteristics.LENS_INFO_AVAILABLE_FOCAL_LENGTHS`, which is apparently not available for your external device.

I wonder what type of device this then is 🤔 

Which of these device types do you think is most fitting for your external camera @KodyFintak? https://github.com/mrousavy/react-native-vision-camera/blob/ae2c28a2c27a2ef6afa276b1b2d2034d2651027c/src/CameraDevice.ts#L11

KodyFintak:
Thanks @mrousavy for taking a look! 100% agree on this not being the right way, figured this change would be the easiest way to show the issue. 

This device is a custom made external display for an Android Device that appears to have a camera but we don't really use it. I'm not positive what type it would be, but I assume it is a  `wide-angle-camera`.

KodyFintak:
Any update on a way for us to fix this?

To unblock our product, I had to fork this library with my change. I don't have access to change the Android device to give it a property that plays nice with `CameraCharacteristics.LENS_INFO_AVAILABLE_FOCAL_LENGTHS`.

mrousavy:
Hey!

I think we could merge this for now, but the effect of this PR is that it essentially removes some Camera devices from the devices list.
The proper way to fix this is that we make the devices handling more generic, allowing you to also use this device that's currently failing for you. I'll make sure to add that feature in V3, when @thomas-coldwell and I rewrite the android part from CameraX to Camera2.

PR #1452, opened 2023-02-02
fix: Unresolved reference resolveView

sheepma:


zhouhao27:
Is this merged? Still have the problem.

PR #1482, opened 2023-02-17
Example: Censor Face in Realtime using Skia

mrousavy:
This is just an Example. We shouldn't merge that PR.

This is an Example that censors a face in realtime by detecting faces with Vision Kit, then censoring them using Skia from the JS Frame Processor.

> The only native code is the face detection here, which I want to remove as well by using PyTorch inside a Frame Processor.

The current issue is that the detect Faces plugin is slow. I wanna use BlazeFace instead (which is also used by Snapchat afaik), since that can run at >60FPS. (cc @thomas-coldwell)

Try it:

```sh
git clone https://github.com/mrousavy/react-native-vision-camera
cd react-native-vision-camera
git checkout example/face-blur
yarn
cd example
yarn
cd ios && pod install && cd ..
```
...then run it on your device through Xcode.

https://user-images.githubusercontent.com/15199031/219749680-9c20bf14-9748-466f-b962-52bb9619bbb9.mov

Code:

```tsx
const faceBoundingBox = useMemo(() => Worklets.createSharedValue<FaceBoundingBox | undefined>(undefined), []);
const paint = Skia.Paint();
const color = Skia.Color('red');
paint.setColor(color);

const frameProcessor = useFrameProcessor(
  (frame) => {
    'worklet';

    runAsync(frame, () => {
      'worklet';
      const faces = detectFaces(frame);
      faceBoundingBox.value = faces[0];
    });

    const face = faceBoundingBox.value;
    if (face != null) {
      const width = face.width * frame.width;
      const height = face.height * frame.height;
      const x = face.x * frame.width;
      const y = frame.height - face.y * frame.height - height;
      const rect = Skia.XYWHRect(x, y, width, height);
      frame.drawRect(rect, paint);
    }
  },
  [paint, faceBoundingBox],
);

return <Camera device={selfieCam} frameProcessor={frameProcessor} />
```


mrousavy:
Closing as @riteshshukla04 or @DaveyEke will implement this as a separate Expo app in Margelo's GitHub :)

PR #1485, opened 2023-02-20
example: PyTorch Live integration

mrousavy:
## What

Uses PyTorch Core to convert a Frame to a Image. This would allow us to run any ML Model inside VisionCamera straight from JS - no native code (Frame Processor Plugins) - proof of concept PR: https://github.com/facebookresearch/playtorch/pull/199


mrousavy:
Closing as @riteshshukla04 or @DaveyEke will implement this as a separate Expo app in Margelo's GitHub :)

PR #1487, opened 2023-02-21
feat: Add `toByteArray()`, `orientation`, `isMirrored` and `timestamp` to `Frame`

mrousavy:
## What

Adds three new props and one new function to the `Frame` object:

* `toByteArray()`: Gets the Frame data as a byte array. The type is `Uint8Array` (TypedArray/`ArrayBuffer`). Keep in mind that Frame buffers are usually allocated on the GPU, so this comes with a performance cost of a GPU -> CPU copy operation. I've optimized it a bit to run pretty fast :)
* `orientation`: The orientation of the Frame. e.g. `"portrait"`
* `isMirrored`: Whether the Frame is mirrored (eg in selfie cams)
* `timestamp`: The presentation timestamp of the Frame

### DIFF
--- a/android/CMakeLists.txt
+++ b/android/CMakeLists.txt
@@ -26,6 +26,7 @@ endif()
 add_library(
         ${PACKAGE_NAME}
         SHARED
+        ../cpp/JSITypedArray.cpp
         src/main/cpp/VisionCamera.cpp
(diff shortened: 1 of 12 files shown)


Overtorment:
can you get frame data inside js with this?

mrousavy:
Yes :)

PR #1492, opened 2023-02-23
feat: Rewrite Android (CameraX -> Camera2)

mrousavy:
## What

This PR rewrites the entire Android part. I replaced CameraX with Camera2 🥲🥲🥲🥲🥲🥲🥲🥲

## Why

There's three Camera libraries:

* Camera1 (not enough features, deprecated)
* Camera2 (lots of features but complicated)
* CameraX (not enough features again)

CameraX is based on Camera2 and aims to make make everything simpler by abstracting device specific code away into a declarative API.
I chose to write VisionCamera in CameraX because Google claims that new apps should be built with CameraX:

image

However CameraX has proven multiple times that it is simply not ready for advanced use-cases.

Reasons:

* Device/Format selection works way different than on iOS and is not predictable with the current API
* FPS selection doesn't work (Camera is capped at 30FPS)
* No Slow-mo support
* No multi-camera device support (fisheye)
* Some devices simply don't work


mrousavy:
lol so much code down the drain. rewritten from scratch in another pr

PR #1466, opened 2023-02-09
feat: ✨ V3 ✨

mrousavy:
See https://github.com/mrousavy/react-native-vision-camera/issues/1376

## Breaking Changes

* Frame Processors are now **synchronous**. Previously they ran on a separate Thread. If you want to run something on a separate Thread now, use `runAsync` inside a Frame Processor
* Frame Processor Plugins are no longer in the global object with the `__` prefix, but rather stored directly in the `FrameProcessorPlugins` object exported by react-native-vision-camera. (e.g. replace `__scanQRCodes(frame)` with `FrameProcessorPlugins.scanQRCodes(frame)`)
* `frameProcessorFps` no longer exists. Use `runAtTargetFps` inside a Frame Processor to throttle some calls.
* `onFrameProcessorPerformanceSuggestionAvailable` no longer exists. Use the FPS display (`enableFpsGraph={true}`) to see how your Frame Processor performs over time. This is more in-line with how React Native works (Dev Tools / Perf Monitor)
* VisionCamera V3 will not work on RN 0.70 or below. You need to use RN 0.71. This is because the build script got way simpler and smaller, making it faster to build and way less error prone. Backwards compatibility is just too complex here.
* Reanimated is no longer used as a Worklet Runtime. Instead, VisionCamera now uses [react-native-worklets-core](https://github.com/margelo/react-native-worklets-core).

## Progress

You can test the latest V3 release by creating a new RN project with RN 0.71 and installing VisionCamera + RNWorklets:

```sh
yarn add react-native-vision-camera@3.0.0-rc.5
yarn add react-native-worklets-core
yarn add @shopify/react-native-skia
```

Things to test:

* TensorFlow Lite plugin to load any `.tflite` model!! ✨ (see [this PR for more info](https://github.com/mrousavy/react-native-vision-camera/pull/1633), will be a separate library soon)
* Drawing onto a Frame using Skia!! 🎉 
* Using `frame.toArrayBuffer()` to get the Frame's byte content in JS
* New Android build script. This should drastically speed up the build time! 💨
* New Worklet library. This replaces Reanimated Worklets. Should be faster and more stable :)
* New synchronous Frame Processors. Should be faster :)
* `runAtTargetFps` and `runAsync` in Frame Processors
* Using HostObjects or HostFunctions (like models from PyTorch) inside a Frame Processor. This will probably require a few native bindings on PyTorch's end to make the integration work (cc @raedle)

[...]

PR #1497, opened 2023-02-25
(WIP) feat: replace `runAtTargetFps` to `useRunAtTargetFps`

bglgwyng:
## What
`runAtTargetFps` doesn't work properly when there are multiple usages of it in a frame processor.

```typescript
  runAtTargetFps(15, () => {
    'worklet';
    scanFaces(frame);
  });
  runAtTargetFps(5, () => {
    'worklet';
    scanCodes(frame);
  });
});
```
`scanFrame` is never called while `scanCodes` works well with throttling.
Also, nested `runAtTargetFps` doesn't work either
```typescript
  runAtTargetFps(15, () => {
    'worklet';
    const faces = scanFaces(frame);
    runAtTargetFps(5, () => {
      'worklet';
      recognizeFaces(faces);
    });
  });
});
```

This is because the earlier call of `runAtTargetFps` updates `lastFrameProcessorCall` to `performance.now()` so that the following call of `runAtTargetFps` never executes `func` since `diffToLastCall` would be very small.

## Changes
Introduce new `useRunAtTargetFps` hook that creates a separated shared value for the last execution time.

Now we can use like
```typescript
const throttleScanFaces = useRunAtTargetFps();
const throttleScanCodes = useRunAtTargetFps();
  throttleScanFaces(15, () => {
    'worklet';
    scanFaces(frame);
  });
  throttleScanCodes(5, () => {
    'worklet';
    scanCodes(frame);
  });
});
```

However, I think we can give a better name to it than `useRunAtTargetFps`.

## Tested on
iPhone 13

## Related issues
https://github.com/mrousavy/react-native-vision-camera/pull/1466

### DIFF
--- a/src/FrameProcessorPlugins.ts
+++ b/src/FrameProcessorPlugins.ts
@@ -1,6 +1,7 @@
 import type { Frame, FrameInternal } from './Frame';
 import { Camera } from './Camera';
 import { Worklets } from 'react-native-worklets/src';
+import { useMemo } from 'react';
 
 // Install VisionCamera Frame Processor JSI Bindings and Plugins
 Camera.installFrameProcessorBindings();
@@ -15,8 +16,6 @@ type TFrameProcessorPlugins = Record<string, FrameProcessor>;
  */
 // @ts-expect-error The global JSI Proxy object is not typed.
 export const FrameProcessorPlugins = global.FrameProcessorPlugins as TFrameProcessorPlugins;
-
-const lastFrameProcessorCall = Worklets.createSharedValue(performance.now());
 
 /**
  * Runs the given function at the given target FPS rate.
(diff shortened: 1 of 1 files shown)


mrousavy:
Ah, good catch! I think that API is a bit too complicated, let me see if I can come up with a better solution.

mrousavy:
Hey! I just found a different solution for this that doesn't require any changes from the user: https://github.com/mrousavy/react-native-vision-camera/pull/1466/commits/af4e366312a2335535f80109e6bf0b3a977444f0

This will only work without Reanimated though, as in Reanimated nested worklets won't get transpiled (no `__workletHash` in nested worklets) - see https://github.com/software-mansion/react-native-reanimated/pull/4053 (cc @tomekzaw @chrfalch @kmagiera jfyi)

PR #1509, opened 2023-03-03
fix: improve "Native Module 'CameraView' was null!" error message

Nivani:
## What

We ran into the error message "Camera: Native Module 'CameraView' was null! Did you run pod install?" while building an app with Expo.

This happened because we were not building a native app directly, but were running the app through the Expo Go app (using the `expo start` command). Running the app with `expo run:ios` or `expo run:android` fixed the problem.

This PR updates the error message to help people who run into the same issue.

## Changes

This PR updates the "Native Module 'CameraView' was null!" error message.

### DIFF
--- a/src/Camera.tsx
+++ b/src/Camera.tsx
@@ -38,7 +38,7 @@ type RefType = React.Component<NativeCameraViewProps> & Readonly<NativeMethods>;
 // NativeModules automatically resolves 'CameraView' to 'CameraViewModule'
 // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
 const CameraModule = NativeModules.CameraView;
-if (CameraModule == null) console.error("Camera: Native Module 'CameraView' was null! Did you run pod install?");
+if (CameraModule == null) console.error("Camera: Native Module 'CameraView' was null! Did you run pod install? Or are you running expo start instead of expo run:ios or expo run:android?");
 
 //#region Camera Component
 /**


mrousavy:
Hey! Thanks for your PR! ❤️ 

 I just added more details to the error in https://github.com/mrousavy/react-native-vision-camera/pull/1515, this will be merged to V3. Wdyt?

Nivani:
Hey @mrousavy, I looked at the pull request and if I understand correctly, the message would be

> react-native-vision-camera is not supported in Expo Go! Use EAS (`expo prebuild`) or eject to a bare workflow instead.

Using EAS would fix the problem, but is not productive for development. That would mean a lot of waiting for long builds.

I had to look up what ejecting to a [bare workflow](https://docs.expo.dev/introduction/managed-vs-bare/#bare-workflow
) means.
 
It's not necessary to do that. It's sufficient to run a development build on a device using `expo run:ios` or `expo run:android` (documented under [Running a project with a development build](https://docs.expo.dev/workflow/run-on-device/#running-a-project-with-a-development-build)) instead of using `expo start`.

So I think this would be more helpful:

> react-native-vision-camera is not supported in Expo Go! Use EAS (`expo prebuild`) or run a development build (`expo run:[ios|android])` instead.

Should I comment in #1515 ?

mrousavy:
Good point, just updated it. I also asked Evan & Brent for their opinions.

Nivani:
Will there be a better error message in an upcoming v2 release as well?

mrousavy:
Hey! No, this will be for V3 only. Thanks for your PR, I took the idea into V3. Sorry for not merging this, but it was easier to target V3 now - will improve DX in the future more :)

PR #1515, opened 2023-03-07
feat: Better Native Module Error Detection

mrousavy:
## What

Throw a more precise error when the Camera Module cannot be found (e.g. telling the user to rebuild, install pods, not use expo, etc)

### DIFF
--- a/src/NativeCameraModule.ts
+++ b/src/NativeCameraModule.ts
@@ -19,7 +19,7 @@ if (CameraModule == null) {
       // We're running Expo Go
       throw new CameraRuntimeError(
         'system/camera-module-not-found',
-        'react-native-vision-camera is not supported in Expo Go! Use EAS (`expo prebuild`) or eject to a bare workflow instead.',
+        `react-native-vision-camera is not supported in Expo Go! Use EAS/expo prebuild instead (\`expo run:${Platform.OS}\`). For more info, see https://docs.expo.dev/workflow/prebuild/.`,
       );
     } else {
       // We're running Expo bare / standalone
--- a/src/NativeCameraModule.ts
+++ b/src/NativeCameraModule.ts
@@ -1,27 +1,25 @@
 import { NativeModules, Platform } from 'react-native';
+import { CameraRuntimeError } from './CameraError';
 
 // NativeModules automatically resolves 'CameraView' to 'CameraViewModule'
 // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
 export const CameraModule = NativeModules.CameraView;
 if (CameraModule == null) {
-  let message =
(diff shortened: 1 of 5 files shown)


src/NativeCameraModule.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  // check if Expo
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  const ExpoConstants = NativeModules.NativeUnimoduleProxy?.modulesConstants?.ExponentConstants;
+  if (ExpoConstants != null) {
+    if (ExpoConstants.appOwnership === 'expo') {
+      // We're running Expo Go
+      throw new CameraRuntimeError(
+        'system/camera-module-not-found',
+        `react-native-vision-camera is not supported in Expo Go! Use EAS/expo prebuild instead (\`expo run:${Platform.OS}\`). For more info, see https://docs.expo.dev/workflow/prebuild/.`,
```
mrousavy, line 31: Hey @EvanBacon & @brentvatne, what do you think about this wording? Is that correct, or should I do something like "Use the new Expo managed workflow instead"?

src/NativeCameraModule.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  let message = 'Failed to initialize VisionCamera: The native Camera Module (`NativeModules.CameraView`) could not be found.';
+  message += '\n* Make sure react-native-vision-camera is correctly autolinked (run `npx react-native config` to verify)';
+  if (Platform.OS === 'ios' || Platform.OS === 'macos') message += '\n* Make sure you ran `pod install` in the ios/ directory.';
+
+  if (Platform.OS === 'android') message += '\n* Make sure gradle is synced.';
+
+  // check if Expo
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  const ExpoConstants = NativeModules.NativeUnimoduleProxy?.modulesConstants?.ExponentConstants;
```
tsapeta, line 25: In the future `NativeModules.NativeUnimoduleProxy` will become deprecated and removed. Now all Expo modules are installed through JSI and accessible from `global.expo.modules` object so something like this might be better 👇 
```suggestion
  const ExpoConstants = global.expo?.modules?.ExponentConstants;
```

mrousavy: Thanks Tomasz, changed this on v3! 👍 

Out of curiosity, since which Expo version is this supported? Do I need to think about backwards compatibility with the `NativeUnimodulesProxy` if VisionCamera only runs on RN 71 and above anyways, or no?

tsapeta: If that's >=0.71 then I guess you don't need to worry about that. `global.expo.modules` was added in SDK 47 in which the recommended RN version is 0.70 😉

PR #1541, opened 2023-03-25
fix: iOS focus

levipro:
## What

iOS focus method did not do anything because the device stayed in `continuousAutoFocus` mode

## Changes

- use `autoFocus` and `autoExpose` instead when focus is called
- enable `isSubjectAreaChangeMonitoringEnabled`
- when subject area is changed then switch back to `continuousAutoFocus` and `continuousAutoExposure`

## Tested on

* iPhone 13 Pro iOS 16.3
* iPhone 6s Plus iOS 13.1.2

## Related issues

https://github.com/mrousavy/react-native-vision-camera/issues/1376#issuecomment-1483633319

### DIFF
--- a/ios/CameraView+Focus.swift
+++ b/ios/CameraView+Focus.swift
@@ -74,7 +74,11 @@ extension CameraView {
       }
 
       // in {0..1} system
-      let normalizedPoint = captureDevicePointConverted(fromLayerPoint: point)
+      var normalizedPoint = captureDevicePointConverted(fromLayerPoint: point)
+      if let previewView = previewView as? PreviewView {
+        // previewView is of type PreviewView can use the built in captureDevicePointConverted
+        normalizedPoint = previewView.videoPreviewLayer.captureDevicePointConverted(fromLayerPoint: point)
+      }
 
       do {
         try device.lockForConfiguration()
--- a/ios/CameraView+Focus.swift
+++ b/ios/CameraView+Focus.swift
@@ -74,10 +74,12 @@ extension CameraView {
       }
 
       // in {0..1} system
-      var normalizedPoint = captureDevicePointConverted(fromLayerPoint: point)
+      var normalizedPoint: CGPoint
       if let previewView = previewView as? PreviewView {
         // previewView is of type PreviewView can use the built in captureDevicePointConverted
         normalizedPoint = previewView.videoPreviewLayer.captureDevicePointConverted(fromLayerPoint: point)
+      } else {
+        normalizedPoint = captureDevicePointConverted(fromLayerPoint: point)
       }
 
       do {
--- a/ios/CameraView+Focus.swift
+++ b/ios/CameraView+Focus.swift
@@ -80,18 +80,52 @@ extension CameraView {
         try device.lockForConfiguration()
 
         device.focusPointOfInterest = normalizedPoint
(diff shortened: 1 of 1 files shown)


ios/CameraView+Focus.swift
```diff
@@ -74,24 +74,64 @@ extension CameraView {
...
 
       // in {0..1} system
-      let normalizedPoint = captureDevicePointConverted(fromLayerPoint: point)
+      var normalizedPoint: CGPoint
+      if let previewView = previewView as? PreviewView {
+        // previewView is of type PreviewView can use the built in captureDevicePointConverted
+        normalizedPoint = previewView.videoPreviewLayer.captureDevicePointConverted(fromLayerPoint: point)
+      } else {
+        normalizedPoint = captureDevicePointConverted(fromLayerPoint: point)
+      }
```
mrousavy, line 83: Let's remove this separate check and just make sure the `captureDevicePointConverted` method works as expected. That way we are future proof - maybe there could be a different Preview View type in the future.

mrousavy, line 83: Let's remove this separate check and just make sure the `captureDevicePointConverted` method works as expected. That way we are future proof - maybe there could be a different Preview View type in the future.

ios/CameraView+Focus.swift
```diff
@@ -74,24 +74,64 @@ extension CameraView {
...
+
         device.unlockForConfiguration()
         return nil
       } catch {
         throw CameraError.device(DeviceError.configureError)
       }
     }
   }
+
+  @objc func subjectAreaDidChange(notification: NSNotification) {
```
mrousavy, line 113: Question: Does this get called immediately after it successfully focused on the given point?

mrousavy (changes requested):
Hi - thanks for your PR! This looks good & thanks for the comments. I would however change this one thing (move `captureDevicePointConverted` into my own method, or at least let's try that).
lmk if you have any questions!
Thanks!

ios/CameraView+Focus.swift
levipro: I am of a different opinion here. I think it is better to use the built in Apple method to do this point conversion when available. Also at the moment `captureDevicePointConverted` does not work. If you log the output of `previewView.videoPreviewLayer.captureDevicePointConverted(fromLayerPoint: point)` and `captureDevicePointConverted(fromLayerPoint: point)` they produce completely different results. 

Can a TODO be added as a task to resolve the discrepancies between the custom method and the built in one? I honestly do not have time at the moment to dig into this any more but would like to get this fix merged so that it is fixed for when people are not using the skia preview layer.

ios/CameraView+Focus.swift
levipro: No it gets called when the subject area changes. This means that if you tap to focus on a "subject" the focus gets locked on that object, but of there is movement in that area then the callback happens and the focus and exposure are set back to `continuous` mode and will adjust automatically (removing the lock that the focus event enabled with `autoFocus` and `autoExpose`

mbenassarou:
Having the same issue with tap to focus, I tested Levipro's version and it works perfectly for me on iOS (iPhone 13 pro)

mrousavy:
So after focus is complete, it returns back to continuous auto focus? makes sense

mrousavy:

PR #1566, opened 2023-04-22
Update build.gradle

aniketjha9179:
upgrading the Kotlin Gradle plugin version in the react-native-vision-camera project to version 1.6.20 or higher, which is supported by the Android Gradle plugin version you are using


donasyst:
I think you should change the version in gradle.propreties file and then comment those two lines:
classpath "org.jetbrains.kotlin:kotlin-android-extensions:$kotlin_version"
apply plugin: 'kotlin-android-extensions'

PR #1570, opened 2023-04-24
fix: use project.hermesEnabled if present otherwise fallback to project.ext.react.enableHermes

asherLZR:
## What

Starting from RN71, the `project.hermesEnabled` flag in `properties.gradle` supersedes the previous `project.ext.react.enableHermes`. However, the option to use the latter is still available and we use it in our project as it is more customisable.

See the RN gradle plugin: https://github.com/facebook/react-native/blob/0.72-stable/packages/react-native-gradle-plugin/src/main/kotlin/com/facebook/react/utils/ProjectUtils.kt#L33

## Changes

This PR implements a similar logic to the gradle plugin to fall back to  `project.ext.react.enableHermes` if `project.hermesEnabled` is not defined.

## Tested on

Project using `project.ext.react.enableHermes`.

## Related issues

None


mrousavy:
Hey! Thanks for your PR, but we no longer have this code in V3. Sorry for the long radio silence here, I have been overwhelmed with PRs previously.

PR #1583, opened 2023-05-04
Delay frame processor initialization until view creation on Android

### LINKED ISSUE #946: 🐛 java.lang.NullPointerException in FrameProcessorRuntimeManager.<init>

We have 40k MAU and we see 230 occurrences of this error for the last 30 days.

### Reproduceable Code

_No response_

### What happened instead?

The issue seems to happen on app start.

### Relevant log output

```shell
java.lang.NullPointerException: 
  at java.util.Objects.requireNonNull (Objects.java:245)
  at com.mrousavy.camera.frameprocessor.FrameProcessorRuntimeManager.<init> (FrameProcessorRuntimeManager.kt:33)
  at com.mrousavy.camera.CameraViewManager$1.run (CameraViewManager.kt:20)
  at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1167)
  at java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:641)
  at java.lang.Thread.run (Thread.java:920)
```

### Device

Android only: most seen on Google Pixel 4a, Google Pixel 6, Samsung Galaxy S20 FE 5G, Samsung Galaxy Z Flip3 5G, all have Android 12

### VisionCamera Version

2.10.0

### Additional information

- [ ] I am using Expo

jthure:
## What
This PR fixes java.lang.NullPointerException in FrameProcessorRuntimeManager.<init> on Android phones

Although I never was able to reproduce the issue myself, looking at the crashlogs from other users, this should fix it

## Changes
Moved the creation `FrameProcessorRuntimeManager` to happen in `onCreateViewInstance` instead of inside `CameraViewModule` initialisation. This should make sure that the react context is fully initialised, which apparently is not the case right now

Inspired by @pke in https://github.com/mrousavy/react-native-vision-camera/issues/946#issuecomment-1506825351

## Tested on

Not properly tested yet. I'm having trouble building and running the example app from this repo. I have tested it in another app which uses this library. Any help to test and verify the change is greatly appreciated

## Related issues
Fixes #946

### DIFF
--- a/android/src/main/java/com/mrousavy/camera/CameraViewManager.kt
+++ b/android/src/main/java/com/mrousavy/camera/CameraViewManager.kt
@@ -12,6 +12,7 @@ class CameraViewManager(reactContext: ReactApplicationContext) : ViewGroupManage
 
   public override fun createViewInstance(context: ThemedReactContext): CameraView {
     val cameraViewModule = context.getNativeModule(CameraViewModule::class.java)!!
+    cameraViewModule.installFrameProcessorBindings()
     return CameraView(context, cameraViewModule.frameProcessorThread)
   }
 
--- a/android/src/main/java/com/mrousavy/camera/CameraViewModule.kt
+++ b/android/src/main/java/com/mrousavy/camera/CameraViewModule.kt
@@ -58,16 +58,6 @@ class CameraViewModule(reactContext: ReactApplicationContext) : ReactContextBase
     frameProcessorManager = null
   }
 
-  override fun initialize() {
-    super.initialize()
-
-    if (frameProcessorManager == null) {
-      frameProcessorThread.execute {
-        frameProcessorManager = FrameProcessorRuntimeManager(reactApplicationContext, frameProcessorThread)
-      }
-    }
-  }
-
   override fun onCatalystInstanceDestroy() {
     super.onCatalystInstanceDestroy()
     cleanup()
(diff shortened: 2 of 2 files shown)


pke:
Thanks for creating this PR!

mrousavy:
Hey! Thanks for creating this PR, but I don't think that this is the right approach to this problem. This just swallows the error and silently fails when you try to use FPs on the JS side later on.

Also, this elevates the internal setup from internal (`CameraViewModule.kt`) to external (`CameraViewManager.kt`; aka the CameraViewModule's caller) code.

mrousavy:
In V3 I have a different solution here which is called from JS - so this is when the JS bundle is 100% ready. That's probably what we want.

Sorry for closing this PR please don't take it personal ❤️

pke:
Hmm... this patch was written against current main, which is not yet V3. Why not allow this backport until V3 is production ready?

mrousavy:
Because of this:

> This just swallows the error and silently fails when you try to use FPs on the JS side later on.

the FP runtime will not be installed and will just not work. You can disable FPs at build time if that's what you want

jthure:
> Hey! Thanks for creating this PR, but I don't think that this is the right approach to this problem. This just swallows the error and silently fails when you try to use FPs on the JS side later on.
> 
> Also, this elevates the internal setup from internal (`CameraViewModule.kt`) to external (`CameraViewManager.kt`; aka the CameraViewModule's caller) code.

Okay sure, I get what you mean. Do you have another approach in mind on how to fix it? Before v3 is ready and adopted?

Edit: Actually, I'm not sure I fully agree. It doesn't just swallow the error. The error should not happen since the fp init happens at a later stage, right?

mrousavy:
> Do you have another approach in mind on how to fix it?

If you do not need Frame Processors, just disable them at build time. 

If you do, then I am not sure - I think the solution in V3 is much more solid since it does init from JS side - but this is a big change

> Actually, I'm not sure I fully agree. It doesn't just swallow the error. The error should not happen since the fp init happens at a later stage, right?

No, it is wrapped in a `try/catch` now and in the `catch` there's no handling - aka it silently fails.

Issue #1600, opened 2023-05-19
❓ Orientation switching on device rotation causes momentary preview freeze


ajp8164:
### Question

How can we prevent the preview video connection from freezing momentarily when switching orientation by changing the vision camera  orientation property. 

### What I tried

Only  looking at iOS at the moment..

I implemented the hook `useDeviceRotationSensor` as described in [624](https://github.com/mrousavy/react-native-vision-camera/issues/624). I also created an alternative local patch for iOS and providing  the same capability in swift. In both cases the result is the same. Rotate the device, the orientation for the video connection (AVCaptureVideoDataOutput) is changed. However, changing the orientation seems to  disconnect the current data writer, re-allocate buffers  and reconnect the data writer.

There must be a strategy to accomplish this since expo-camera does so as well as the iOS Camera app.

### VisionCamera Version

2.15.4

### Additional information

- [ ] I am using Expo
- [X] I have read the [Troubleshooting Guide](https://react-native-vision-camera.com/docs/guides/troubleshooting)
- [X] I agree to follow this project's [Code of Conduct](https://github.com/mrousavy/react-native-vision-camera/blob/main/CODE_OF_CONDUCT.md)
- [X] I searched for similar questions [in the issues page](https://github.com/mrousavy/react-native-vision-camera/issues?q=is%3Aopen+is%3Aissue+label%3A%22❓+question%22) as well as [in the discussions page](https://github.com/mrousavy/react-native-vision-camera/discussions) and found none.

ajp8164:
In case others would like the camera to behave as in iOS Messages app (which is my goal)...

The momentary pause is caused by an orientation listener that is forcing a change to the camera output connection (buffer reallocation etc) when the device is rotated. To prevent this I have patched the iOS implementation to avoid the orientation listener under certain conditions and have added an output connection settings change to be done immediately prior to taking a photo and starting video recording. This works reasonably well - it's good enough for me to take to production. Android suffers a similar issue and so I'm working on that solution as well.  I'll issue a PR when I'm ready. In the mean time you can checkout my fork if you're interested - and make improvement suggestions ;-)

ajp8164:
Please see [PR 1619](https://github.com/mrousavy/react-native-vision-camera/pull/1619) for a native implementation.

Lots of open PRs. Probably need to live with patch..   🙁

myselfuser1:
[...]

Issue #1572, opened 2023-04-25
🐛 Orientation always shows as 6


lucksp:
### What were you trying to do?

I am trying to determine orientation so that if a photo taken in landscape (horizontal device), it can be viewed as a horizontal photo while holding the device vertically in a portrait orientation.

Since the `orientation` prop is not functioning as expected (there are open bugs & PR #1544 ), i need to make a work around.

I am trying to read the EXIF to have logic around what to do to manipulate the image being saved.

As you can see, the console output shows BOTH images as Orientation of `6`

![IMG_7083 (2)](https://user-images.githubusercontent.com/15162169/234346407-132be61c-75e6-4e09-902e-48e82e5bc456.jpeg) ![IMG_7084 (2)](https://user-images.githubusercontent.com/15162169/234346412-7a3b226e-73ce-4bf5-9a91-799732cecec5.jpeg)

### Reproduceable Code

```tsx
const capturePhoto = async () => {
    if (cameraRef?.current && !snapping) {
      setSnapping(true);
      const photo = await cameraRef.current.takePhoto({
        flash: flashMode,
      });
      console.error('photo.metadata.Orientation: ', photo.metadata.Orientation); // outputs 6

      handleSaveAsset([photo]);
    }
  };

<PinchGestureHandler onGestureEvent={onPinchGesture} enabled={isActive}>
  <ReanimatedCamera
    // Do not add enableZoomGesture since using the Animated lib
    // enableZoomGesture
    ref={cameraRef}
    photo
    device={device}
    isActive={isActive}
    format={formats?.[0]}
    animatedProps={animatedProps}
    orientation={'portrait'} // this doesn't seem to impact output
/>
</PinchGestureHandler>
<TouchableOpacity onPress={capturePhoto} disabled={speedbump}>
  <CameraButtonIcon />
</TouchableOpacity>
```
```


### What happened instead?



I would have expected a different number than 6.

I cannot use the EXIF data to read the orientation of the image.

### Relevant log output

_No response_

### Device

iPhone 13 Pro

### VisionCamera Version

^2.15.4

### Additional information

- [X] I am using Expo
- [X] I have read the [Troubleshooting Guide](https://react-native-vision-camera.com/docs/guides/troubleshooting)
- [X] I agree to follow this project's [Code of Conduct](https://github.com/mrousavy/react-native-vision-camera/blob/main/CODE_OF_CONDUCT.md)
[...]

PR #1544, opened 2023-03-28
fix: updated orientation calculations

sroske:
## What

This PR should fix the issues in video orientation (preview and recording) being "fixed" to portrait orientation regardless of device orientation in iOS.

Adapted from https://stackoverflow.com/a/36575423

## Changes

* Changed the CameraView orientation extension so that it takes the UIDevice's reported orientation if the userOrientation is not provided
* On rotation change it will update the preview layer orientation and update the preview layer bounds accordingly

## Tested on

* iPhone 14 Pro
* iPad Mini

## Related issues

* Fixes #1539 
* Fixes #1455 
* Fixes #1445

Issue #1455, opened 2023-02-06
🐛 [iPadOS] If app is set to landscape only, vision camera gets preview orientation wrong


zzz08900:
### What were you trying to do?

We have an iPadOS-only app that only supports landscape orientation. We tried to use react-native-vision-camera@2.15.4 with this app but found the preview is displayed with wrong orientation - it was rotated 90 degrees.
So when the iPad moves left and right, the image in react-native-vision-camera preview moves up and down.

### What happened instead?

With an app that only supports landscape orientation, when the iPad moves left and right, the image in react-native-vision-camera preview moves up and down.

### Device

iPad (9th Generation)

### VisionCamera Version

2.15.4

### Additional information

- [ ] I am using Expo
- [X] I have read the [Troubleshooting Guide](https://mrousavy.github.io/react-native-vision-camera/docs/guides/troubleshooting)
- [X] I agree to follow this project's [Code of Conduct](https://github.com/mrousavy/react-native-vision-camera/blob/main/CODE_OF_CONDUCT.md)
- [X] I searched for [similar issues in this repository](https://github.com/mrousavy/react-native-vision-camera/issues) and found none.

zzz08900:
This bug is not present in version 2.14.x, I highly suspect this piece of code is what causing this issue.
![Screenshot from 2023-02-06 14-04-02](https://user-images.githubusercontent.com/51951532/216895589-7dba7190-e801-4176-906a-e1085a0f1d4d.png)

Before 2.15.3 there were codes that detects current orientation on camera launch but after this commit, it seems that react-native-vision-camera assumes the device is portrait on camera launch, which caused our issue.

I'll fiddle around a little more and return with what happened later.

JackHull:
Can confirm this issue with newest version.

mrousavy:
Can you record a video (with another phone) and post that here?

zzz08900:
> Can you record a video (with another phone) and post that here?

Sorry I got my hands really busy these days - almost overwhelmed by customer demands. Will do this weekend.

zzz08900:
https://user-images.githubusercontent.com/51951532/220416909-70097676-6578-433c-bd5c-2b216a0de46d.mp4


@mrousavy here's the video. You can see from the texture of the background that the content of `<Camera />` view is moving perpendicular to the movement direction of the iPad.

[...]

PR #1619, opened 2023-06-12
feat: Implement camera orientation type "device", allows user device orientation to configure camera output

ajp8164:
## What

Implements camera orientation type `device`.

The existing implementation has four main orientations as values to the `orientation` attribute for the `<Camera>`. Using any of these four values informs the camera to provide output in only that specific orientation regardless of how the user has oriented the device to take a photo or video. This change introduces an additional value called `device`. When `orientation={'device'}` the camera responds to the users device orientation during preview and when a photo or video is created the camera will reconfigure itself to use  the device orientation at the moment the photo or video is created.

Other implementations have created a similar solution from the consuming app by detecting rotation using an RN library and then re-rendering the camera with a changed `orientation` attribute. Functionally this works but the user experience is very bad on iOS - the preview freezes and resumes at the moment the camera re-renders. The preview freeze is caused when the camera output connection is removed and reallocated using the changed configuration. The native implementation in this PR provides excellent user experience and a simpler implementation for consuming apps - developer only needs to set `orientation={'device'}`.

## Changes

Adds orientation type `device` to the RN component.
Adds re-writing the camera output configuration to be equal to the device rotation at the moment a photo or video is captured. Android and iOS implementations provide the same behavior.

All changes are backward compatible.

## Tested on

- iPhone 12 mini
- Samsung Galaxy Note 9

## Related issues

This PR should address and enable use cases identified in these issues:

- #1600
- #1572 
- #1544
- #1455 
- #1445
- #1375
- #1327

### DIFF
--- a/android/src/main/java/com/mrousavy/camera/CameraView+TakePhoto.kt
+++ b/android/src/main/java/com/mrousavy/camera/CameraView+TakePhoto.kt
@@ -3,6 +3,7 @@ package com.mrousavy.camera
 import android.annotation.SuppressLint
 import android.hardware.camera2.*
 import android.util.Log
+import android.view.Surface
 import androidx.camera.camera2.interop.Camera2CameraInfo
 import androidx.camera.core.ImageCapture
 import androidx.camera.core.ImageProxy
@@ -95,10 +96,28 @@ suspend fun CameraView.takePhoto(options: ReadableMap): WritableMap = coroutineS
     exif = if (skipMetadata) null else ExifInterface(file)
   }
 
+  var width = photo.width
+  var height = photo.height
+
+  if (orientation == "device") {
+    when (deviceRotation) {
+      Surface.ROTATION_0, Surface.ROTATION_180 -> {
+        val temp = Math.min(width, height)
+        height = Math.max(width, height)
+        width = temp;
+      }
+      Surface.ROTATION_90, Surface.ROTATION_270 -> {
+        val temp = Math.max(width, height)
+        height = Math.min(width, height)
+        width = temp;
+      }
+    }
+  }
+
   val map = Arguments.createMap()
   map.putString("path", file.absolutePath)
-  map.putInt("width", photo.width)
-  map.putInt("height", photo.height)
+  map.putInt("width", width)
+  map.putInt("height", height)
   map.putBoolean("isRawPhoto", photo.isRaw)
 
   val metadata = exif?.buildMetadataMap()
--- a/android/src/main/java/com/mrousavy/camera/CameraView.kt
+++ b/android/src/main/java/com/mrousavy/camera/CameraView.kt
@@ -24,13 +24,11 @@ import androidx.lifecycle.*
 import com.facebook.jni.HybridData
 import com.facebook.proguard.annotations.DoNotStrip
 import com.facebook.react.bridge.*
-import com.facebook.react.uimanager.events.RCTEventEmitter
 import com.mrousavy.camera.frameprocessor.FrameProcessorPerformanceDataCollector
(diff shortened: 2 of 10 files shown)


mrousavy:
Hey! Thanks for your PR. This is a really interesting topic, but I'm not sure if this is the right approach.

I will be taking a closer look at orientation soon, but there's just many things to consider.
What we want is the best developer experience for the majority of the cases, and easy APIs for some advanced use cases. Maybe orientation should follow the App UI orientation by default, and only lock in place if the user overrides orientation.

ajp8164:
> Hey! Thanks for your PR. This is a really interesting topic, but I'm not sure if this is the right approach.
> 
> I will be taking a closer look at orientation soon, but there's just many things to consider. What we want is the best developer experience for the majority of the cases, and easy APIs for some advanced use cases. Maybe orientation should follow the App UI orientation by default, and only lock in place if the user overrides orientation.

Thanks for having a look!

I'm happy to make changes if there  are additional requirements. I agree that this PR does not recognize the users ability to disable rotation. However, disabling device rotation is generally targeting whole app rotation. This implementation is designed to be narrowly focused on  capturing the media according to device rotation (doesn't target  the app or views outside the camera). OTOH, I suppose app requirements for implementation could state to capture rotated media according to device rotation but allow the user to disable this as desired (OS device rotation setting and/or in app settings). Such an implementation (given this PR) then requires the developer to change the `orientation` setting on the vision camera component which then forces a re-render of the camera (not sure if that's performance optimized - likely not is my bet). Perhaps adding a function on the component that allows toggle on/off when component is built with `orientation` set to `device`. See below as an example. This proposes a new function called `lockOrientation()`.

`lockOrientation(orientation: OrientationType | undefined): void`

When `value` is not undefined for `lockOrientation(value)` then behavior precisely follows that when Camera is built with `orientation={'portrait'}` (the default, and so _**on..) Setting `value` is `device` is a no-op. Calling this function when Camera is built with something other than `device` is a no-op.

Anyway - lmk what you think.. you may be able to simplify ;-)


```
type OrientationType = 'portrait' | 'portraitUpsideDown' | 'landscapeLeft' | 'landscapeRight' | 'device';

function App() {
  const camera = useRef<Camera>(null);
  const [rotationLock, setRotationLock] = useState<OrientationType | undefined>();
  ...

  // Toggle between landscape left and any orientation.
  // This function could also consider the devices orientation lock setting at the OS.
  const toggleOrientationLock = () => {
    const newOrientation = 'landscapeLeft';
    if (!orientationLock) {
      newOrientation = undefined
    }
      ref.lockOrientation(newOrientation); // Proposal for locking orientation when Camera built using 'device' orientation.
      setOrientationLock(newOrientation);
  };

  return (
    <Camera
      ref={camera}
      orientation={'device'}
      ...
    />
  )
}
```

ajp8164:
@mrousavy Would love to have your input/direction on this. Need to have this target V3 as well. Thanks for the review!

mrousavy:
Hey yea I think I am a bit too overwhelmed right now with changes in VisionCamera, I'd love to leave this PR open and use it for reference for implementing this into v3.
I really appreciate your time and effort here, this is a complex topic and I don't want to break anything. Hope you understand!

mrousavy:
Hey! Orientation is really complex. Sorry for the long radio silence here, I have been overwhelmed with PRs and work in general. This issue is now used to track orientation: https://github.com/mrousavy/react-native-vision-camera/issues/1891

Your Pr is extremely relevant, but I think `device` should be the users responsibility by passing `orientation` as a state (listen of orientation changes as a hook).

mrousavy:
Hey - orientation is now implemented in latest Visioncamera! :)

PR #1691, opened 2023-08-14
feat: Video Recording with Camera2

mrousavy:
Implements video recording with Camera2.

Goals:

- [x] Set up Video Recorder helper class
- [x] Write video frames to an video file (mp4?)
- [ ] Write audio buffers to the video file (mp4?)
- [x] Control fps setting
- [x] Control resolution
- [ ] Allow flipping the Camera device while recording / no interruptions
- [x] Use efficient codecs if possible (hevc/h265)
- [ ] Handle interruptions accordingly
- [ ] Use no middleman if Frame Processor is not enabled - direct write to Surface


mrousavy:
It works!! 🎉🎉🎉

I found the `ImageWriter` API which allows me to pass-through the Images to the Recorder Surface. :)

There are a few hickups after a few seconds of recording, no idea why. I'll hopefully fix this when I switch to MediaCodec instead of MediaRecorder

PR #1717, opened 2023-08-24
feat: Flash

mrousavy:
## What

Flash is **waaaaaaaaaay** more complicated than I thought.

On iOS, this is a simple `flash: .on` parameter to the take photo options.

On Android, you need to:

1. Set the auto-exposure to `CONTROL_AE_MODE_ON_ALWAYS_FLASH`
2. Restart the repeating request (preview) with that auto-exposure mode so it is ready to adjust auto-exposure when the flash fires
3. Create a single **PRECAPTURE** request with `CONTROL_AE_PRECAPTURE_TRIGGER_START` 
4. Capture a single output with that **PRECAPTURE** request to trigger an auto-exposure event (effectively firing the flash)
5. Wait until the **PRECAPTURE** state went on, and then off again
6. Reset the auto-exposure to it's original state
7. Then take a photo

This is so much callback/waiting stuff. Insane. Good luck finding any documentation on that online 😂


mrousavy:
lol thankfully CameraX works now

PR #1739, opened 2023-08-31
chore: add frame processor plugins examples for Kotlin & Swift

mateusz1913:
## What

Right now I'm working on [support for VisionCamera v3 plugins in vision-camera-plugin-builder](https://github.com/mateusz1913/vision-camera-plugin-builder/pull/39) and I found that ~`v3`~ `main` branch does not have examples for Kotlin & Swift plugins

## Changes

This PR adds Kotlin & Swift example for Frame Processor plugin. It also slightly enhances Frame Processor docs section

## Tested on

- iPhone 12 Pro, iOS 16.6
- Xiaomi Mi Note 10 Lite, Android 12

### DIFF
--- a/docs/docs/guides/FRAME_PROCESSOR_CREATE_FINAL.mdx
+++ b/docs/docs/guides/FRAME_PROCESSOR_CREATE_FINAL.mdx
@@ -16,8 +16,9 @@ const plugin = VisionCameraProxy.getFrameProcessorPlugin('scanFaces')
 /**
  * Scans faces.
  */
-export function scanFaces(frame: Frame): object {
+export function scanFaces(frame: Frame) {
   'worklet'
+  if (plugin == null) throw new Error("Failed to load Frame Processor Plugin!")
   return plugin.call(frame)
 }
 ```
--- a/docs/docs/guides/FRAME_PROCESSOR_CREATE_PLUGIN_ANDROID.mdx
+++ b/docs/docs/guides/FRAME_PROCESSOR_CREATE_PLUGIN_ANDROID.mdx
@@ -29,12 +29,13 @@ For reference see the [CLI's docs](https://github.com/mateusz1913/vision-camera-
 
 2. Register the package in MainApplication.java
 
-```java {6}
+```java
         @Override
         protected List<ReactPackage> getPackages() {
           @SuppressWarnings("UnnecessaryLocalVariable")
           List<ReactPackage> packages = new PackageList(this).getPackages();
-          ...
+          // ...
+// highlight-next-line
           packages.add(new FaceDetectorFrameProcessorPluginPackage()); // <- add
           return packages;
         }
@@ -54,14 +55,14 @@ For reference see the [CLI's docs](https://github.com/mateusz1913/vision-camera-
 2. Create a Java source file, for the Face Detector Plugin this will be called `FaceDetectorFrameProcessorPlugin.java`.
 3. Add the following code:
 
-```java {8}
+```java
 import com.mrousavy.camera.frameprocessor.Frame;
 import com.mrousavy.camera.frameprocessor.FrameProcessorPlugin;
 
 public class FaceDetectorFrameProcessorPlugin extends FrameProcessorPlugin {
-
   @Override
   public Object callback(Frame frame, Map<String, Object> arguments) {
+// highlight-next-line
     // code goes here
     return null;
   }
@@ -75,26 +76,24 @@ The Frame Processor Plugin will be exposed to JS through the `VisionCameraProxy`
 4. **Implement your Frame Processing.** See the [Example Plugin (Java)](https://github.com/mrousavy/react-native-vision-camera/blob/main/package/example/android/app/src/main/java/com/mrousavy/camera/example/ExampleFrameProcessorPlugin.java) for reference.
 5. Create a new Java file which registers the Frame Processor Plugin in a React Package, for the Face Detector plugin this file will be called `FaceDetectorFrameProcessorPluginPackage.java`:
 
-```java {12}
+```java
 import com.facebook.react.ReactPackage;
 import com.facebook.react.bridge.NativeModule;
 import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.uimanager.ViewManager;
 import com.mrousavy.camera.frameprocessor.FrameProcessorPlugin;
 import com.mrousavy.camera.frameprocessor.FrameProcessorPluginRegistry;
-import javax.annotation.Nonnull;
 
 public class FaceDetectorFrameProcessorPluginPackage implements ReactPackage {
-  @NonNull
   @Override
-  public List<NativeModule> createNativeModules(@NonNull ReactApplicationContext reactContext) {
+  public List<NativeModule> createNativeModules(ReactApplicationContext reactContext) {
+// highlight-next-line
     FrameProcessorPluginRegistry.addFrameProcessorPlugin("detectFaces", options -> new FaceDetectorFrameProcessorPlugin());
     return Collections.emptyList();
   }
 
-  @Nonnull
   @Override
-  public List<ViewManager> createViewManagers(@Nonnull ReactApplicationContext reactContext) {
+  public List<ViewManager> createViewManagers(ReactApplicationContext reactContext) {
     return Collections.emptyList();
   }
 }
@@ -102,12 +101,13 @@ public class FaceDetectorFrameProcessorPluginPackage implements ReactPackage {
 
 6. Register the package in MainApplication.java
 
-```java {6}
+```java
         @Override
         protected List<ReactPackage> getPackages() {
           @SuppressWarnings("UnnecessaryLocalVariable")
           List<ReactPackage> packages = new PackageList(this).getPackages();
-          ...
+          // ...
+// highlight-next-line
           packages.add(new FaceDetectorFrameProcessorPluginPackage()); // <- add
           return packages;
         }
@@ -120,13 +120,14 @@ public class FaceDetectorFrameProcessorPluginPackage implements ReactPackage {
 2. Create a Kotlin source file, for the Face Detector Plugin this will be called `FaceDetectorFrameProcessorPlugin.kt`.
 3. Add the following code:
 
-```kotlin {7}
+```kotlin
 import com.mrousavy.camera.frameprocessor.Frame
 import com.mrousavy.camera.frameprocessor.FrameProcessorPlugin
 
 class FaceDetectorFrameProcessorPlugin: FrameProcessorPlugin() {
 
-  override fun callback(frame: Frame, arguments: Map<String, Object>): Any? {
+  override fun callback(frame: Frame, arguments: Map<String, Object>?): Any? {
+// highlight-next-line
     // code goes here
     return null
   }
@@ -137,10 +138,10 @@ class FaceDetectorFrameProcessorPlugin: FrameProcessorPlugin() {
 The Frame Processor Plugin will be exposed to JS through the `VisionCameraProxy` object. In this case, it would be `VisionCameraProxy.getFrameProcessorPlugin("detectFaces")`.
 :::
 
-4. **Implement your Frame Processing.** See the [Example Plugin (Java)](https://github.com/mrousavy/react-native-vision-camera/blob/main/package/example/android/app/src/main/java/com/mrousavy/camera/example/ExampleFrameProcessorPlugin.java) for reference.
+4. **Implement your Frame Processing.** See the [Example Plugin (Kotlin)](https://github.com/mrousavy/react-native-vision-camera/blob/main/package/example/android/app/src/main/java/com/mrousavy/camera/example/ExampleKotlinFrameProcessorPlugin.kt) for reference.
 5. Create a new Kotlin file which registers the Frame Processor Plugin in a React Package, for the Face Detector plugin this file will be called `FaceDetectorFrameProcessorPluginPackage.kt`:
 
-```kotlin {9}
+```kotlin
 import com.facebook.react.ReactPackage
 import com.facebook.react.bridge.NativeModule
 import com.facebook.react.bridge.ReactApplicationContext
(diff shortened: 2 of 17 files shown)


docs/docs/guides/FRAME_PROCESSOR_CREATE_PLUGIN_ANDROID.mdx
```diff
@@ -29,12 +29,13 @@ For reference see the [CLI's docs](https://github.com/mateusz1913/vision-camera-
 
 2. Register the package in MainApplication.java
 
-```java {6}
```
mateusz1913, line 32: This is discouraged way of highlighting content in docusaurus

docs/docs/guides/FRAME_PROCESSOR_CREATE_PLUGIN_IOS.mdx
```diff
@@ -87,52 +87,47 @@ The Frame Processor Plugin will be exposed to JS through the `VisionCameraProxy`
...
-```objc
-#import <VisionCamera/FrameProcessorPlugin.h>
-#import <VisionCamera/Frame.h>
-```
-
-4. In the Swift file, add the following code:
+3. In the Swift file, add the following code:
 
 ```swift
+import VisionCamera
```
mateusz1913, line 93: Here I found that there's no need to use bridging header. Instead we can use VisionCamera as a Swift module.

For plugin implemented inside an application codebase nothing more is needed.

For plugin implemented as a library, VisionCamera needs to be referenced as a podspec dependency (I've already [mentioned it as one of post-steps](https://github.com/mateusz1913/vision-camera-plugin-builder/pull/39/files#diff-6eefc925eb2820a960acc6aff9d4f54af5cc66e446ebb6a480daac62f89ab8baR291) in the PR to vision-camera-plugin-builder)

example/android/app/build.gradle
```diff
@@ -1,5 +1,6 @@
 apply plugin: "com.android.application"
 apply plugin: "com.facebook.react"
+apply plugin: "kotlin-android"
```
mateusz1913: Here, I just added kotlin support for example android app, to make it possible to include kotlin file with the example plugin

rocket13011:
Hello @mateusz1913 Does not work since rc9, ;)

Frame function has changed and so have the callback arguments.

If you manage to make it work tell me, I'm testing a lot of modifications but nothing works.

Best regards.

mateusz1913:
👋 @mrousavy let me know if I should improve sth here

docs/docs/guides/FRAME_PROCESSOR_CREATE_FINAL.mdx
```diff
@@ -16,9 +16,9 @@ const plugin = VisionCameraProxy.getFrameProcessorPlugin('scanFaces')
 /**
  * Scans faces.
  */
-export function scanFaces(frame: Frame): object {
+export function scanFaces(frame: Frame) {
   'worklet'
-  return plugin.call(frame)
+  return plugin?.call(frame)
```
mrousavy: I think it'd be better if we don't swallow errors here. This should be non-null
```suggestion
  if (plugin == null) throw new Error("Failed to load Frame Processor Plugin!")
  return plugin.call(frame)
```

docs/docs/guides/FRAME_PROCESSOR_CREATE_PLUGIN_ANDROID.mdx
mrousavy: ahh nice!

docs/docs/guides/FRAME_PROCESSOR_CREATE_PLUGIN_ANDROID.mdx
```diff
@@ -54,14 +55,17 @@ For reference see the [CLI's docs](https://github.com/mateusz1913/vision-camera-
...
+import androidx.annotation.Nullable;
 import com.mrousavy.camera.frameprocessor.Frame;
 import com.mrousavy.camera.frameprocessor.FrameProcessorPlugin;
 
 public class FaceDetectorFrameProcessorPlugin extends FrameProcessorPlugin {
-
+  @Nullable
   @Override
-  public Object callback(Frame frame, Map<String, Object> arguments) {
+  public Object callback(@NonNull Frame frame, @Nullable Map<String, Object> arguments) {
```
mrousavy: I think to keep this simple we can skip the `@NonNull` and `@Nullable` annotations here and in the return decl and in the imports - just less overhead for someone not as affine with native development.

docs/docs/guides/FRAME_PROCESSOR_CREATE_PLUGIN_ANDROID.mdx
```diff
@@ -75,39 +79,41 @@ The Frame Processor Plugin will be exposed to JS through the `VisionCameraProxy`
...
+// highlight-next-line
     FrameProcessorPluginRegistry.addFrameProcessorPlugin("detectFaces", options -> new FaceDetectorFrameProcessorPlugin());
     return Collections.emptyList();
   }
 
-  @Nonnull
+  @NonNull
   @Override
-  public List<ViewManager> createViewManagers(@Nonnull ReactApplicationContext reactContext) {
+  public List<ViewManager> createViewManagers(@NonNull ReactApplicationContext reactContext) {
```
mrousavy: Same with nonnull annotations here

docs/docs/guides/FRAME_PROCESSOR_CREATE_PLUGIN_IOS.mdx
```diff
@@ -60,7 +60,7 @@ For reference see the [CLI's docs](https://github.com/mateusz1913/vision-camera-
   CMSampleBufferRef buffer = frame.buffer;
   UIImageOrientation orientation = frame.orientation;
   // code goes here
-  return @[];
+  return @{};
```
mrousavy: Maybe this should also be `null`/`nil` so it's the same on both iOS and Android? Now this returns an empty object/Dictionary...

docs/docs/guides/FRAME_PROCESSOR_CREATE_PLUGIN_IOS.mdx
```diff
@@ -87,52 +87,47 @@ The Frame Processor Plugin will be exposed to JS through the `VisionCameraProxy`
...
+
 @objc(FaceDetectorFrameProcessorPlugin)
 public class FaceDetectorFrameProcessorPlugin: FrameProcessorPlugin {
-  public override func callback(_ frame: Frame!, withArguments arguments: [String:Any]) -> Any {
+  public override func callback(_ frame: Frame, withArguments arguments: [AnyHashable : Any]?) -> Any {
     let buffer = frame.buffer
     let orientation = frame.orientation
     // code goes here
-    return []
+    return [:]
```
mrousavy: Maybe this should also be `null`/`nil` so it's the same on both iOS and Android? Now this returns an empty object/Dictionary...

docs/docs/guides/FRAME_PROCESSOR_CREATE_PLUGIN_IOS.mdx
```diff
@@ -87,52 +87,47 @@ The Frame Processor Plugin will be exposed to JS through the `VisionCameraProxy`
...
-4. In the Swift file, add the following code:
+3. In the Swift file, add the following code:
 
 ```swift
+import VisionCamera
+
 @objc(FaceDetectorFrameProcessorPlugin)
 public class FaceDetectorFrameProcessorPlugin: FrameProcessorPlugin {
-  public override func callback(_ frame: Frame!, withArguments arguments: [String:Any]) -> Any {
+  public override func callback(_ frame: Frame, withArguments arguments: [AnyHashable : Any]?) -> Any {
```
mrousavy, line 97: Is it really `AnyHashable`? I thought it'd always be string keys, but let's keep it at AnyHashable - good point on the nullable part

package/example/android/app/src/main/java/com/mrousavy/camera/example/ExampleKotlinFrameProcessorPlugin.kt
```diff
@@ -0,0 +1,37 @@
...
+
+        for (key in params.keys) {
+            val value = params[key]
+            Log.d("ExampleKotlinPlugin", "  -> " + if (value == null) "(null)" else value.toString() + " (" + value.javaClass.name + ")")
+        }
+
+        val map: MutableMap<String, Any> = HashMap()
+        map["example_str"] = "KotlinTest"
+        map["example_bool"] = false
+        map["example_double"] = 6.7
```
mrousavy: Could you use `mapOf`?

package/example/android/app/src/main/java/com/mrousavy/camera/example/ExampleKotlinFrameProcessorPlugin.kt
```diff
@@ -0,0 +1,37 @@
...
+
+        val array: MutableList<Any> = ArrayList()
+        array.add("Good bye")
+        array.add(false)
+        array.add(21.37)
```
mrousavy: Could you use `arrayListOf(..)`?

package/example/ios/Podfile
```diff
@@ -46,5 +46,48 @@ target 'VisionCameraExample' do
...
+          current_other_swift_flag_setting = "#{current_other_swift_flag_setting} #{flag}"
+        elsif !has_worklets && current_other_swift_flag_setting.include?(flag)
+          current_other_swift_flag_setting.slice! flag
+        end
+
+        config.build_settings[key] = current_other_swift_flag_setting
+      end
+      project.save()
+    end
+end
```
mrousavy, line 93: Holy shit haha do we really need that? Is the flag not available in that context? I know that it is available in ObjC compiler options...

mrousavy: I would really prefer to avoid that to keep the `Podfile` simple and maintanable..

package/example/src/CameraPage.tsx
```diff
@@ -201,6 +202,7 @@ export function CameraPage({ navigation }: Props): React.ReactElement {
 
     console.log(`${frame.timestamp}: ${frame.width}x${frame.height} ${frame.pixelFormat} Frame (${frame.orientation})`);
     examplePlugin(frame);
+    console.log(exampleKotlinSwiftPlugin(frame));
```
mrousavy: Can you remove that again? I don't think we need to call both plugins tbh.

mrousavy (changes requested):
Hey - thanks for your PR, just had some time on the airport to review this - nice work!

Left a few small comments, other than that it looks really good. 👍

docs/docs/guides/FRAME_PROCESSOR_CREATE_PLUGIN_IOS.mdx
mateusz1913: The Swift module (`import VisionCamera`) translates ObjC generic `NSDictionary` into `[AnyHashable : Any]`

Screenshot 2023-09-06 at 16 11 30

package/example/ios/Podfile
mateusz1913: The **_example project_** didn't have any custom swift flags and disabling code based on swift flag was the only way to make the example app be built _without_ frame processors. I also didn't find any setup regarding ObjC flag for **_example project_**. And if I'm not mistaken, we cannot use flags defined for the library code inside application's code

I may see, if I can call some of those functions directly from the react-native cocoapods helpers (as I based above solution on those helpers)

package/example/ios/VisionCameraExample-Bridging-Header.h
```diff
@@ -1,8 +1,3 @@
 //
 //  Use this file to import your target's public headers that you would like to expose to Swift.
 //
-
-#if VISION_CAMERA_ENABLE_FRAME_PROCESSORS
```
mateusz1913, line 5: That flag was never set for the example project and because now example swift plugin uses `import VisionCamera`, I decided to remove the dead code

package/example/ios/VisionCameraExample/AppDelegate.mm
```diff
@@ -1,10 +1,6 @@
 #import "AppDelegate.h"
 
 #import <React/RCTBundleURLProvider.h>
-#import "VisionCameraExample-Swift.h"
-#if VISION_CAMERA_ENABLE_FRAME_PROCESSORS
```
mateusz1913, line 5: Same as in the bridging header

mateusz1913:
Thx for the review, I fixed (or answered) the comments, see you in Wrocław (possibly :D)

package/example/ios/Podfile
mateusz1913: Using RN Cocoapods utils will not work, because [these can only set](https://github.com/facebook/react-native/blob/850349b1d274f0cc2595eee2f6bb361a958bb2e2/packages/react-native/scripts/cocoapods/utils.rb#L322) `OTHER_CFLAGS` & `OTHER_CPLUSPLUSFLAGS`, while here we need `OTHER_SWIFT_FLAGS`

~How about sth like this~

```ruby
# it's just import for some internal RN CocoaPods utils
require Pod::Executable.execute_command('node', ['-p',
  'require.resolve(
    "react-native/scripts/cocoapods/utils.rb",
    {paths: [process.argv[1]]},
  )', __dir__]).strip

def enable_frame_processor_swift_flag_in_example_project(installer)
  forceDisableFrameProcessors = false
  if defined?($VCDisableFrameProcessors)
    forceDisableFrameProcessors = true
  end
  has_worklets = installer.pods_project.pod_group("react-native-worklets-core") != nil && !forceDisableFrameProcessors
  flag = "-DVISION_CAMERA_ENABLE_FRAME_PROCESSORS"
  if has_worklets
    ReactNativePodsUtils.add_compiler_flag_to_project(installer, flag)
  else
    ReactNativePodsUtils.remove_compiler_flag_from_project(installer, flag)
  end
end
```

~The downside is that it's not meant to be a public API and may be changed between RN releases.~

~Let me know, what do you think~

mateusz1913:
Closing this one in favor of two smaller PRs

The first one with docs is here - https://github.com/mrousavy/react-native-vision-camera/pull/1877
The second with examples will be prepared later

PR #1746, opened 2023-09-01
fix: savePhotoToFile would write empty files

vonovak:
## What

this fixes the `savePhotoToFile` function in v3 which on my device, would write empty files to the disk.
Closing the stream fixed that but I also believe I made it a little more efficient thanks to removing a buffer allocation in case of non-selfie images

also added one more entry to `android/.editorconfig` because android studio was formatting the code differently from the rest, not sure why but it's always annoying 😅 

## Tested on

Honor 20 Pro

### DIFF
--- a/package/android/.editorconfig
+++ b/package/android/.editorconfig
@@ -1,5 +1,6 @@
 [*.{kt,kts}]
 indent_size=2
+indent_style=space
 insert_final_newline=true
 max_line_length=off
 disabled_rules=no-wildcard-imports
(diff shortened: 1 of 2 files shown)


mrousavy:
Hey - oh how could I forget this! Thanks man this looks good 👍

mrousavy:
Sorry I just merged a PR that moves everything into `package/`. Can you quickly merge `v3` into your PR to fix that? Thanks

vonovak:
@mrousavy conflicts fixed 👍

PR #1749, opened 2023-09-02
fix(android): detect buffer state prior to writing it

vonovak:
## What

this is a follow-up on #1746 which I somehow didn't test properly (or so it seems), sorry about that.

## Changes

It seems that sometimes buffer needs to be flipped before writing it. Other times, it does not 🤷 . I also `close`d the stream for selfies, but I didn't test that.

The method now returns the number of bytes written, which could be used later on. 

## Tested on

Honor 20 Pro

### DIFF
--- a/package/android/src/main/java/com/mrousavy/camera/CameraView+TakePhoto.kt
+++ b/package/android/src/main/java/com/mrousavy/camera/CameraView+TakePhoto.kt
@@ -18,7 +18,6 @@ import com.mrousavy.camera.utils.*
 import kotlinx.coroutines.*
 import java.io.File
 import java.io.FileOutputStream
-import java.io.OutputStream
 
 private const val TAG = "CameraView.takePhoto"
 
@@ -64,7 +63,7 @@ suspend fun CameraView.takePhoto(optionsMap: ReadableMap): WritableMap {
   }
 }
 
-private fun writePhotoToFile(photo: CameraSession.CapturedPhoto, file: File) {
+private fun writePhotoToFile(photo: CameraSession.CapturedPhoto, file: File): Int {
   val byteBuffer = photo.image.planes[0].buffer
   if (photo.isMirrored) {
     val imageBytes = ByteArray(byteBuffer.remaining()).apply { byteBuffer.get(this) }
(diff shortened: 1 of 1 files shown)


mrousavy:
Ahh lol whoops!

1. Is there any reason we need to return the num of bytes written though? If not, let's remove it.
2. Is that because the buffer might be little or big endian? Not sure but this seems weird that we need to flip it 🤔

package/android/src/main/java/com/mrousavy/camera/CameraView+TakePhoto.kt
```diff
@@ -74,12 +73,19 @@ private fun writePhotoToFile(photo: CameraSession.CapturedPhoto, file: File) {
...
+      val didSucceed = processedBitmap.compress(Bitmap.CompressFormat.JPEG, 100, stream)
+      stream.close()
+      return if (didSucceed) imageBytes.size else 0
     }
   } else {
     val channel = FileOutputStream(file).channel
-    channel.write(byteBuffer)
+    val isReadyForWriting = byteBuffer.position() == 0
+    if (!isReadyForWriting) {
+      byteBuffer.flip()
```
mrousavy, line 84: Did you test this codebranch? Does that work? I think if `position()` isn't `0` you need to call `rewind()` in the ByteBuffer. But I wouldn't even know when position isn't 0 tbh...

package/android/src/main/java/com/mrousavy/camera/CameraView+TakePhoto.kt
```diff
@@ -74,12 +73,19 @@ private fun writePhotoToFile(photo: CameraSession.CapturedPhoto, file: File) {
...
+    }
```
mrousavy, line 85:
```suggestion
    if (byteBuffer.position() != 0) byteBuffer.rewind()
```

Because `flip` sets the limit to whatever position previously was, `rewind` sets the limit to the actual end of the ByteBuffer.

mrousavy:
Hey @vonovak ! Is this still relevant?

vonovak:
hey @mrousavy I didn't have a chance to properly test vision camera but from the usage that I did have, it didn't seem like this was needed. Kind of a mystery how I came up with this TBH :D

PR #1750, opened 2023-09-02
pass options to frame processors on iOS

rkmackinnon:
## What

This PR allows iOS frame processors to access passed-in options.
## Changes

This PR changes the frame processor callback arguments in FrameProcessorPluginHostObject.mm to pass in the options dictionary constructed prior instead of nil.
## Tested on
iPhone Xs

## Related issues
I didn't see any.


AZbang:
lol, very strange that this got into production

mrousavy:
Ahh, thanks so much @rkmackinnon, nice catch! 🙏

PR #1764, opened 2023-09-05
fix: store android images with correct orientation

vonovak:
## What

I noticed that Images captured on Android are not always stored with the right orientation. This is also mentioned in the release notes. I'm trying to fix that here. The PR needs testing on more devices, I just used Honor 20 Pro. If anyone can do more testing it'll be appreciated.

As a side note, this uses changes from https://github.com/mrousavy/react-native-vision-camera/pull/1749 - when they are merged the diff will be smaller.

## Tested on

To test this: take a picture and display it using expo-image. Regardless of how the phone is oriented (rotated up, right...) and whether or not the screen orientation is locked, the image should render "human friendly" - not with a wrong rotation.

Honor 20 Pro

### DIFF
--- a/package/android/src/main/java/com/mrousavy/camera/CameraView.kt
+++ b/package/android/src/main/java/com/mrousavy/camera/CameraView.kt
@@ -15,6 +15,7 @@ import androidx.core.content.ContextCompat
 import com.facebook.react.bridge.ReadableMap
 import com.mrousavy.camera.core.CameraSession
 import com.mrousavy.camera.core.PreviewView
+import com.mrousavy.camera.core.outputs.CameraOutputs
 import com.mrousavy.camera.extensions.containsAny
 import com.mrousavy.camera.extensions.installHierarchyFitter
 import com.mrousavy.camera.frameprocessor.FrameProcessor
@@ -22,7 +23,7 @@ import com.mrousavy.camera.parsers.Orientation
 import com.mrousavy.camera.parsers.PixelFormat
 import com.mrousavy.camera.parsers.Torch
 import com.mrousavy.camera.parsers.VideoStabilizationMode
-import com.mrousavy.camera.core.outputs.CameraOutputs
+import com.mrousavy.camera.utils.DeviceRotationLiveData
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
@@ -79,6 +80,7 @@ class CameraView(context: Context) : FrameLayout(context) {
   internal val cameraSession: CameraSession
   private var previewView: View? = null
   private var previewSurface: Surface? = null
+  private lateinit var deviceRotation: DeviceRotationLiveData
 
   internal var frameProcessor: FrameProcessor? = null
     set(value) {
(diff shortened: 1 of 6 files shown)


mrousavy:
Hey thanks so much for your PR man, this is amazing!

I'll test this to make sure it works for me, but just a quick tip that some Samsung devices handle orientation differently for whatever reason. See this blog post: https://medium.com/@kenodoggy/solving-image-rotation-on-android-using-camera2-api-7b3ed3518ab6

vonovak:
Haha, that's what I was afraid of 😄.
I'm new to working with cameras so this might need a few more iterations...
I'll read the blogpost and see what needs to be done, thanks 👍

package/android/src/main/java/com/mrousavy/camera/CameraViewManager.kt
```diff
@@ -182,10 +182,7 @@ class CameraViewManager : ViewGroupManager<CameraView>() {
     val cameraViewTransactions: HashMap<CameraView, ArrayList<String>> = HashMap()
 
     private fun addChangedPropToTransaction(view: CameraView, changedProp: String) {
-      if (cameraViewTransactions[view] == null) {
-        cameraViewTransactions[view] = ArrayList()
-      }
-      cameraViewTransactions[view]!!.add(changedProp)
+      cameraViewTransactions.getOrPut(view) { ArrayList() }.add(changedProp)
```
mrousavy, line 185: tbh I don't like that syntax. Kotlin has lots of weird ways to cram things into one line instead of being expressive. Can you revert that? Thanks!

package/android/src/main/java/com/mrousavy/camera/core/CameraSession.kt
```diff
@@ -213,6 +213,7 @@ class CameraSession(private val context: Context,
                         enableShutterSound: Boolean,
                         enableRedEyeReduction: Boolean,
                         enableAutoStabilization: Boolean,
+                        imageDataNeedsToBeRotatedByDegrees: Int,
```
mrousavy, line 216: I feel like this is a bit too verbose - can't we just move rotation to `this` (`CameraSession`), or does it have to be in `CameraView`? 

If not, can we rename it to something like `deviceOrientation`? I'm just not really happy with the name here

package/android/src/main/java/com/mrousavy/camera/core/CameraSession.kt
```diff
@@ -241,6 +241,8 @@ class CameraSession(private val context: Context,
       val isMirrored = cameraCharacteristics.get(CameraCharacteristics.LENS_FACING) == CameraCharacteristics.LENS_FACING_FRONT
 
       Log.i(TAG, "Photo capture 3/3 complete - received ${image.width} x ${image.height} image.")
+      val orientation = outputOrientation.toSensorRelativeOrientation(cameraCharacteristics)
```
mrousavy, line 244: Isn't that a second source of truth now? Is that still true?

package/android/src/main/java/com/mrousavy/camera/CameraView.kt
```diff
@@ -101,13 +115,20 @@ class CameraView(context: Context) : FrameLayout(context) {
     super.onAttachedToWindow()
     if (!isMounted) {
       isMounted = true
+      if (!::deviceRotation.isInitialized) {
```
mrousavy, line 118: What's `::`? Syntax looks pretty verbose here - is that this live data thingy?

mrousavy (changes requested):
hey - quick review on the airport, left some thoughts thanks

mrousavy:
Orientation is really tricky... we use #1891 to track orientation now

PR #1765, opened 2023-09-05
refactor: more swift-y PhotoCaptureDelegate

vonovak:
## What

This converts the PhotoCaptureDelegate to a more switf-y syntax.

I also _believe_ (I'm not using the `orientation` field) it fixes a bug in the `orientation` field:

previously:

```
case 5, 6:
      return "landscape-left"
```

but 5 and 6 actually correspond to `leftMirrored` and `right`
```
    case leftMirrored = 5 // 0th row on left,   0th column at top

    case right = 6 // 0th row on right,  0th column at top    - 90 deg CW
```

## Changes

This PR uses Swift constants instead of "magic numbers"

## Tested on

iPhone 11


mrousavy:
Looks great, is indeed more swifty! The syntax for getting the EXIF metadata is a bit ugly/verbose but can't really do much about that. Thanks again @vonovak!

PR #1782, opened 2023-09-07
fix: Consider everything between 24mm and 43mm a wide-angle lense

mrousavy:
## What

[On Wikipedia, a lens is considered to be wide-angle if it is smaller than 35mm and bigger than 24mm in focal length](https://en.wikipedia.org/wiki/Wide-angle_lens), but there are some devices like the Xiaomi A3 that have a focal length of 37mm on their "default" camera.

Previously, that code made that specific device be detected as a telephoto, while it really doesn't look like a telephoto Camera. This meant, that code that used `useCameraDevices('wide-angle')` would break, since there is no matching Camera on that device, so it would show a blackscreen. In a future PR I want to rework that hook to make it easier and logical to get a Camera device without such mistakes.

So now I increased this value from `35` to `43`, a value which seemed more logical to me. If a lens has a focal length of `43`, I would still say this is a wide-angle camera, not a telephoto.

These are now the new values for device types / focal lengths:

* 0mm-24mm: `ultra-wide-angle` device
* 24mm-43mm: `wide-angle` device
* 43mm-∞mm: `telephoto` device

## Tested on

* Xiaomi A3

### DIFF
--- a/package/android/src/main/java/com/mrousavy/camera/core/CameraDeviceDetails.kt
+++ b/package/android/src/main/java/com/mrousavy/camera/core/CameraDeviceDetails.kt
@@ -110,6 +110,7 @@ class CameraDeviceDetails(private val cameraManager: CameraManager, private val 
         l in 24f..43f -> deviceTypes.pushString("wide-angle-camera")
         // https://en.wikipedia.org/wiki/Telephoto_lens
         l > 43f -> deviceTypes.pushString("telephoto-camera")
+        else -> throw Error("Invalid focal length! (${focalLength}mm)")
       }
     }
 
--- a/package/android/src/main/java/com/mrousavy/camera/core/CameraDeviceDetails.kt
+++ b/package/android/src/main/java/com/mrousavy/camera/core/CameraDeviceDetails.kt
@@ -100,19 +100,18 @@ class CameraDeviceDetails(private val cameraManager: CameraManager, private val 
 
     val deviceTypes = Arguments.createArray()
 
-    // https://en.wikipedia.org/wiki/Telephoto_lens
-    val containsTelephoto = focalLengths.any { l -> (l * cropFactor) > 43 } // TODO: Telephoto lenses are > 85mm, but we don't have anything between that range..
-    // https://en.wikipedia.org/wiki/Wide-angle_lens
-    val containsWideAngle = focalLengths.any { l -> (l * cropFactor) >= 24 && (l * cropFactor) <= 43 }
-    // https://en.wikipedia.org/wiki/Ultra_wide_angle_lens
-    val containsUltraWideAngle = focalLengths.any { l -> (l * cropFactor) < 24 }
-
-    if (containsTelephoto)
-      deviceTypes.pushString("telephoto-camera")
-    if (containsWideAngle)
-      deviceTypes.pushString("wide-angle-camera")
-    if (containsUltraWideAngle)
-      deviceTypes.pushString("ultra-wide-angle-camera")
+    focalLengths.forEach { focalLength ->
+      // scale to the 35mm film standard
+      val l = focalLength * cropFactor
+      when {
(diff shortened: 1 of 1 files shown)


mrousavy:
cc @aldo-expensify

package/android/src/main/java/com/mrousavy/camera/core/CameraDeviceDetails.kt
```diff
@@ -95,27 +95,24 @@ class CameraDeviceDetails(private val cameraManager: CameraManager, private val
...
+      // scale to the 35mm film standard
+      val l = focalLength * cropFactor
+      when {
+        // https://en.wikipedia.org/wiki/Ultra_wide_angle_lens
+        l < 24f -> deviceTypes.pushString("ultra-wide-angle-camera")
+        // https://en.wikipedia.org/wiki/Wide-angle_lens
+        l in 24f..43f -> deviceTypes.pushString("wide-angle-camera")
+        // https://en.wikipedia.org/wiki/Telephoto_lens
+        l > 43f -> deviceTypes.pushString("telephoto-camera")
+        else -> throw Error("Invalid focal length! (${focalLength}mm)")
```
aldo-expensify, line 113: It is kind of impossible for this to happen, right? unless `l` is not a number at all, which I guess can't be, right?

package/android/src/main/java/com/mrousavy/camera/core/CameraDeviceDetails.kt
```diff
@@ -95,27 +95,24 @@ class CameraDeviceDetails(private val cameraManager: CameraManager, private val
...
+    focalLengths.forEach { focalLength ->
+      // scale to the 35mm film standard
+      val l = focalLength * cropFactor
+      when {
+        // https://en.wikipedia.org/wiki/Ultra_wide_angle_lens
+        l < 24f -> deviceTypes.pushString("ultra-wide-angle-camera")
+        // https://en.wikipedia.org/wiki/Wide-angle_lens
+        l in 24f..43f -> deviceTypes.pushString("wide-angle-camera")
+        // https://en.wikipedia.org/wiki/Telephoto_lens
+        l > 43f -> deviceTypes.pushString("telephoto-camera")
```
aldo-expensify, line 112: Can `focalLengths` contain more than one element? if it can, you may end up pushing repeated types to  `deviceTypes`. The old code handled this better I think unless `focalLengths` contains always a single element.

package/android/src/main/java/com/mrousavy/camera/core/CameraDeviceDetails.kt
mrousavy: yea, better safe than sorry i dont know 🤣

package/android/src/main/java/com/mrousavy/camera/core/CameraDeviceDetails.kt
mrousavy: Here's the thing; technically there could be a single camera device that had two similar cameras (eg two wide angle cameras; for example one for depth and one for color).

With the old code, that resulted in [wide-angle] (1x), with the new code that results in [wide-angle, wide-angle] (2x), being more accurate.

Do you think that's confusing?

aldo-expensify: > Do you think that's confusing?

Not necessarily, I think it should just be mentioned in the documentation (is there documentation?)

aldo-expensify (approved):
Approving since the potentially repeated device types is by design

package/android/src/main/java/com/mrousavy/camera/core/CameraDeviceDetails.kt
aldo-expensify: Wondering if it qualifies as "breaking changes".

PR #1783, opened 2023-09-07
docs: Upgrade to latest Docusaurus/Typedoc

mrousavy:


mrousavy:
Oh no, it now breaks with SSR. I have no idea how SSR works lol. This is the output of `yarn build` in `docs/`:

```
[ERROR] Docusaurus server-side rendering could not render static page with path /docs/api/.
[ERROR] Docusaurus server-side rendering could not render static page with path /docs/api/classes/Camera.
[ERROR] Docusaurus server-side rendering could not render static page with path /docs/api/classes/CameraCaptureError.
[ERROR] Docusaurus server-side rendering could not render static page with path /docs/api/classes/CameraRuntimeError.
[ERROR] Docusaurus server-side rendering could not render static page with path /docs/api/interfaces/CameraDevice.
```

@slorber sorry for pinging you here, but do you maybe have any idea what's going wrong here?

mrousavy:
Screenshot 2023-09-07 at 12 02 04

docs/package.json
```diff
@@ -12,13 +12,13 @@
...
-    "@docusaurus/plugin-sitemap": "^2.3.1",
-    "@docusaurus/preset-classic": "^2.3.1",
-    "@mdx-js/react": "^1.6.22",
-    "@vercel/analytics": "^0.1.11",
-    "clsx": "^1.2.1",
+    "@docusaurus/core": "^2.4.1",
+    "@docusaurus/plugin-google-gtag": "^2.4.1",
+    "@docusaurus/plugin-sitemap": "^2.4.1",
+    "@docusaurus/preset-classic": "^2.4.1",
+    "@mdx-js/react": "^2.3.0",
```
slorber, line 19: MDX v2 only works with Docusaurus v3, that's probably the problem.

Also we don't maintain the typedoc plugin on the Docusaurus repo, if there are issues on api pages I'm not sure I'll be able to help much on that, I don't really know how this plugin works 😅

mrousavy: gotcha, well I guess I'll stick with the older version for now then, thank you sebastian 🙏

mrousavy: How do I get the full error logs/stacktrace? Is there a verbose param?

slorber: Hmm I think I fixed that but maybe it's only in v3 or latest v2?

Not sure why it's so complicated to get full stacktraces printed 😅 this has always been annoying to me.

PR #1786, opened 2023-09-08
fix: missing plugin name in assertion

xseignard:
## What

The assertion message that ensures frame processors name uniqueness on Android is missing the name to be shown.

**Before**
Capture d’écran 2023-09-08 à 17 00 39

**After**
Capture d’écran 2023-09-08 à 17 38 10

## Tested on
Pixel 5/API 33 emulator


mrousavy:
Nice catch, thanks!

PR #1789, opened 2023-09-11
perf: Use `ImageWriter` instead of OpenGL Pipeline for faster processing

mrousavy:
## What

Previously, the C++ OpenGL pipeline internally converted the Buffers to RGB, and then did a two-pass rendering step to pass them to the Frame Processor and to the MediaRecorder.

Apparently `ImageWriter` does actually not use OpenGL rendering under the hood, but instead just passes the native graphics buffers around, which would mean that it is **much more efficient**, **more performant**, and works with every Image format (`yuv`, `rgb` and even `private`).

So this PR replaces the OpenGL pipeline with an `ImageWriter` setup.


mrousavy:
Can anyone do some CPU profiling/memory consumption profiling/GPU profiling and compare before this PR with after this PR? I wonder if this PR is more efficient than the OpenGL C++ GPU pipeline...

jjbwii:
Hello,

it sounds good!!!
Perhaps it would be a good idea to see how the “androidx way” copy camera buffers.  I am going to search and open some files of the old way. I am curious… perhaps it use the opengl or perhaps use the cpu approach…

PR #1798, opened 2023-09-13
fix: Find closest matching aspect ratio for Preview

mrousavy:


mrousavy:
Closing in favor of #1817

PR #1799, opened 2023-09-13
fix: Fix reference-counting on RecordingSession `ImageWriter`

mrousavy:


mrousavy:
Ah this introduces a crash when trying to start recording a video. Something's going wrong with the `ImageWriter` writing to the `MediaRecorder`'s surface....

PR #1800, opened 2023-09-13
docs: don't try to re-add an already registered plugin on Android

xseignard:
## What

Update android docs to propose how to avoid re-adding an existing frame processor plugin that would trigger the following assertion: https://github.com/mrousavy/react-native-vision-camera/blob/main/package/android/src/main/java/com/mrousavy/camera/frameprocessor/FrameProcessorPluginRegistry.java#L17

## Changes

Rely on a simple boolean flag

## Tested on

Pixel5/API 33


mrousavy:
Hey sorry for long radio silence here but thanks for your PR! We no longer have this code so this is obsolete I think.. sorry!

PR #1817, opened 2023-09-19
✨ feat: Respect format's aspect ratio in Preview

blancham:
## What

* This PR make the camera preview respect the aspect ratio of the given format
* This PR allows the camera preview to fit whatever container it is in

## Changes

* This PR updates the way `getPreviewSize()` function works
* This PR updates the way the `PreviewView` is measured (thus rendered on the screen)

## Tested on

* Vivo Y01 (Android 12)

## Related issues

* Related to https://github.com/mrousavy/react-native-vision-camera/pull/1798

## Description

A year ago I worked with `react-native-vision-camera` (v2.15.2) in order to replace the deprecated `react-native-camera` in our app at Kraaft.

I managed to do some pretty nice things with it (thanks @mrousavy) but when testing on Android, we realised that there were a lot of small bugs making it impossible to ship the app (lot of regressions). 
The preview not being accurate on some devices (Vivo being one of the worst case) was the most important one.

This has been in stand-by for quite a long time, hoping that v3.0.0 would solve most of the issues and it did!

...except for the preview 😄 

So two weeks ago I decided to setup a sandbox app and try to solve it myself (I ain't no Kotlin developer sorry).
That's what I came up with, would be glad to chat about it if you think some parts are weirdly done 😭

On the pictures below, I have a `16:9` aspect ratio, when taking the picture the image is basically the same as the preview.

On the current version of the library, the preview overflows the screen

| Before | After |
|---|---|
|Preview<br><br><br>Photo<br>|Preview<br><br><br>Photo<br>|


mrousavy:
Hey Maxime, first of all huge thanks for taking a stab at this. I absolutely love this PR!!! 😍😍😍😍 

Also really nice to hear that you guys are using VisionCamera in your app and that V3 solved a bunch of stuff for you guys. It's still early so I appreciate any feedback I get!

I tried to implement this by moving everything into `PreviewView` in my other PR, but having `previewSize` as a computed prop on `CameraView` makes much more sense. This is amazing.

Quick question before I do the code review: Does this still do a aspectRatio: 'cover' for the Preview, aka a center crop fill? In your screenshots it looks like aspectRatio: 'contain' instead of 'cover', or is that just your app's UI? 🤔 

e.g. in the Example app, does it fill the screen accordingly and does the resulting video have the same aspect ratio as the preview? I think we should use the same aspect ratio for preview as we do for the output video. Photo is a different pipeline, the user can choose a format where photo aspect == video aspect. 

If that's the case, then I'll review all of that in a second, but at first glance it looks pretty solid!

blancham:
Hello Marc !

Thanks for the quick reply 😁

The idea behind this enhancement is : **what you see is what you get**. The preview will always try to fit whatever container it's in, using _contain_ mode.

In the current version of the lib, the preview is overflowing which makes it harder to control its placement.

|BEFORE|AFTER|
|---|---|
|Size:652x250px<br><br>Size:224x243px<br>|Size:652x250px<br><br>Size:224x243px<br>|

As of your Example app, here is how it looks :
|BEFORE|AFTER|
|---|---|
|||

I played a bit with it and if you really want to you can have a full screen camera but it's only done on the JS side, you can tweak the style you give to the component (instead of using `StyleSheet.absoluteFill`)

Finally, I will update my code to use the video aspect ratio instead of the photo's. Makes more sense, you are right!

mrousavy:
I see, thanks for the insights!

This is aspectRatio: 'contain', but for a Camera we (I think) always want aspectRatio: 'cover'. In my experience working with Camera apps, that's always what the user expects.
Black bars around the Camera (as seen in the Example app in your screenshot, the Preview just stops at the bottom) are weird.
This is btw also how it works on iOS, it does a center crop (aka aspectRatio: 'cover') so the PreviewView covers the entire View.

Is there any chance you could update your PR to do a center crop (aspectRatio: 'cover') instead of the fit/contain mode?

image

package/android/src/main/java/com/mrousavy/camera/core/outputs/CameraOutputs.kt
```diff
@@ -28,7 +28,7 @@ class CameraOutputs(val cameraId: String,
     const val PHOTO_OUTPUT_BUFFER_SIZE = 3
   }
 
-  data class PreviewOutput(val surface: Surface)
+  data class PreviewOutput(val surface: Surface, val aspectRatio: Double? = null)
```
mrousavy: I think we probably want this to be `targetSize`, not `aspectRatio`.

package/android/src/main/java/com/mrousavy/camera/extensions/CameraCharacteristics+getPreviewSize.kt
```diff
@@ -0,0 +1,42 @@
...
+  val outputSizes = config.getOutputSizes(SurfaceHolder::class.java).sortedByDescending { it.width * it.height }
+
+  return outputSizes.first { it.bigger <= maximumPreviewSize.bigger && it.smaller <= maximumPreviewSize.smaller }
+}
+
+fun CameraCharacteristics.getPreviewSize(aspectRatio: Double?): Size {
+  return if (aspectRatio != null)
+    getPreviewSizeFromAspectRatio(aspectRatio)
+  else
+    getAutomaticPreviewSize()
```
mrousavy:
```suggestion
  return if (aspectRatio != null) {
    getPreviewSizeFromAspectRatio(aspectRatio)
  } else {
    getAutomaticPreviewSize()
  }
```

package/android/src/main/java/com/mrousavy/camera/extensions/CameraCharacteristics+getPreviewSize.kt
```diff
@@ -0,0 +1,42 @@
...
+import android.util.Size
+import android.view.SurfaceHolder
+import kotlin.math.abs
+
+private fun getMaximumPreviewSize(): Size {
+  // See https://developer.android.com/reference/android/hardware/camera2/params/StreamConfigurationMap
+  // According to the Android Developer documentation, PREVIEW streams can have a resolution
+  // of up to the phone's display's resolution, with a maximum of 1920x1080.
+  val display1080p = Size(1920, 1080)
+  val displaySize = Size(Resources.getSystem().displayMetrics.widthPixels, Resources.getSystem().displayMetrics.heightPixels)
```
mrousavy:
```suggestion
  val displaySize = Size(Resources.getSystem().displayMetrics.widthPixels,
                         Resources.getSystem().displayMetrics.heightPixels)
```

package/android/src/main/java/com/mrousavy/camera/extensions/CameraCharacteristics+getPreviewSize.kt
```diff
@@ -0,0 +1,42 @@
...
+  val isHighResScreen = displaySize.bigger >= display1080p.bigger || displaySize.smaller >= display1080p.smaller
+
+  return if (isHighResScreen) display1080p else displaySize
+}
+
+fun CameraCharacteristics.getPreviewSizeFromAspectRatio(aspectRatio: Double): Size {
+  val config = this.get(CameraCharacteristics.SCALER_STREAM_CONFIGURATION_MAP)!!
+  val maximumPreviewSize = getMaximumPreviewSize()
+  val outputSizes = config.getOutputSizes(SurfaceHolder::class.java).sortedByDescending { it.width * it.height }.sortedBy { abs(aspectRatio - (it.bigger.toDouble() / it.smaller)) }
```
mrousavy:
```suggestion
  val outputSizes = config.getOutputSizes(SurfaceHolder::class.java)
    .sortedByDescending { it.width * it.height }
    .sortedBy { abs(aspectRatio - (it.bigger.toDouble() / it.smaller)) }
```

package/android/src/main/java/com/mrousavy/camera/extensions/CameraCharacteristics+getPreviewSize.kt
```diff
@@ -0,0 +1,42 @@
+package com.mrousavy.camera.extensions
```
mrousavy, line 1: The Preview size is also an output size, was there a specific reason you moved this to a separate file?

package/android/src/main/java/com/mrousavy/camera/core/PreviewView.kt
```diff
@@ -45,23 +42,25 @@ class PreviewView(context: Context,
...
     } else {
-      newWidth = width
-      newHeight = (width / actualRatio).roundToInt()
+      newHeight = (viewWidth / aspectRatio).roundToInt()
     }
 
-    Log.d(TAG, "Measured dimensions set: $newWidth x $newHeight")
+    Log.d(TAG, "Fitted dimensions set: $newWidth x $newHeight")
     setMeasuredDimension(newWidth, newHeight)
   }
```
mrousavy: We want to use a center-crop (aspectRatio: 'cover') transform here, not a aspectRatio: 'contain'. The Camera should fill the entire parent View and be horizontally or vertically centered, just like this:

image

package/android/src/main/java/com/mrousavy/camera/core/PreviewView.kt
```diff
@@ -8,22 +8,19 @@ import android.util.Size
...
-  private val targetSize: Size
-  private val aspectRatio: Float
-    get() = targetSize.width.toFloat() / targetSize.height.toFloat()
+  private val aspectRatio: Double
+    get() = targetSize.smaller.toDouble() / targetSize.bigger
 
   init {
-    val characteristics = cameraManager.getCameraCharacteristics(cameraId)
-    targetSize = characteristics.getPreviewSize()
-
```
mrousavy: Nice, I like that we can get rid of passing around cameraManager/cameraId here! :)

package/android/src/main/java/com/mrousavy/camera/CameraView.kt
```diff
@@ -173,13 +191,15 @@ class CameraView(context: Context) : FrameLayout(context) {
...
+      val previewSize = previewSize ?: return
+
       val format = format
       val targetVideoSize = if (format != null) Size(format.getInt("videoWidth"), format.getInt("videoHeight")) else null
       val targetPhotoSize = if (format != null) Size(format.getInt("photoWidth"), format.getInt("photoHeight")) else null
       // TODO: Allow previewSurface to be null/none
       val previewSurface = previewSurface ?: return
 
-      val previewOutput = CameraOutputs.PreviewOutput(previewSurface)
+      val previewOutput = CameraOutputs.PreviewOutput(previewSurface, previewSize.width.toDouble() / previewSize.height)
```
mrousavy: Again, targetSize, not aspectRatio imo.

package/android/src/main/java/com/mrousavy/camera/CameraView.kt
```diff
@@ -173,13 +191,15 @@ class CameraView(context: Context) : FrameLayout(context) {
       }
       val cameraId = cameraId ?: throw NoCameraDeviceError()
 
+      val previewSize = previewSize ?: return
```
mrousavy: This should probably be `previewView.size`, right? Single source of truth - if we attach a PreviewView we need exactly this size - otherwise it could go out of sync.

package/android/src/main/java/com/mrousavy/camera/CameraView.kt
```diff
@@ -79,6 +82,7 @@ class CameraView(context: Context) : FrameLayout(context) {
   internal val cameraSession: CameraSession
   private var previewView: View? = null
   private var previewSurface: Surface? = null
+  private var previewSize: Size? = null
```
mrousavy: This should be removed, the single source of truth should be `previewView?.targetSize` - that way we can't go out of sync :)

package/android/src/main/java/com/mrousavy/camera/CameraView.kt
```diff
@@ -111,12 +115,26 @@ class CameraView(context: Context) : FrameLayout(context) {
     updateLifecycle()
   }
 
+  private fun getPreviewSize(): Size {
+    val cameraId = cameraId ?: throw NoCameraDeviceError()
+
+    val format = format
+    val targetPhotoSize = if (format != null) Size(format.getInt("photoWidth"), format.getInt("photoHeight")) else null
```
mrousavy: Should be video pipeline size, not photo

package/android/src/main/java/com/mrousavy/camera/CameraView.kt
```diff
@@ -111,12 +115,26 @@ class CameraView(context: Context) : FrameLayout(context) {
...
+    val formatAspectRatio = if (targetPhotoSize != null) targetPhotoSize.bigger.toDouble() / targetPhotoSize.smaller else null
+
+    return this.cameraManager.getCameraCharacteristics(cameraId).getPreviewSize(formatAspectRatio)
```
mrousavy: Nice!

package/android/src/main/java/com/mrousavy/camera/CameraView.kt
mrousavy: Also, by default, if no `format` is passed, the `VideoOutput` uses the largest (in resolution) format for the YUV format (getVideoOutputSize then maxBy width * height). We should use that as a fallback instead of `null` here.

package/android/src/main/java/com/mrousavy/camera/CameraView.kt
```diff
@@ -111,12 +115,26 @@ class CameraView(context: Context) : FrameLayout(context) {
...
+
   private fun setupPreviewView() {
     removeView(previewView)
     this.previewSurface = null
 
     val cameraId = cameraId ?: return
-    val previewView = PreviewView(context, cameraManager, cameraId) { surface ->
+
+    val previewSize = this.getPreviewSize()
+    this.previewSize = previewSize
```
mrousavy: again remove the member var, only pass it to `PreviewView` ctor for single source of truth

mrousavy (changes requested):
Hey - reviewed the code :)

overall looks good, only some minor code changes and it needs to be aspectRatio cover imo.

blancham:
Okay, I'll update my PR to fit the container with a `cover` method. In our case we need to change the ratio on the go so having a `contain` method is more coherent but I'll be able to achieve the same behaviour even with a `cover` (using JS styling) so it's fine.

Also, iPhone's aren't using a `cover` method, at least not on my iPhone 13 PRO Max. At first glance seems so but upon further inspection, the bottom and top parts (which are not in the final picture), are visually added using another camera view.

In the below picture you can see that the bottle opener is visible on the two cameras (don't know if it makes sense). The idea is that if it were really only one `cover` camera preview, it would have not been shown like it is.

IMG_3403

package/android/src/main/java/com/mrousavy/camera/extensions/CameraCharacteristics+getPreviewSize.kt
blancham: Not at all, I'm new to _Kotlin_ and unfamiliar with the concept of Extensions file. If you think that both files should be merged back, I can 🤓

package/android/src/main/java/com/mrousavy/camera/CameraView.kt
blancham: There is a function called `getAutomaticPreviewSize` that is called when `aspectRatio`, it uses the format with the highest resolution. I feel like it's fine like that so we don't depend on the logic of CameraCharacteristics in CameraView

mrousavy:
Good point. I always used Instagram and Snapchat for comparison.

Maybe we could add a `aspectRatio="cover | contain"` prop? I can fully do the iOS implementation.

Also, unrelated to that - regarding your first screenshot which showed my 'cover' implementation - is that a bug? It only extends below the Frame, not above it 🤔

mrousavy:
Also, that burger in your camera roll looks tasty asf

blancham:
We definitely should have a prop to specify the resize mode, I'm glad your are okay with that 😁

For the overflow, to be fair I didn't deep dive into why it was showing “like that”, it's probably related to the style I'm using in my sandbox app. Had to do some weird stuff in order to have the camera taking only the space it should. I could upload a gist with my code if you wanna test it yourself 😉

I am struggling a bit with the "cover" implementation in Kotlin. I find it hard to understand how to tweak the `onMeasure` in order to render the Preview with an overflow. I'll still be working on it tomorrow so it should be fine 💪🏼



PS: the burger was indeed amazing 😋

jlsneto:
wow!

blancham:
Okay @mrousavy I implemented the `resizeMode` prop and it's working great!

|COVER|CONTAIN|
|---|---|
|||

The **orange** rectangle is the `View` the `Camera` is in 🤓


Also we'll need to find a permanent solution for when computing the aspect ratio of the format, I found that on some devices I get `size: 1280x720` which means `width: 1280, height: 720` and that's incorrect because I am in portrait mode 😔

In order to make it work with my device I had to invert `width` and `height` here :

```kt
val contentAspectRatio = contentSize.height.toDouble() / contentSize.width
```

Do you know how we can make this work reliably on every device?

mrousavy:
@blancham wow, amazing stuff and really nice visual examples!!!!! ❤️❤️❤️❤️❤️❤️❤️

Regarding width/height; yes this is expected! Camera Sensors are usually in landscape orientation (90deg rotated relative to device). I say _usually_, because I'm sure there's some weird Android phones that don't have it that way, but I haven't seen any yet.

Do you need to adjust anything in your code given that new information of Camera sensor orientation? If not, we're good to merge and I'll add iOS support in a follow-up PR

package/android/src/main/java/com/mrousavy/camera/core/outputs/CameraOutputs.kt
```diff
@@ -101,7 +104,11 @@ class CameraOutputs(
     // Preview output: Low resolution repeating images (SurfaceView)
     if (preview != null) {
       Log.i(TAG, "Adding native preview view output.")
-      previewOutput = SurfaceOutput(preview.surface, characteristics.getPreviewSize(), SurfaceOutput.OutputType.PREVIEW)
+      val previewSizeAspectRatio = if (preview.targetSize != null) preview.targetSize.bigger.toDouble() / preview.targetSize.smaller else null
```
mrousavy, line 107: I'm not sure if we should pass aspect ratios around - it just feels a bit dangerous to have `aspectRatio: Int` being passed, because:

1. We lose orientation (no information if this is `width / height` or `height / width`)
2. We lose context - is it the video's aspect ratio? photo's? preview view's?

Is it possible that we can somehow simplify this, or do you think that's required?

mrousavy:
@blancham what's your Twitter?

mrousavy:
merged! follow up PR for iOS: https://github.com/mrousavy/react-native-vision-camera/pull/1838 💪 

thanks again @blancham! amazing stuff

blancham:
My Twitter (X) is [@nqxine](https://twitter.com/nqxine) (new account  just for you ahah)

I'm glad you are happy with this work 😉

package/android/src/main/java/com/mrousavy/camera/core/outputs/CameraOutputs.kt
blancham: We can still change this to have like `getPreviewTargetSizeFromFormatSize`. We don't **need** to have an aspect ratio if we can compute it from the size!

WissamCodit:
Hello, I'm encountering a similar issue. When attempting to take a photo of a document, the resulting image appears zoomed out compared to the preview. Could someone offer assistance with this discrepancy?



 <Camera
            ref={camera}
            isActive={isActive && isFocused}
            device={device}
            photo={true}
            format={cameraFormat}
            pixelFormat="yuv"
            torch={isActive ? 'on' : 'off'}
            exposure={0}
            style={StyleSheet.absoluteFill}
            onError={onError}
          />
  const device = useCameraDevice('back', {
    physicalDevices: [
      'ultra-wide-angle-camera',
      'wide-angle-camera',
      'telephoto-camera',
    ],
  });

  const isFocused = useIsFocused();
  const cameraFormat = useCameraFormat(device, [
    {videoResolution: {width: 1280, height: 720}},
    // {videoAspectRatio: aspectRatio},
    // {photoAspectRatio: aspectRatio},

    {fps: 60},
  ]);

PR #1848, opened 2023-09-25
fix: Allow `getCameraDevice` to return `undefined` when no Devices are available (e.g. iOS Simulator)

### LINKED ISSUE #1847: 🐛 No camera device found in IOS simulator
### What's happening?

image

### Reproduceable Code

```tsx
const device = useCameraDevice(cameraPosition, {
    physicalDevices: [
      'ultra-wide-angle-camera',
      'wide-angle-camera',
      'telephoto-camera',
    ],
  });
// this hook should check if we are in the simulator it should return null or undefined, without throwing an error in case of a simulator.
```

### Relevant log output

```shell
No camera device could be found!
```

### Camera Device

```json
{

}
```

### Device

Simulator iPhone 15

### VisionCamera Version

3.1.0

### Can you reproduce this issue in the VisionCamera Example app?

I didn't try (⚠️ your issue might get ignored & closed if you don't try this)

### Additional information

- [X] I am using Expo
- [X] I have enabled Frame Processors (react-native-worklets-core)

mrousavy:
## What

On an iOS simulator, `getAvailableCameraDevices()` returns `[]` (there are no cameras unfortunately).

This PR returns `undefined` instead of throwing an error if no device can be found.

## Related issues

- Fixes #1847 

cc @hamdij0maa

### DIFF
--- a/package/src/devices/getCameraDevice.ts
+++ b/package/src/devices/getCameraDevice.ts
@@ -1,5 +1,4 @@
 import { CameraDevice, CameraPosition, PhysicalCameraDeviceType } from '../CameraDevice';
-import { CameraRuntimeError } from '../CameraError';
 
 export interface DeviceFilter {
   /**
@@ -24,7 +23,7 @@ export interface DeviceFilter {
 }
 
 /**
- * Get the best matching Camera device that best satisfies your requirements using a sorting filter.
+ * Get the best matching Camera device that best satisfies your requirements using a sorting filter, or `undefined` if {@linkcode devices} does not contain any devices.
  * @param position The position of the Camera device relative to the phone.
  * @param filter The filter you want to use. The Camera device that matches your filter the closest will be returned
  * @returns The Camera device that matches your filter the closest, or `undefined` if no such Camera Device exists on the given {@linkcode position}.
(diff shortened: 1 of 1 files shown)


hamdij0maa:
<3 🚀

hirbod:
@mrousavy just an idea: could it be possible to have a look how [Tesla Motors](https://github.com/teslamotors/react-native-camera-kit) achieves fake camera previews inside the simulator with their RN module? Maybe it could be copied

https://github.com/teslamotors/react-native-camera-kit/blob/master/ios/ReactNativeCameraKit/SimulatorCamera.swift
https://github.com/teslamotors/react-native-camera-kit/blob/master/ios/ReactNativeCameraKit/SimulatorPreviewView.swift

mrousavy:
@hirbod unfortunately not, no. Also that lib was written by Wix and just got moved to Tesla.

The reason why this will not work is because VisionCamera builds upon Camera Devices, allowing the user to select any given camera device. I cannot simulate a fake device, as that would mean I have to:

1. Create a dummy `CameraDevice` specifically for iOS simulators
2. Add dummy implementation for every single Camera Device specific function (start session, enable flash, etc) because those depend on the low-level hardware device type
3. Resolve takePhoto and recordVideo with dummy files
4. Simulate a fake preview using Metal (that's surprisingly the easy part)

Which is a shit ton of extra code. Like really, a shit ton. 

The real solution would be to add support to Cameras on iOS simulators (e.g the macbooks camera, I am surprised this still doesn't work.)

hirbod:
Yeah, makes sense. That extra work is indeed too much.

xts-bit:
@hirbod @mrousavy Like Will it support FaceTime support on Macs?

PR #1873, opened 2023-09-27
fix: Use Video RGB pixel format (`0x1` instead of `0x100`)

mrousavy:
## What

Previously the JPEG format (0x100) was used, when instead we should've used RGBA_8888 (0x1).

This fixes a crash.


rodgomesc:
overtaken by #1875 ?

PR #1877, opened 2023-09-27
chore: improved FP docs

mateusz1913:
## What

I split the changes from initial pull request for v3 plugins examples https://github.com/mrousavy/react-native-vision-camera/pull/1739/files - this PR includes improved Frame Processor docs

## Changes

- improved Swift FP plugins creation section
- small improvements to Android FP plugins creation section & FP plugins overview

### DIFF
--- a/.github/workflows/build-ios.yml
+++ b/.github/workflows/build-ios.yml
@@ -56,7 +56,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: |
-            example/ios/Pods
+            package/example/ios/Pods
             ~/Library/Caches/CocoaPods
             ~/.cocoapods
           key: ${{ runner.os }}-pods-${{ hashFiles('**/Podfile.lock') }}
@@ -117,7 +117,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: |
-            example/ios/Pods
+            package/example/ios/Pods
             ~/Library/Caches/CocoaPods
             ~/.cocoapods
           key: ${{ runner.os }}-pods-${{ hashFiles('**/Podfile.lock') }}
--- a/docs/docs/guides/FRAME_PROCESSORS_CREATE_OVERVIEW.mdx
+++ b/docs/docs/guides/FRAME_PROCESSORS_CREATE_OVERVIEW.mdx
@@ -51,8 +51,9 @@ Similar to a TurboModule, the Frame Processor Plugin Registry API automatically 
 Return values will automatically be converted to JS values, assuming they are representable in the ["Types" table](#types). So the following Java Frame Processor Plugin:
 
 ```java
+@Nullable
 @Override
-public Object callback(Frame frame, Object[] params) {
+public Object callback(@NonNull Frame frame, @Nullable Map<String, Object> arguments) {
   return "cat";
 }
 ```
@@ -70,8 +71,9 @@ export function detectObject(frame: Frame): string {
 You can also manipulate the buffer and return it (or a copy of it) by returning a [`Frame`][2]/[`Frame`][3] instance:
 
 ```java
+@Nullable
 @Override
-public Object callback(Frame frame, Object[] params) {
+public Object callback(@NonNull Frame frame, @Nullable Map<String, Object> arguments) {
   Frame resizedFrame = new Frame(/* ... */);
   return resizedFrame;
 }
@@ -107,12 +109,13 @@ const frameProcessor = useFrameProcessor((frame) => {
 To let the user know that something went wrong you can use Exceptions:
 
 ```java
+@Nullable
 @Override
-public Object callback(Frame frame, Object[] params) {
-  if (params[0] instanceof String) {
+public Object callback(@NonNull Frame frame, @Nullable Map<String, Object> arguments) {
+  if (arguments != null && arguments.get("codeType") instanceof String) {
     // ...
   } else {
-    throw new Exception("First argument has to be a string!");
+    throw new Exception("codeType property has to be a string!");
   }
 }
 ```
@@ -123,7 +126,7 @@ Which will throw a JS-error:
 const frameProcessor = useFrameProcessor((frame) => {
   'worklet'
   try {
-    const codes = scanCodes(frame, true)
+    const codes = scanCodes(frame, { codeType: 1234 })
   } catch (e) {
     console.log(`Error: ${e.message}`)
   }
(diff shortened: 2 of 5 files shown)


.github/workflows/build-ios.yml
```diff
@@ -117,7 +117,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: |
-            example/ios/Pods
+            package/example/ios/Pods
```
mrousavy, line 120: Are you sure that didn't use `package/` as a default path already? I need to double check in the CI if that works today

docs/docs/guides/FRAME_PROCESSORS_CREATE_OVERVIEW.mdx
```diff
@@ -51,8 +51,9 @@ Similar to a TurboModule, the Frame Processor Plugin Registry API automatically
 Return values will automatically be converted to JS values, assuming they are representable in the ["Types" table](#types). So the following Java Frame Processor Plugin:
 
 ```java
+@Nullable
 @Override
-public Object callback(Frame frame, Object[] params) {
+public Object callback(@NonNull Frame frame, @Nullable Map<String, Object> arguments) {
```
mrousavy, line 56: Hmm I think leaving out the `@NonNull` and `@Nullable` props is easier to read for non-native developers, this is pretty verbose. Wdyt?

docs/docs/guides/FRAME_PROCESSORS_CREATE_OVERVIEW.mdx
```diff
@@ -123,7 +126,7 @@ Which will throw a JS-error:
 const frameProcessor = useFrameProcessor((frame) => {
   'worklet'
   try {
-    const codes = scanCodes(frame, true)
+    const codes = scanCodes(frame, { someString: true })
```
mrousavy: Maybe let's use more accurate examples? E.g. `codeType: 'qr' | 'barcode'`.

Or maybe we wanna refactor this to object detection with `objectType: 'banana'`. I'll think about this

docs/docs/guides/FRAME_PROCESSOR_CREATE_FINAL.mdx
```diff
@@ -18,6 +18,7 @@ const plugin = VisionCameraProxy.getFrameProcessorPlugin('scanFaces')
  */
 export function scanFaces(frame: Frame): object {
   'worklet'
+  if (plugin == null) throw new Error("Failed to load Frame Processor Plugin!")
```
mrousavy:
```suggestion
  if (plugin == null) throw new Error('Failed to load Frame Processor Plugin "scanFaces"!')
```

docs/docs/guides/FRAME_PROCESSORS_CREATE_OVERVIEW.mdx
mrousavy: Actually we are using scanFaces in other examples, no?

docs/docs/guides/FRAME_PROCESSOR_CREATE_PLUGIN_IOS.mdx
```diff
@@ -60,7 +60,7 @@ For reference see the [CLI's docs](https://github.com/mateusz1913/vision-camera-
   CMSampleBufferRef buffer = frame.buffer;
   UIImageOrientation orientation = frame.orientation;
   // code goes here
-  return @[];
+  return nil;
```
mrousavy, line 63: nice catch!

docs/docs/guides/FRAME_PROCESSOR_CREATE_PLUGIN_IOS.mdx
```diff
@@ -89,56 +89,49 @@ The Frame Processor Plugin will be exposed to JS through the `VisionCameraProxy`
 
 ![Xcode "Create Bridging Header" alert](https://docs-assets.developer.apple.com/published/7ebca7212c/2a065d1a-7e53-4907-a889-b7fa4f2206c9.png)
 
-3. Inside the newly created Bridging Header, add the following code:
```
mrousavy, line 92: Do we not need that anymore?

mrousavy (changes requested):
Hey, thanks so much for those updates, they look good overall, left some minor comments!

docs/docs/guides/FRAME_PROCESSORS_CREATE_OVERVIEW.mdx
mateusz1913: I think that those decorators should stay. If anyone will be interacting with the docs, they will either copy-paste it to their IDE and will try to run it or they will try to compare it to what they have in their IDEs. Android Studio has quite good autocompletion and if somebody will try to generate it via Android Studio and then will see that docs are not referencing `@Nullable` or `@NonNull`, they will be confused regardless of their native development experience

.github/workflows/build-ios.yml
mateusz1913: This wasn't updated when you moved the code to `package` directory, so I assume CI doesn't know that it should look for `package/`

docs/docs/guides/FRAME_PROCESSOR_CREATE_PLUGIN_IOS.mdx
mateusz1913: If you meant the point 2. above, then I think yes, saying "Yes" to "Create Bridging Header" alert should stay - it will just be empty, but if someone will decide to bridge some additional ObjC to Swift later, then they will not need to manually creating and linking that bridging header

docs/docs/guides/FRAME_PROCESSORS_CREATE_OVERVIEW.mdx
mateusz1913: In the overview page there're multiple examples, but that line is actually about passing some argument that will raise an exception. However I'm open to changing the name and value to be more accurate :D

mateusz1913: So I changed it to `codeType: string` on Java side and for raising exception there is `codeType: 1234` on JS side (so String != Int)

.github/workflows/build-ios.yml
mrousavy: Yea but I thought that CI is running in the `package/` working directory, I might be wrong though...

mrousavy:
Cool, let's merge that - awesome stuff as always, thank you! ❤️ 

Let's adjust all the examples so that we use a face detector as a plugin example though, since I will add a QR Code/Barcode reader to the codebase today in a livestream anyways. Face detector is a much better example

PR #1886, opened 2023-09-29
docs: use `Fira Mono` for codeblocks

benschlegel:
## What

Use [Fira Mono](https://fonts.google.com/specimen/Fira+Mono) for codeblocks/inline syntax highlights in docs.

image

## Tested on
- Chrome
- Firefox
- Safari


mrousavy (changes requested):
So this is the difference:

<table>
<tr>
<th>
Before
</th>
<td>

</td>
</tr>

<tr>
<th>
After

</th>

<td>

</td>
</tr>
</table>


I feel like Fira Mono is a bit too thin in this case, is there a thicker variant? If not I think we should just leave it with the default font, that loads fine and looks good with the current theme imo

benschlegel:
there is also a variant with font weight 500, otherwise using the default font is fine too

mrousavy:
danke

PR #1892, opened 2023-09-30
perf: Use `jsi::Scope` to trigger GC after Frame Processor call

mrousavy:


mrousavy:
nvm, this is not any more efficient.

PR #1908, opened 2023-10-03
fix: Fix `+load` not available in Xcode 15 error

mrousavy:
## What

cc @mateusz1913


mateusz1913:
I made similar change to CLI and it's available in v0.4.1 🚀

PR #1916, opened 2023-10-04
build: add Android MLKit Vision model downloading to expo config plugin

fobos531:
## What

https://www.react-native-vision-camera.com/docs/guides/code-scanning#setup

Adds this logic to the Expo config plugin. I know that the model is already being downloaded lazily, but why not make use of this as soon as the user installs the app :) If needed, we can make this a configurable property so that it is not downloaded via the config plugin if the user doesn't want that, but I doubt that it'll be the case

## Changes

Added AndroidManifest modification logic to the config plugin, reference: https://docs.expo.dev/config-plugins/development-and-debugging/#modify-androidmanifestxml

## Tested on

N/A


mrousavy:
Hey, nice PR!! 🤩 

Did you test this in your project?

mrousavy:
Doesn't this add it to the Manifest everytime? This looks like the plugin always runs

mrousavy:
Changed it to be configurable and added it to the docs, thanks!

PR #1918, opened 2023-10-04
doc: add full list of supported codes to example and add closing xml tag

NTKrieger:
## What
Minor documentation enhancements to the new Code Scanner guide

## Changes

This PR adds a closing xml tag to the AndroidManifest.xml meta-data element
This PR adds the full enum of readable code types to the example

## Tested on

no tests

### DIFF
--- a/docs/docs/guides/CODE_SCANNING.mdx
+++ b/docs/docs/guides/CODE_SCANNING.mdx
@@ -37,7 +37,7 @@ On Android, the [MLKit Vision Barcode Scanning](https://developers.google.com/ml
       ...
       <meta-data
           android:name="com.google.mlkit.vision.DEPENDENCIES"
-          android:value="barcode">
+          android:value="barcode" />
 </application>
 ```
 
@@ -56,7 +56,21 @@ To use a codescanner, simply create a [`CodeScanner`](/docs/api/interfaces/CodeS
 
 ```tsx
 const codeScanner = useCodeScanner({
-  codeTypes: ['qr', 'ean-13'],
+  codeTypes: [
+    'code-128',
+    'code-39',
+    'code-93',
+    'codabar',
+    'ean-13',
+    'ean-8',
+    'itf',
+    'upc-e',
+    'qr',
+    'pdf-417',
+    'aztec',
+    'data-matrix',
+    // remove unused code types
+  ],
   onCodeScanned: (codes) => {
     console.log(`Scanned ${codes.length} codes!`)
   }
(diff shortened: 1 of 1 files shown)


docs/docs/guides/CODE_SCANNING.mdx
```diff
@@ -37,7 +37,7 @@ On Android, the [MLKit Vision Barcode Scanning](https://developers.google.com/ml
       ...
       <meta-data
           android:name="com.google.mlkit.vision.DEPENDENCIES"
-          android:value="barcode">
+          android:value="barcode" />
```
mrousavy, line 40: Are you 100% sure this is correct? In the Google MLKit documentation they also don't have a closing tag:

image

https://developers.google.com/ml-kit/vision/barcode-scanning/android

docs/docs/guides/CODE_SCANNING.mdx
```diff
@@ -56,7 +56,21 @@ To use a codescanner, simply create a [`CodeScanner`](/docs/api/interfaces/CodeS
...
+    'ean-13',
+    'ean-8',
+    'itf',
+    'upc-e',
+    'qr',
+    'pdf-417',
+    'aztec',
+    'data-matrix',
+    // remove unused code types
+  ],
```
mrousavy: I think this should be reverted tbh, QR and EAN-13 give the user an idea that both codes can be scanned, no reason to hardcode everything in there. The user will get the idea that he can pass different values here. :)

mrousavy (changes requested):
Hi, thanks for your PR!

Left some comments :)

docs/docs/guides/CODE_SCANNING.mdx
hsjoberg: As described in #1922, it does not compile for me at least if it isn't a syntactically correct XML closing tag, with the error `com.android.manifmerger.ManifestMerger2$MergeFailureException: Error parsing /[...]/AndroidManifest.xml` indicating it's a parsing error.

I'm not sure* how to address this otherwise.

mrousavy: Cool - @NTKrieger can you revert the other change then we can merge this XML fix!

docs/docs/guides/CODE_SCANNING.mdx
NTKrieger: I just thought junior users might not know to go find the TS type to find the enum of accepted values.  Reverted.

docs/docs/guides/CODE_SCANNING.mdx
NTKrieger: Done

mrousavy:
Awesome, thanks!

Issue #1890, opened 2023-09-30
✨ Implement Flash for Android ($3,000)


mrousavy:
### What feature or enhancement are you suggesting?

Flash (photo/video) or Torch currently does not work on Android because of the Camera2 rewrite. It's quite tricky to add flash capture, as the exposure has to be handled manually in Camera2.

### What Platforms whould this feature/enhancement affect?

Android

LiamBateman:
I'm happy to take a look at this as this seems like a critical item for this package. Is there any starting point on this implementation?

jthure:
> I'm happy to take a look at this as this seems like a critical item for this package. Is there any starting point on this implementation?

Take a look at my comment here: https://github.com/mrousavy/react-native-vision-camera/issues/1726#issuecomment-1697352602. I would guess this is related to implementing the pre-capture sequence.

mrousavy:
Thanks to everyone who already sponsored towards this goal in the polar pool linked here - we're at $190 right now. $2,810 left to go 😅 

As a teaser I just implemented very basic flash capture which uses an OS feature - classic Android; this does not work on every phone. It's a short torch burst which does not wait for AE/AF exposure to be settled.

A proper implementation requires a fully custom precapture sequence that enables the torch, then waits for AE/AF to adjust, then lock AE/AF to that value, turn off the torch again, then capture with a single flash burst, and then turn the torch off again, and return to auto-AE/AF. This is quite complex, that's why this feature request is marked at $3,000.

mrousavy:
Update: We're far off from the $3,000, but **I decided to implement Flash anyways** - PR is up, I'm testing everything now!!! 🥳🎉🚀 

[**feat: Flash with AE Pre-capture trigger for Android (#2558)**](https://github.com/mrousavy/react-native-vision-camera/pull/2558)

(check out the code changes, there's a ton of stuff required for this lol)

mrousavy:
I just released [VisionCamera 3.9.0-beta.4](https://github.com/mrousavy/react-native-vision-camera/releases/tag/v3.9.0-beta.4), which contains flash support! (will continue testing more, then release 3.9.0 as stable)

With that I'd say this feature request is now finished - thanks to everyone who donated to the money pool!!! ❤️

PR #1932, opened 2023-10-05
feat: activate/enable the torch mode

erwanlpfr:
## What
First of all, I would like to thank you for your huge work.

On my projects, we need to use the torch mode, which makes easier users to see/display some stuff before taking the picture. Even if we aim iOS, I wanted to make sure that Android users could do it.

Maybe it is duplicated with the upcoming "flash feature", but let me know if it can be merged here!

## Changes
In this simple PR, you'll find the call with the additional boolean to properly use the prop `torchMode`.

## Tested on
Android 13, Xiaomi

### DIFF
--- a/package/android/src/main/java/com/mrousavy/camera/core/CameraSession.kt
+++ b/package/android/src/main/java/com/mrousavy/camera/core/CameraSession.kt
@@ -575,8 +575,9 @@ class CameraSession(
       val videoStabilizationMode = videoStabilizationMode
       val lowLightBoost = lowLightBoost
       val hdr = hdr
+      val enableTorch = enableTorch
 
-      val repeatingRequest = getPreviewCaptureRequest(fps, videoStabilizationMode, lowLightBoost, hdr)
+      val repeatingRequest = getPreviewCaptureRequest(fps, videoStabilizationMode, lowLightBoost, hdr, enableTorch)
       Log.d(TAG, "Setting Repeating Request..")
       session.setRepeatingRequest(repeatingRequest, null, null)
     }


mrousavy:
Hey! Oh, thanks I missed that!

Does that actually work for you though? I tested this on a Huawei P10 and it didn't enable the torch a while ago (not sure if something changed since then)

erwanlpfr:
Hello ! 

Today, I will try on an Honor 9 Lite and Huawei P40.
For now, everything'is work.

Note that I build with Android 24.

package/android/src/main/java/com/mrousavy/camera/core/CameraSession.kt
```diff
@@ -575,8 +575,9 @@ class CameraSession(
       val videoStabilizationMode = videoStabilizationMode
       val lowLightBoost = lowLightBoost
       val hdr = hdr
+      val enableTorch = enableTorch
```
erwanlpfr, line 578: Btw, is there a reason to redeclare the variables here?
I assume with the thread?
Just being curious since I never coded in kotlin!

mrousavy: Yea it had something to do with threading previously, but I think now it could be removed. I'll do that in a later PR :)

mrousavy:
> Today, I will try on an Honor 9 Lite and Huawei P40

Cool, let me know and if that works I will merge it!

erwanlpfr:
It works as expected!
I'll continue testing btw
Thank you for the merge!

mrousavy:
awesome, thanks. I'll still need to implement a precapture sequence for flash (see https://github.com/mrousavy/react-native-vision-camera/issues/1890)

PR #1948, opened 2023-10-06
flipImage was doing nothing on landscape orientated images

almost:
Added the case for ORIENTATION_NORMAL

## What

On Android when the front camera is used the code *tries* to flip the image (to match the preview). On Portrait and Landscape Right orientations this works. On Landscape Left it does not. The reason for this is that the flipImage function checks the orientation of the image and applies a different transform based on that, but it's missing the case for ORIENTATION_NORMAL which is what is set when the image was tken in Landscape Left.

This PR adds that ORIENTATION_NORMAL case.

## Changes

Add a case for ORIENTATION_NORMAL to the when block in flipImage

## Tested on

Android Samsung tablet and phone.


mrousavy:
Great, thank you!

PR #1960, opened 2023-10-10
Fix Frame Processor is not removed when Camera view unmount

r0b0t3d:
## What
This PR fix issue frame proccessor is not removed when Camera view unmount

## Changes
- Add `componentWillUnmount` method
- Check if frame processor is not null
- Unset frame processor

## Tested on
- iPhone 7, ios 15
- I don't have physical android devices

### DIFF
--- a/package/src/Camera.tsx
+++ b/package/src/Camera.tsx
@@ -434,7 +434,7 @@ export class Camera extends React.PureComponent<CameraProps> {
 
   /** @internal */
   componentWillUnmount(): void {
-    if (this.props.frameProcessor !== null) {
+    if (this.lastFrameProcessor !== null) {
       this.unsetFrameProcessor()
     }
   }
--- a/package/src/Camera.tsx
+++ b/package/src/Camera.tsx
@@ -431,6 +431,13 @@ export class Camera extends React.PureComponent<CameraProps> {
       this.lastFrameProcessor = frameProcessor
     }
   }
+
+  /** @internal */
+  componentWillUnmount(): void {
+    if (this.props.frameProcessor !== null) {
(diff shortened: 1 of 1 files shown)


package/src/Camera.tsx
```diff
@@ -431,6 +431,13 @@ export class Camera extends React.PureComponent<CameraProps> {
       this.lastFrameProcessor = frameProcessor
     }
   }
+
+  /** @internal */
+  componentWillUnmount(): void {
+    if (this.props.frameProcessor !== null) {
```
mrousavy:
```suggestion
    if (this.lastFrameProcessor !== null) {
```

mrousavy (changes requested):
Hey, thanks for the PR, this is good!

Just one minor detail, let's use `this.lastFrameProcessor` as that is the source of truth of the currently in-memory FP.

mrousavy:
Actually on second thought, do we really need this? Was there any problem before?

If the Camera unmounts on the native side, it will delete the `FrameProcessor` instance anyways. Calling unset here is probably an unnecessary call to native...

r0b0t3d:
@mrousavy I got the issue in my local development. When the camera unmount, I still got console.log from frame processor

mrousavy:
Did it crash? Or an error?

r0b0t3d:
Not crash or error. But it's not right when frame processor is still running even if no camera active

mrousavy:
How long was it running for? When the Camera fully unmounts it will stop, maybe it will run for one or two more frames because the Camera is not fully unmounted.

Please share a screen recording or something.

PR #1972, opened 2023-10-10
fix: Always call `CaptureSession` fully synchronously under Mutex

mrousavy:
## What

Fixes an issue in the Expensify app


ajitpradhan925:
How to resolve this issue?

PR #1980, opened 2023-10-11
fix: "Main Thread Checker: UI API called on a background thread: -[UI…

carlblock:
## What

When scanning codes for the first time the initial code scan takes several seconds 3-5 sec before calling onCodeScanned. 

This PR fixes the warning in Xcode and also improves the scan performance.

## Changes

This PR wraps the frame processing block in `DispatchQueue.main.async {}`

## Tested on

iPhone 12 Pro (iOS 17.0.2)

### DIFF
--- a/package/ios/CameraView+CodeScanner.swift
+++ b/package/ios/CameraView+CodeScanner.swift
@@ -18,12 +18,15 @@ extension CameraView: AVCaptureMetadataOutputObjectsDelegate {
       return
     }
 
+    // Run this on the main queue, since we access the view's layout from here to convert the coordinates
     DispatchQueue.main.async {
+      // Map codes to JS values
       let codes = metadataObjects.map { object in
           var value: String?
           if let code = object as? AVMetadataMachineReadableCodeObject {
               value = code.stringValue
           }
+
           let frame = self.previewView.layerRectConverted(fromMetadataOutputRect: object.bounds)
           
           return [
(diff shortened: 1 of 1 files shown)


mrousavy:
Oh that's why it was taking so long, good catch! Thanks for your PR, but you removed one comment here - could you add the comment back again?

Also, above the main.async stuff could you please add this comment:

```
// Run this on the main queue, since we access the view's layout from here to convert the coordinates
```

Thanks.

carlblock:
> Oh that's why it was taking so long, good catch! Thanks for your PR, but you removed one comment here - could you add the comment back again?
> 
> Also, above the main.async stuff could you please add this comment:
> 
> ```
> // Run this on the main queue, since we access the view's layout from here to convert the coordinates
> ```
> 
> Thanks.

Hey! Great, re-added the comment and your suggestion! Also, if time allows I will start investigate the code scanner performance issues on Android also.

Thanks for a great library!

mrousavy:
Awesome, thanks!

Wdym with performance issues on Android? From what I've seen, the Android implementation is really fast. There was an issue with stalling, which I fixed here afaik: https://github.com/mrousavy/react-native-vision-camera/commit/915ef331d3b3f3630f4b8d5a8cdd7f5fcf30c335

mrousavy:
Hey! I changed the setup (almost a full rewrite again lol) and now it should work differently, it uses the video size now instead of Preview size. This does not require us to switch Threads at all, making it effectively even faster!

Thanks foryour PR tho!

PR #1999, opened 2023-10-14
fix: Always set to auto-focus

mrousavy:
## What

Configures the session to use continuous auto-focus and auto-whitebalance.

### DIFF
--- a/package/ios/Core/CameraSession+Configuration.swift
+++ b/package/ios/Core/CameraSession+Configuration.swift
@@ -244,6 +244,14 @@ extension CameraSession {
         try device.setTorchModeOn(level: 1.0)
       }
     }
+    
+    // Configure Auto-Focus
+    if device.isFocusModeSupported(.continuousAutoFocus) {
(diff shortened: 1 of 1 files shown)


haresh12 (approved):
this looks good and workable solution

mrousavy:
@haresh12 you just approved a few other random PRs - did you actually test this? Is there a difference? I doubt it, I tested it and this is exactly the same result as before

haresh12:
@mrousavy  i tested this one on simulator and it looked fine and i had little bit work experience with IOS so i went through the doc also if you thing something its not working then i could check again if needed.

mrousavy:
My man, VisionCamera does not work on a simulator 🤣

haresh12:
https://accounts.lambdatest.com/login
this one i am talking please guide me if i missed something or done it on wrong way on learning face even i would love to solve issue its still exist. Sorry for the mistakes

mrousavy:
Fixed in V5

PR #2025, opened 2023-10-17
[V2] Correct Frame Processor loading between Objc/Swift

iBotPeaches:
## What

This PR fixes Frame Processors (Swift/Objc). I used the following for research:
* https://developer.apple.com/documentation/objectivec/nsobject/1418815-load?language=objc
* https://github.com/facebook/react-native/commit/ff66600224e78fec5d0e902f8a035b78ed31a961

Doing initialize appeared to be too early (or not consistent) as we might load code before reanimated is ready. When we move to "load" - that is no longer allowed on Swift due to the interop between objc. attribute constructor runs after load and seems the best place.

## Changes

* Corrects frame processor loading for objc/swift based processors.

## Tested on

* iPhone 13 Pro

## Related issues

* refs: https://github.com/mrousavy/react-native-vision-camera/issues/2017
* refs: https://github.com/mrousavy/react-native-vision-camera/issues/1543
* refs: https://github.com/mrousavy/react-native-vision-camera/issues/1959

PR #2027, opened 2023-10-17
feat: reintroduce macros for FP registration

mateusz1913:
## What

in VisionCamera v1 & v2 there were two ObjC macros that were helping in creation/registration of Frame Processors, but these were removed with v3

This PR reintroduces such macros, which will not only make FP development easier, but also it will also fix issues people had with registration of Swift Frame Processors (+load vs +initialize issues)

Docs were also updated to reflect that the macros should be used to correctly initialize and register ObjC/Swift Frame Processors

## Changes

- macros for Objc/Swift plugin registration
- updated examples
- updated docs

## Tested on

iPhone 12 Pro

## Related issues

Similar change regarding the Swift FP registration is made for V2 branch - https://github.com/mrousavy/react-native-vision-camera/pull/2025

### DIFF
--- a/package/example/android/app/src/main/java/com/mrousavy/camera/example/ExampleFrameProcessorPlugin.java
+++ b/package/example/android/app/src/main/java/com/mrousavy/camera/example/ExampleFrameProcessorPlugin.java
@@ -40,7 +40,7 @@ public class ExampleFrameProcessorPlugin extends FrameProcessorPlugin {
         return map;
     }
 
-    ExampleFrameProcessorPlugin() {
-
+    ExampleFrameProcessorPlugin(@Nullable Map<String, Object> options) {
+        Log.d("ExamplePlugin", " - options: " + options.toString());
     }
 }
--- a/package/ios/Frame Processor/FrameProcessorPlugin.h
+++ b/package/ios/Frame Processor/FrameProcessorPlugin.h
@@ -21,8 +21,39 @@
 /// VisionCamera Runtime.
 @interface FrameProcessorPlugin : NSObject
 
+/// The initializer for a Frame Processor Plugin class that takes optional object that consists
+/// options passed from JS layer
+- (instancetype _Nonnull)initWithOptions:(NSDictionary* _Nullable)options;
+
 /// The actual callback when calling this plugin. Any Frame Processing should be handled there.
 /// Make sure your code is optimized, as this is a hot path.
 - (id _Nullable)callback:(Frame* _Nonnull)frame withArguments:(NSDictionary* _Nullable)arguments;
 
 @end
+
+#define VISION_CONCAT2(A, B) A##B
+#define VISION_CONCAT(A, B) VISION_CONCAT2(A, B)
+
+#define VISION_EXPORT_FRAME_PROCESSOR(frame_processor_class, frame_processor_plugin_name)                                                  \
+  +(void)load {                                                                                                                            \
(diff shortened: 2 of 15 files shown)


package/ios/Frame Processor/FrameProcessorPlugin.h
```diff
@@ -26,3 +26,30 @@
...
+  @end                                                                                                                                     \
+                                                                                                                                           \
+  @implementation frame_processor_class (FrameProcessorPluginLoader)                                                                       \
+                                                                                                                                           \
+  __attribute__((constructor)) static void VISION_CONCAT(initialize_, frame_processor_plugin_name)(void) {                                 \
+    [FrameProcessorPluginRegistry addFrameProcessorPlugin:@ #frame_processor_plugin_name                                                   \
+                                          withInitializer:^FrameProcessorPlugin* _Nonnull(NSDictionary* _Nullable options) {               \
+                                            return [[frame_processor_class alloc] init];                                                   \
```
mrousavy: `options` need to be passed here as well!

mrousavy (changes requested):
Looks great, thanks!

I should've never removed that 😅 

We just need to pass `options` to the init func then we're good!

package/ios/Frame Processor/FrameProcessorPlugin.h
```diff
@@ -26,3 +26,30 @@
...
 @end
+
+#define VISION_CONCAT2(A, B) A##B
+#define VISION_CONCAT(A, B) VISION_CONCAT2(A, B)
+
+#define VISION_EXPORT_FRAME_PROCESSOR(frame_processor_class, frame_processor_plugin_name)                                                  \
+  +(void)load {                                                                                                                            \
+    [FrameProcessorPluginRegistry addFrameProcessorPlugin:@ #frame_processor_plugin_name                                                   \
+                                          withInitializer:^FrameProcessorPlugin*(NSDictionary * options) {                                 \
+                                            return [[frame_processor_class alloc] init];                                                   \
```
mrousavy: Here as well! `options` are not passed.

Maybe we should always have `init`/constructors with options, but we can just make them nullable?

mateusz1913: Good catch, let me fix it

mateusz1913: I added default `- initWithOptions` constructor to the `FrameProcessorPlugin` class, so users may override it to consume that `options` object. Also it seems that object wasn't marked in TypeScript, so I added it and adjusted rest of iOS/Android code to consume it properly. Let me know if it's ok

mateusz1913:
I prepared corresponding [PR for the plugin builder CLI](https://github.com/mateusz1913/vision-camera-plugin-builder/pull/57), ping me when this PR will be merged and included in next release

package/example/android/app/src/main/java/com/mrousavy/camera/example/ExampleFrameProcessorPlugin.java
```diff
@@ -40,7 +40,7 @@ public Object callback(@NotNull Frame frame, @Nullable Map<String, Object> param
         return map;
     }
 
-    ExampleFrameProcessorPlugin() {
-
+    ExampleFrameProcessorPlugin(@Nullable Map<String, Object> options) {
+        Log.d("ExamplePlugin", " - options: " + options.toString());
```
mrousavy, line 44: this will crash if it's null, but I'll fix it dw

mrousavy (approved):
Awesome stuff Mateusz!!!!! Thank you ❤️

PR #2032, opened 2023-10-18
fix: Fix delay in `startRecording()` by eagerly starting AssetWriter

mrousavy:
## What

Fixes the delay in `startRecording()` by eagerly starting the AVAssetWriter session instead of lazily starting it in the `onFrame` callback.

The difference is that by eagerly starting it we are blocking the native module camera  queue (async), whereas the lazy start would block the video streaming queue (effectively causing a delay to the user)

### DIFF
--- a/package/ios/Core/CameraSession+Video.swift
+++ b/package/ios/Core/CameraSession+Video.swift
@@ -105,7 +105,6 @@ extension CameraSession {
         onError(.capture(.createRecorderError(message: error.description)))
         return
       }
-      self.recordingSession = recordingSession
 
       // Init Video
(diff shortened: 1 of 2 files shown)


mrousavy:
@thomas-coldwell I didn't test this yet, just had this idea on the road - but do you think that makes sense? not sure how we deal with the timestamps - we start at `CACurrentMediaTime()` but then use the sample buffer's presentation timestamps...

mrousavy:
Exciting news: this is now obsolete, as VisionCamera V5 now supports this - we can pre-create a `Recorder` in JS to warm everything up before starting the recording! 🥳

PR #2037, opened 2023-10-18
[V2] Upgrade to RN72.6

iBotPeaches:
## What

This PR adds RN72 into /Example for testing confirmation of compatibility on the v2 branch as well as upgrading to the latest in reanimated v3.

## Changes

* Changes were made following upgrade guide, skipping anything was already pre-modified.
* This entails things like
  * Skipping Flipper
  * ESLint
  * Jest

## Tested on

* Galaxy A13 5G
* Pixel 7 Pro
* iPhone 13 Pro

## Related issues

Nothing related. I recently fixed frame processors on v2 on the example app here. I did this in hopes that it would fix our plugin (https://github.com/sourcetoad/vision-camera-plugin-barcode-scanner) and internal application that are still on v2.

I did this because our application, our sample plugin application and the vision sample application all had the same exact error of some form of `ReanimatedError: property: __scanQRCodes doesn't exist`

While I fixed the sample here - it did not solve the problem on our plugin or internal application. So once I realized our internal application was on RN72 and this was not, I upgraded the example to RN72 to presumably see an error. However, I did not which saddened me.

This points at an undiscovered error on my side with our plugin. So I figured I might as well upstream my work on v2 for RN72.

video attached of proof.

https://github.com/mrousavy/react-native-vision-camera/assets/611784/2a255ee3-ab1a-468c-9199-93eff018837c

PR #2053, opened 2023-10-19
[V2] Fix swift macro on Xcode 15

iBotPeaches:
## What

Xcode 15 seems to enforce the protocol segment on an interface in my best research. As this unchanged line of code in Xcode 14 works without an issue.

Unfortunately on the flip between swapping back to +load() or removing the protocol seems to not work on 15 at the benefit of working on lower Xcodes. I believe the best path forward here is to support Xcode 15 and update the docs (as included) to enforce this.

V2 is basically EOL anyway for v3 - so I think this is acceptable. This is now confirmed working on the sample application, my sample application for https://github.com/sourcetoad/vision-camera-plugin-barcode-scanner/pulls and our internal application.

I'm personally unaware myself how this works without the protocol type on the sample application here, but fails on my own application. This exceeds my knowledge but must be related to how Xcode handles a 1st party package vs a dependency.

## Changes

* Adjust swift macro for frame processor on Xcode 15

## Tested on

* iPhone 13 Pro

https://github.com/mrousavy/react-native-vision-camera/assets/611784/bd6e5a59-b86d-42c2-9bb0-73063c32c6d9

## Related issues

* Discovered during https://github.com/mrousavy/react-native-vision-camera/pull/2037 while patching internal project/plugin to RN72, Xcode 15.
* Which I'll revisit after this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ It is recommended that everybody upgrades to VisionCamera V3. If you have any is
 > [!NOTE]
 > The documentation hosted on https://react-native-vision-camera.com represents the documentation for V3. For V2, read the [docs/](docs/) folder here.
 
-VisionCamera V2 requires React Native 0.70+, iOS 11+, Android API 21+.
+VisionCamera V2 requires React Native 0.70+, iOS 11+, Android API 21+, Xcode 15+.
 
 
 ### Documentation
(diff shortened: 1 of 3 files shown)


mrousavy:
Cool, sounds good, thanks! LGTM!

PR #2067, opened 2023-10-24
feat: Pass `ReactApplicationContext` along to Plugin Initializer

mrousavy:
## What

Passes the `ReactApplicationContext` along to the Frame Processor Plugin Initializer function.

Before:

```java
FrameProcessorPluginRegistry.addFrameProcessorPlugin("example_plugin", options -> {
  return new ExampleFrameProcessorPlugin(options)
});
```

After:

```java
FrameProcessorPluginRegistry.addFrameProcessorPlugin("example_plugin", (context, options) -> {
  return new ExampleFrameProcessorPlugin(options)
});
```

This is considered a breaking change since the new parameter has to be added to all FP Plugins


mrousavy:
@xulihang is that something you need?

xulihang:
It turns out that it is not neccessary as I can use static methods to access the context and do my job.

mrousavy:
Awesome!

PR #2084, opened 2023-10-25
Add depth data support validation and explicitly embeds it in the photo

Filocava99:
This PR checks if the depth data feature is actually supported from the current camera before trying to enable it, otherwise it will crash the app. It also explicitly embeds the depth data inside the output photo; even though it's the default behaviour I think it's better to add it so that everyone is aware of it.

## Changes

Changes are highlighted in bold:

      * photoSettings.isDepthDataDeliveryEnabled = photoOutput.isDepthDataDeliveryEnabled **&& photoOutput.isDepthDataDeliverySupported**
      * **photoSettings.embedsDepthDataInPhoto = photoSettings.isDepthDataDeliveryEnabled**

## Tested on

    * iPhone 15 Pro, iOS 17

### DIFF
--- a/package/ios/Core/CameraSession+Photo.swift
+++ b/package/ios/Core/CameraSession+Photo.swift
@@ -63,7 +63,8 @@ extension CameraSession {
       let enableShutterSound = options["enableShutterSound"] as? Bool ?? true
 
       // depth data
-      photoSettings.isDepthDataDeliveryEnabled = photoOutput.isDepthDataDeliveryEnabled
+      photoSettings.isDepthDataDeliveryEnabled = photoOutput.isDepthDataDeliveryEnabled && photoOutput.isDepthDataDeliverySupported
+      photoSettings.embedsDepthDataInPhoto = photoSettings.isDepthDataDeliveryEnabled
       if #available(iOS 12.0, *) {
         photoSettings.isPortraitEffectsMatteDeliveryEnabled = photoOutput.isPortraitEffectsMatteDeliveryEnabled
       }


package/ios/Core/CameraSession+Photo.swift
```diff
@@ -63,7 +63,8 @@ extension CameraSession {
       let enableShutterSound = options["enableShutterSound"] as? Bool ?? true
 
       // depth data
-      photoSettings.isDepthDataDeliveryEnabled = photoOutput.isDepthDataDeliveryEnabled
+      photoSettings.isDepthDataDeliveryEnabled = photoOutput.isDepthDataDeliveryEnabled && photoOutput.isDepthDataDeliverySupported
```
mrousavy, line 66: Is this change really required? Isn't `isDepthDataDeliveryEnabled` only `true` when `isDepthDataDeliverySupported` is also `true`? I can't see a scenario where it is enabled but not supported 😅 

Maybe the method throws an error if you try to set any value to `photoSettings.isDepthDataDeliveryEnabled`, even if it's `false`? In that case we **MIGHT** need to wrap it under an if instead:

```swift
if photoOutput.isDepthDataDeliverySupported {
  photoSettings.isDepthDataDeliveryEnabled = photoOutput.isDepthDataDeliveryEnabled
}
```

package/ios/Core/CameraSession+Photo.swift
```diff
@@ -63,7 +63,8 @@ extension CameraSession {
...
+      photoSettings.embedsDepthDataInPhoto = photoSettings.isDepthDataDeliveryEnabled
```
mrousavy, line 67: Does that really change something? As per Apple Developer Documentation, the default value is already `true`: 

image

Did you test this? Does this make a difference?

mrousavy (changes requested):
Hey - thanks for your PR! Much appreciated.

Left a few small comments :)

package/ios/Core/CameraSession+Photo.swift
Filocava99: Yes I know it's the default behavior but to know that you have to look it up in the iOS documentation, which is a setback given how rnvc wants to simplify the use of those features. By making it explicit, you can find out why the depth data are being embedded in the photo by simply debugging the code.

package/ios/Core/CameraSession+Photo.swift
Filocava99: Apparently yes, you are right. It wasn't mentioned in the isDepthDataDeliveryEnabled documentation page, but it was mentioned at the bottom of the isDepthDataDeliverySupported. I guess you can scrap the PR then

mrousavy:
Thanks for your replies! I think this does not change any behaviour in code at all, so I'm closing this

PR #2194, opened 2023-11-21
feat: Make `startRecording()` awaitable :tada:

### LINKED ISSUE #742: ✨ startRecording callback/promise

Would be lovely to have a callback or / be able to use `await startRecording()` in order to make sure that recording indeed did start.

https://github.com/mrousavy/react-native-vision-camera/blob/main/example/src/views/CaptureButton.tsx#L119

iOS, Android

### Alternatives/Workarounds

No workaround currently

### Additional information

- [X] I searched for [similar feature requests in this repository](https://github.com/mrousavy/react-native-vision-camera/issues?q=is%3Aopen+is%3Aissue+label%3A%22✨+enhancement%22) and found none.

### LINKED ISSUE #1016: ❓ How can I know the exact timing of video recording has been started?
### Question

Hi, thank you for this awesome library!

I'm developing an app that supports video recording with a timer and trying to migrate from react-native-camera.
My current implementation is like this:
```
const record = (duration) => {
  camera.startRecording();
  setTimeout(() => {
    camera.stopRecording();
  }, 5000);
}
```

In this way, the actual duration of a recorded video would fluctuate from 3 to 5 seconds.

To set a timer accurately, I need to start a timer when the actual recording happens, not when I call `camera.startRecording()`.

I looked at the implementation bit, but couldn't find an easy way to know the timing. I was imagining like
```
startRecording({
  onRecordingStart: () => { /* start a timer here */ };
})
```

I would like to know if I'm doing something wrong or if there is a technical limitation on this.

I also found a similar issue: https://github.com/mrousavy/react-native-vision-camera/issues/742
The code mentioned in the issue gives me an impression that it's not supported yet: https://github.com/mrousavy/react-native-vision-camera/blob/main/example/src/views/CaptureButton.tsx#L119

Thank you!

### What I tried

Use setTimeout after calling startRecording()

### VisionCamera Version

2.13.0

### Additional information

- [ ] I am using Expo

### LINKED ISSUE #1319: ✨ Make startRecording() async

Hey Marc, this package is amazing and I appreciate all the time and effort you've put into it! 

I noticed this in your ToDo file:

* [ ] Make `startRecording()` async. Due to NativeModules limitations, we can only have either one callback or one promise in a native function. For `startRecording()` we need both, since you probably also want to catch any errors that occurred during a `startRecording()` call (or wait until the recording has actually started, since this can also take some time)

Any thoughts on when this might be completed? I'm trying to get an accurate timestamp from when video recording actually starts. Without it being async, the timestamps are off by about 1.3-1.6 seconds in iOS. Android seems to be off by 0.3-0.4 seconds.

I tested iOS on an iPhone 6 and Android on a Galaxy S10 5G

Thanks again,

Eli

iOS, Android

### Alternatives/Workarounds

I'm currently testing on a couple of devices to see if I can find a standard offset. Considering I'm still fact-finding I don't have any exact answers. Though, I imagine this approach will be fruitless. Hard-coding offsets doesn't seam like a viable solution, but I'm going to try it anyway. 

### Additional information

- [x] I agree to follow this project's [Code of Conduct](https://github.com/mrousavy/react-native-vision-camera/blob/main/CODE_OF_CONDUCT.md)
- [X] I searched for [similar feature requests in this repository](https://github.com/mrousavy/react-native-vision-camera/issues?q=is%3Aopen+is%3Aissue+label%3A%22✨+enhancement%22) and found none.

mrousavy:
## What

Finally - `startRecording()` can now be awaited!!!! I'm using a little workaround that @cipolleschi from Facebook showed me - we're just using two Callbacks now.

My only concern is that after calling the first callback, the second callback might get garbage-collected. I hope this is not the case, I need to test this further.

## Breaking Change

Any errors that occur while trying to start the recording will not be delegated through the `VideoRecordOptions.onError` callback, but will instead cause the returned promise to reject.

So callers will now need to await `startRecording()`.

For your codebase, this now means:

```diff
-camera.current.startRecording({
+await camera.current.startRecording({
   // ... other props
   onError: (error) => {
     // ... your onError handling
   }
 })
```

## Related issues

- Fixes https://github.com/mrousavy/react-native-vision-camera/issues/1319
- Fixes https://github.com/mrousavy/react-native-vision-camera/issues/742
- Fixes https://github.com/mrousavy/react-native-vision-camera/issues/1016

### DIFF
--- a/docs/docs/guides/RECORDING_VIDEOS.mdx
+++ b/docs/docs/guides/RECORDING_VIDEOS.mdx
@@ -81,7 +81,7 @@ await camera.current.resumeRecording()
 The [`startRecording(...)`](/docs/api/classes/Camera#startrecording) function can be configured to enable the [flash](/docs/api/interfaces/RecordVideoOptions#flash) while recording, which natively just enables the [`torch`](/docs/api/interfaces/CameraProps#torch) under the hood:
 
 ```ts
-camera.current.startRecording({
+await camera.current.startRecording({
   flash: 'on',
   ...
 })
@@ -97,7 +97,7 @@ VisionCamera also supports H.265 ([HEVC](https://en.wikipedia.org/wiki/High_Effi
 If you can handle H.265 on your backend, configure the video recorder to encode in H.265:
 
 ```ts
-camera.current.startRecording({
+await camera.current.startRecording({
   ...props,
   videoCodec: 'h265'
 })
@@ -110,7 +110,7 @@ Videos are recorded with a target bit-rate, which the encoder aims to match as c
 To simply record videos with higher quality, use a [`videoBitRate`](/docs/api/interfaces/RecordVideoOptions#videobitrate) of `'high'`, which effectively increases the bit-rate by 20%:
 
 ```ts
-camera.current.startRecording({
+await camera.current.startRecording({
   ...props,
   videoBitRate: 'high'
 })
@@ -119,7 +119,7 @@ camera.current.startRecording({
 To use a lower bit-rate for lower quality and lower file-size, use a [`videoBitRate`](/docs/api/interfaces/RecordVideoOptions#videobitrate) of `'low'`, which effectively decreases the bit-rate by 20%:
 
 ```ts
-camera.current.startRecording({
+await camera.current.startRecording({
   ...props,
   videoBitRate: 'low'
 })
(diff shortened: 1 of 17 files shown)


mrousavy:
Okay I got it working on iOS! 🎉 

On Android it still doesn't work though:

mrousavy:
Okay update; it will not work like this. At least not with the Bridge, not sure if TurboModules will fix this.

The problem is this code:

https://github.com/facebook/react-native/blob/ce0dabddf369587311198290b5bfc8e058a39a28/packages/react-native/Libraries/BatchedBridge/NativeModules.js#L98-L162

It does not work when multiple functions are passed, as there can only always be one success and one error function.

mrousavy:
Exciting news: this is now obsolete, as VisionCamera V5 now supports this, thanks to the Nitro rewrite! 🥳

PR #2200, opened 2023-11-22
perf: Use the truly native pixel format on iOS if available

mrousavy:
## What

According to [the `videoSettings` documentation](https://developer.apple.com/documentation/avfoundation/avcapturevideodataoutput/1389945-videosettings#discussion), the truly native format is used when `videoSettings` is set to `[:]` - so this adds a path where this can happen (hdr is false, buffer compression is not enabled or available, and pixelFormat is `native`)

Not sure if this makes any difference, I think I already pick the format that is truly native for that format.

Before merging this PR, let's confirm that the pixelFormat is now different when calling `CMSampleBufferGetFormatDescription(frame.buffer)` in a FP Plugin.


mrousavy:
V5 now supports this

PR #2204, opened 2023-11-22
fix: barcode scanner crash

necmettindev:
## What

i've made changes that I believe fix the issue of crashes after barcode scans on android. i tested it with my device and didn't encounter any problems with this version.

### DIFF
--- a/package/android/src/main/java/com/mrousavy/camera/core/CodeScannerPipeline.kt
+++ b/package/android/src/main/java/com/mrousavy/camera/core/CodeScannerPipeline.kt
@@ -59,10 +59,18 @@ class CodeScannerPipeline(
             callback.onCodeScanned(barcodes, CodeScannerFrame(inputImage.width, inputImage.height))
           }
         }
+        .addOnCanceledListener {
+          image.close()
+          isBusy = false
+        }
         .addOnFailureListener { error ->
           image.close()
           isBusy = false
           callback.onError(error)
+        }
+        .addOnCompleteListener {
+          image.close()
+          isBusy = false
         }
     }, CameraQueues.videoQueue.handler)
   }


necmettindev:
@mrousavy is there any problem with merging?

mrousavy:
Hey, thanks for your PR!

No, not yet - I'd have to test this to confirm this works.
It would help if other people that have this bug could test if this fixes it for them!

mrousavy:
Can you link those issues that will be fixed by this PR? And maybe also ping some people from those issues to test this PR. Thanks!

only1chi:
> Hey, thanks for your PR!
> 
> No, not yet - I'd have to test this to confirm this works. It would help if other people that have this bug could test if this fixes it for them!

@necmettindev I haven't been able to get my app to crash with barcode scanning after applying this change. Initially, my app would crash every once in a while after scanning a QR code.

MehPhisto:
This is not solving the issue for ean-13

only1chi:
> This is not solving the issue for ean-13

You are right! My testing was limited.

MehPhisto:
Downgrading to v3.5.1 fixes the issue for me

lBroth:
> This is not solving the issue for ean-13

Same here

only1chi:
@MehPhisto @lBroth 

Have you tried this [PR](https://github.com/mrousavy/react-native-vision-camera/pull/2213)?

lBroth:
> @MehPhisto @lBroth 
> 
> 
> 
> Have you tried this [PR](https://github.com/mrousavy/react-native-vision-camera/pull/2213)?

Not yet. I hope I can try it tonight

vonovak (commented):
to be sure, I believe it'd be better to call `image.close()` only once on each image. That means it should be called in `onCancel` and `onComplete`. The way it's done at the time of this comment, the code calls it several times.

mrousavy:
Hey - I already added this to the codebase now by only calling `close` once; in the completion callback. 

https://github.com/mrousavy/react-native-vision-camera/blob/2fb56e580253d240a0d7bf857c353f8bf96870fd/package/android/src/main/java/com/mrousavy/camera/core/CodeScannerPipeline.kt#L64

Thanks for your PR though!

PR #2229, opened 2023-11-28
feat: Stereo Audio

mrousavy:
## What

Adds the option to record stereo audio when recording videos.

Stereo audio has left and right channels, and uses a stereo polar pattern. Left and right will depend on the input orientation.


mrousavy:
Hmm maybe in the future.

PR #2251, opened 2023-12-04
Feat/add prefab support

rodgomesc:
## What
add prefabs support


mrousavy:
Heyo - thanks for your PR!

I don't think we need that extra task/step in the `build.gradle` though, prefabs publishing _should_ already do that. Can you try and see if it works without that extra task?

rodgomesc:
hmm, without it i can't find vision camera headers. maybe i'm missing something ? how prefab look for the headers automatically? i think there's no way to know that's in `./src/main/cpp` unless it does some dark magic with `Cmakelists.txt` to detect include directories

image

```diff
-    prefab {
-        VisionCamera {
-            headers "${project.buildDir}/headers/visioncamera/"
-        }
-    }
```
```diff
- task prepareHeaders(type: Copy) {
-   from fileTree('./src/main/cpp').filter { it.isFile() }
-   into "${project.buildDir}/headers/visioncamera/react-native-vision-camera/"
-   includeEmptyDirs = false
- }
-
- preBuild.dependsOn(prepareHeaders)
```

mrousavy:
Awesome, thanks for testing - LGTM man!!

PR #2258, opened 2023-12-06
feat: provide permission status in hooks

vonovak:
## What

Motivation: let's say I'm using the `useCameraPermission` hook on a screen along with the CameraView. (sure, I could call the hook even sooner to learn about the permission status, but let's but that aside). The user has already given my app the permission to render Camera View. Initially, however, `hasPermission` is going to be false, even though the permission is granted (and after `getCameraPermissionStatus` resolves, it'll switch to true). This PR adds extra `status` field that can be used to distinguish this case.

let me know if there's need to add docs for this, can do 👍 

Thanks for reviewing

## Changes

* add a `status` field to `useCameraPermission` and `useMicrophonePermission`

## Tested on

iphone 12 device
android 10 device (huawei)

### DIFF
--- a/package/src/hooks/useCameraPermission.ts
+++ b/package/src/hooks/useCameraPermission.ts
@@ -1,7 +1,13 @@
 import { useCallback, useEffect, useState } from 'react'
-import { Camera } from '../Camera'
+import { Camera, CameraPermissionStatus } from '../Camera'
 
 interface PermissionState {
+  /**
+   * The current permission status, initially `not-determined`.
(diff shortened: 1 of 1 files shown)


mrousavy:
Hey man! Thank you for your PR, that's great!

I have a different idea though; I'll expose the getters as synchronous - that way we can even avoid an additional render and permission status is always available :)

mrousavy:
Or do you need to know whether the status is not determined or not?

vonovak:
yep, that sounds good! :)

PR #2267, opened 2023-12-08
possible fix for a scenario that image proxy is not closed for code scanner correctly in some cases

rodgomesc:


mrousavy:
I wonder in which cases neither of the callbacks is called though? It's either the `onSuccess` or the `onFailure` listener - how could there be a third state? lol

mrousavy:
lol ur right 
Screenshot 2023-12-12 at 11 18 06

PR #2270, opened 2023-12-08
[v2] fix: set `NDEBUG=1` preprocessor flag in release mode

### LINKED ISSUE #2265: Xcode Archive Error: Undefined symbols for WorkletRuntime::runGuarded 🔧
### How were you trying to build the app?

Hi

I've run into an Xcode archive issue when using `react-native-vision-camera` v2.16.5 together with the newest version of `react-native-reanimated` (3.6.1). The issue arises because `react-native-reanimated` introduced some branching behaviour in v3.6.0 which relies on the `PRODUCTION` flag used when installing pods. This means that you have to run `PRODUCTION=1 pod install` in order to be able to create iOS release builds. In v3.5.4 of `react-native-reanimated` it was enough to just run `pod install` before creating iOS release builds. This is an issue for us because we have pods committed to the repo so we cannot rely on two different sets on pods based on whether you need to create debug or release builds.

I've filed an issue with `react-native-reanimated` (https://github.com/software-mansion/react-native-reanimated/issues/5465) but thought this might be relevant for users of `react-native-vision-camera` since that combination was what broke it for me. And also I hope that someone in here can come up with a solution to make it work. I don't know if the issue exists with `react-native-vision-camera` v3, but we cannot use that yet because it doesn't work with JSC (and we cannot yet update to Hermes due to its excluded JS features). 

The issue can be reproduced with the `react-native-vision-camera` example app when installing `react-native-reanimated` v3.6.1 and then making a call to `WorkletRuntime::runGuarded` e.g. like this:

```
#import "AppDelegate.h"

#import <React/RCTBundleURLProvider.h>
#import <RNReanimated/WorkletRuntime.h>

@implementation AppDelegate

std::shared_ptr<reanimated::WorkletRuntime> workletRuntime;

- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
{
  self.moduleName = @"VisionCameraExample";
  // You can add your custom initial props in the dictionary below.
  // They will be passed down to the ViewController used by React Native.
  self.initialProps = @{};

  jsi::Runtime& runtime = *(jsi::Runtime*)"@123";
  jsi::Value value = jsi::Value(42);
  auto rt = reanimated::extractWorkletRuntime(runtime, value);
  jsi::Runtime &rtt = rt->getJSIRuntime();
  auto worklet = reanimated::extractShareableOrThrow<reanimated::ShareableWorklet>(rtt, value);
  workletRuntime = reanimated::extractWorkletRuntime(runtime, value.asObject(runtime));
  std::weak_ptr<reanimated::WorkletRuntime> weakWorkletRuntime = workletRuntime;
  auto workletRuntime = weakWorkletRuntime.lock();
  workletRuntime->runGuarded(worklet);

  return [super application:application didFinishLaunchingWithOptions:launchOptions];
}

- (NSURL *)sourceURLForBridge:(RCTBridge *)bridge
{
#if DEBUG
  return [[RCTBundleURLProvider sharedSettings] jsBundleURLForBundleRoot:@"index"];
#else
  return [[NSBundle mainBundle] URLForResource:@"main" withExtension:@"jsbundle"];
#endif
}

/// This method controls whether the `concurrentRoot`feature of React18 is turned on or off.
///
/// @see: https://reactjs.org/blog/2022/03/29/react-v18.html
/// @note: This requires to be rendering on Fabric (i.e. on the New Architecture).
/// @return: `true` if the `concurrentRoot` feature is enabled. Otherwise, it returns `false`.
- (BOOL)concurrentRootEnabled
{
  return true;
}

@end

and the finally archiving the app via Xcode (requires signing to be set up for the target).

### Full build logs

```tsx
ld: Undefined symbols:
  reanimated::getCallGuard(facebook::jsi::Runtime&), referenced from:
      void reanimated::runOnRuntimeGuarded<>(facebook::jsi::Runtime&, facebook::jsi::Value const&) in AppDelegate.o
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

### Project dependencies

```json
"dependencies": {
  "react-native": "0.72.6",
  "react-native-vision-camera": "2.16.5",
  "react-native-reanimated": "3.6.1",
}
```

### VisionCamera Version

2.16.5

### Target platforms

iOS

### Operating system

MacOS

### Can you build the VisionCamera Example app?

Yes, I can successfully build the Example app here

### Additional information

- [ ] I am using Expo
- [X] I have enabled Frame Processors (react-native-worklets-core)

tomekzaw:
## What

This PR fixes the following error reported in #2265 that occurs when trying to build a RN app with Reanimated 3.6.x and RNVisionCamera 2.16.5 with Frame Processors enabled in release mode on iOS on Paper:

```
Undefined symbols for architecture x86_64:
  "reanimated::getCallGuard(facebook::jsi::Runtime&)", referenced from:
      void reanimated::WorkletRuntime::runGuarded<facebook::jsi::Object&>(std::__1::shared_ptr<reanimated::ShareableWorklet> const&, facebook::jsi::Object&) const in libVisionCamera.a(FrameProcessorRuntimeManager.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

The root cause of this issue is that Reanimated 3.6.0 uses a well-known `NDEBUG` flag but unfortunately React Native does not set it on iOS on Paper (it is set correctly on Fabric, though). The workaround is to set environmental variable `PRODUCTION=1` when running `pod install` (so that `RNReanimated.podspec` knows that it should inject `NDEBUG` flag) but this flag is not well-documented so many people don't know about it.

## Changes

* Added `NDEBUG=1` flag to `GCC_PREPROCESSOR_DEFINITIONS` for `Release` variant in `VisionCamera.podspec`

## Tested on

I reproduced the error on a fresh RN 0.72.7 app with Reanimated 3.6.0 and RNVisionCamera 2.16.5 on iOS in release mode. Then I made the change in VisionCamera's podspec and the error was gone.

## Related issues

Fixes #2265.

See https://github.com/software-mansion/react-native-reanimated/pull/5366 for full explanation.

### DIFF
--- a/VisionCamera.podspec
+++ b/VisionCamera.podspec
@@ -28,7 +28,8 @@ Pod::Spec.new do |s|
 
   s.pod_target_xcconfig = {
     "USE_HEADERMAP" => "YES",
-    "HEADER_SEARCH_PATHS" => "\"$(PODS_TARGET_SRCROOT)/ReactCommon\" \"$(PODS_TARGET_SRCROOT)\" \"$(PODS_ROOT)/RCT-Folly\" \"$(PODS_ROOT)/boost\" \"$(PODS_ROOT)/boost-for-react-native\" \"$(PODS_ROOT)/DoubleConversion\" \"$(PODS_ROOT)/Headers/Private/React-Core\" "
+    "HEADER_SEARCH_PATHS" => "\"$(PODS_TARGET_SRCROOT)/ReactCommon\" \"$(PODS_TARGET_SRCROOT)\" \"$(PODS_ROOT)/RCT-Folly\" \"$(PODS_ROOT)/boost\" \"$(PODS_ROOT)/boost-for-react-native\" \"$(PODS_ROOT)/DoubleConversion\" \"$(PODS_ROOT)/Headers/Private/React-Core\" ",
+    "GCC_PREPROCESSOR_DEFINITIONS[config=Release]" => "$(inherited) NDEBUG=1"
   }
   s.compiler_flags = folly_compiler_flags + ' ' + boost_compiler_flags
   s.xcconfig = {


mrousavy:
You're the 🐐! LGTM thanks Tomasz! ❤️

tomekzaw:
@mrousavy Thanks for merging this PR! Would it be possible to publish a new v2 release with this fix?

mrousavy:
Sorry, went for lunch & forgot haha - just [released 2.16.5](https://github.com/mrousavy/react-native-vision-camera/releases/tag/v2.16.6)! :)

thanks again!

tomekzaw:
Thank you on behalf of the entire community!

PR #2273, opened 2023-12-09
feat: Add `onStarted` and `onStopped` events

mrousavy:
## What

Adds two events:

- `onStarted`: Called when the Camera started streaming frames
- `onStopped`: Called when the Camera stops streaming frames

### DIFF
--- a/package/android/src/main/java/com/mrousavy/camera/CameraView+Events.kt
+++ b/package/android/src/main/java/com/mrousavy/camera/CameraView+Events.kt
@@ -19,6 +19,20 @@ fun CameraView.invokeOnInitialized() {
   reactContext.getJSModule(RCTEventEmitter::class.java).receiveEvent(id, "cameraInitialized", null)
 }
 
+fun CameraView.invokeOnStarted() {
+  Log.i(CameraView.TAG, "invokeOnStarted()")
+
+  val reactContext = context as ReactContext
+  reactContext.getJSModule(RCTEventEmitter::class.java).receiveEvent(id, "cameraStarted", null)
+}
+
+fun CameraView.invokeOnStopped() {
+  Log.i(CameraView.TAG, "invokeOnStopped()")
+
+  val reactContext = context as ReactContext
+  reactContext.getJSModule(RCTEventEmitter::class.java).receiveEvent(id, "cameraStopped", null)
+}
+
 fun CameraView.invokeOnError(error: Throwable) {
   Log.e(CameraView.TAG, "invokeOnError(...):")
   error.printStackTrace()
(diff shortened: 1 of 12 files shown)


package/android/src/main/java/com/mrousavy/camera/CameraView+Events.kt
```diff
@@ -19,6 +19,20 @@ fun CameraView.invokeOnInitialized() {
   reactContext.getJSModule(RCTEventEmitter::class.java).receiveEvent(id, "cameraInitialized", null)
 }
 
+fun CameraView.invokeOnStarted() {
+  Log.i(CameraView.TAG, "invokeOnStarted()")
+
+  val reactContext = context as ReactContext
+  reactContext.getJSModule(RCTEventEmitter::class.java).receiveEvent(id, "cameraStarted", null)
```
angelos3lex, line 26: Hey @mrousavy Maybe we should probably add the native timestamp in this event? 
As otherwise, JS thread may delay a lot to consume these events, so if someone wants absolute precision he won't be able to do so in JS thread, making this feature less useful?

mrousavy: Hey - why would you need the timestamp, what would that actually change?

angelos3lex: I can think of a feature where you want to synchronize video frames with another source, i.e bluetooth data from a device, in matter of ms delay.
To do so, one may want to have absolute control of when the video started/stopped recording frames, and match this with the device's stream input.
Without native timestamp on these events, we can only mark the timestamp on JS thread when this event is captured. Though, in case the JS thread is blocked and this event is processed later in the queue, we can end with completely false timestamp

mrousavy: Tbh I don't agree here - this should definitely be done with a Frame Processor then, where you get a full stream of frames with their exactly accurate timestamps, and they are definitely in time as it is synchronous and blocking.

xulihang:
It is possible to return the frame's size in the `onStarted` event so that I don't have to get it in the frame processor?

mrousavy:
No, because there is no "frame" if there is no VideoPipeline for example. `onStarted` also fires if `video={false}`, in which case I do not have access to the frame.

You could get the size from your `format` (videoWidth x videoHeight).

If you are not using `format`, I could add a `onFormatChanged` callback which gives you the format that was just configured which is just the default format?

xulihang:
If I specify the video format to 1080P, the frame's size is 1920x1080 on Android and 1080x1920 on iOS. If this is a fixed behavior, I can handle this difference based on platforms and don't need to get the frame size in an event.

But it would be great to have an event to return the actual frame size after the camera starts.

Issue #2257, opened 2023-12-06
🐛 Camera has been disconnected


gsagergithub:
### What's happening?

Hi everyone. I find myself surrounded by the darkest darkness trying to use the rear camera of my Motorola Moto g6 plus, since I'm getting multiple errors. When I navigate to the screen containing the Camera component this one shows up:
`Camera.onError(session/camera-has-been-disconnected): [session/camera-has-been-disconnected] The given Camera device (id: 0) has been disconnected! Error: UNKNOWN_CAMERA_DEVICE_ERROR [session/camera-has-been-disconnected: [session/camera-has-been-disconnected] The given Camera device (id: 0) has been disconnected! Error: UNKNOWN_CAMERA_DEVICE_ERROR]`
And then, navigating back and forth again I'm getting these other two
`Camera.onError(unknown/unknown): [unknown/unknown] CameraDevice was already closed [unknown/unknown: [unknown/unknown] CameraDevice was already closed]`

`Camera.onError(session/camera-not-ready): [session/camera-not-ready] The Camera is not ready yet! Wait for the onInitialized() callback! [session/camera-not-ready: [session/camera-not-ready] The Camera is not ready yet! Wait for the onInitialized() callback!]`

I don't want to leave without telling you that I've tested the native camera application of the phone  and it's working properly,  and that I have at least three other devices where the application that integrates react native vision camera, runs well. One of them is a Motorola Moto e6 Plus. 

I would appreciate any kind of help/guidance to solve this issue. Thank you!

### Reproduceable Code

```tsx
const device = useCameraDevice('back');

<View>
  {
    isActive &&
    <Camera
      style={styles.container}
      device={device}
      isActive={isActive}
      codeScanner={codeScanner}
      enableZoomGesture={true}
    />
  }
</View>
```

### Relevant log output

```shell
2023-12-06 15:16:08.545 10591-10591 CameraView              pid-10591                            I  Updating CameraSession...
2023-12-06 15:16:08.547 10591-11047 CameraSession           pid-10591                            I  Updating CameraSession Configuration...
2023-12-06 15:16:08.548 10591-11047 CameraSession           pid-10591                            I  Successfully updated CameraSession Configuration! isActive: false
2023-12-06 15:16:08.559 10591-10591 CameraSession           pid-10591                            I  PreviewView Surface destroyed! Surface(name=null)/@0xff010d1
[...]

PR #2279, opened 2023-12-12
Re-open camera device, if it has been closed by Android

jthure:
## What
This PR adds a state variable to the CameraSession which keeps track of wether the camera device is opened or if has been closed/disconnected. It fixes the problem where if you minimize the camera app for a while, the camera gets disconnected and when entering the camera app again, it wouldn't be re-opened.

## Changes
- Adds a member `isOpen` to `CameraSession`
- `isOpen` is set to `true` during `CameraManager.open` and to false in the `onDisconnected` callback

## Tested on
Tested in the example app on a Samsung Galaxy S10

## Related issues
Possible related issues:
#2257, #2223

### DIFF
--- a/package/android/src/main/java/com/mrousavy/camera/core/CameraSession.kt
+++ b/package/android/src/main/java/com/mrousavy/camera/core/CameraSession.kt
@@ -78,6 +78,7 @@ class CameraSession(private val context: Context, private val cameraManager: Cam
   private val photoOutputSynchronizer = PhotoOutputSynchronizer()
   private val mutex = Mutex()
   private var isRunning = false
+  private var isOpen = false
 
   override val coroutineContext: CoroutineContext
     get() = CameraQueues.cameraQueue.coroutineDispatcher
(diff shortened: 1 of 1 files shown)


mrousavy:
Hey, thanks for your PR!

To be honest I'm a bit hesitant with introducing an additional state (here, the boolean `isOpen`) to the codebase.

I want it to be as atomic as possible, and I think too much state can go out of sync, which leads to messy and unmaintained projects.

I am thinking if we can use the `isRunning` prop here? Or wrap `CameraDevice` with isolated state and auto-reopen it if needed?

Let's leave this PR open for now, I'll take a look at Android this & next week to make sure we're as atomic as on iOS, then everything should be stable!

jthure:
> Hey, thanks for your PR!
> 
> To be honest I'm a bit hesitant with introducing an additional state (here, the boolean `isOpen`) to the codebase.
> 
> I want it to be as atomic as possible, and I think too much state can go out of sync, which leads to messy and unmaintained projects.
> 
> I am thinking if we can use the `isRunning` prop here? Or wrap `CameraDevice` with isolated state and auto-reopen it if needed?
> 
> Let's leave this PR open for now, I'll take a look at Android this & next week to make sure we're as atomic as on iOS, then everything should be stable!

I totally agree, it's not pretty. I'll try to take another look at it with your suggestions in mind :)

jthure:
@mrousavy I think you fixed this issue in another PR right? I'll close this one

mrousavy:
Hey yea thanks I attempted to fix this in another PR, and I think I did, but I will rewrite some of that code now anyways. Thanks again for your PR and sorry it didn't land in main! :)

PR #2305, opened 2023-12-19
fix: Allow 60+ FPS on Samsung phones

mrousavy:
## What

Newer Samsung phones such as the Samsung S21 actually support third party 60 FPS or higher.

This PR removes the lock we had that disabled 60 FPS on Samsung devices, so now we can do more than that.

If you are trying to use 60 FPS on an older Samsung where it does not work, you might see a blackscreen. This is Samsung's fault, because they report 60 FPS as supported but then simply don't open the Camera when you try to use 60 FPS - so in that case please report a bug with the exact device model so we can disable 60 FPS for those specific devices. 

## Tested on

- Samsung S21


dakiquang:
Has this been fixed yet? I have seen that this code is no longer available on version 3.7.0

mrousavy:
Yep already merged

PR #2355, opened 2024-01-08
feat: Always download `CodeScanner` model instead of relying on Google Play Services

mrousavy:
## What

Instead of using RegEx replaces, this change adds a new prop to the `build.gradle` setup of VisionCamera: `VisionCamera_enableCodeScanner`.

`VisionCamera_enableCodeScanner` can be configured in your app's `gradle.properties` file: https://github.com/mrousavy/react-native-vision-camera/blob/fdbc2dbe7888b59699753e7645068f2f23652019/package/example/android/gradle.properties#L44-L45

- When `VisionCamera_enableCodeScanner ` is set to `false` (the default), the Google Play Services MLKit library will be used which only includes the code for Barcode detection but not the 2.4 MB model. On phones that support Google Play Services, the model can be dynamically downloaded once you use a CodeScanner (`useCodeScanner`).
- When `VisionCamera_enableCodeScanner` is set to `true`, VisionCamera will include the full MLKit dependency instead of the dynamic google play services one, and will bundle the 2.4 MB model into your app. This will work on every Android phone then, even without Google Play Services.

https://github.com/mrousavy/react-native-vision-camera/blob/fdbc2dbe7888b59699753e7645068f2f23652019/package/android/build.gradle#L168-L175

### DIFF
--- a/docs/docs/guides/CODE_SCANNING.mdx
+++ b/docs/docs/guides/CODE_SCANNING.mdx
@@ -72,7 +72,7 @@ Add the `enableCodeScanner` flag to your Expo config (`app.json`, `app.config.js
 
 ## Usage
 
-To use a codescanner, simply create a [`CodeScanner`](/docs/api/interfaces/CodeScanner) and pass it to the `<Camera>`:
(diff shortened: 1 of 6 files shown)


kyjus25:
v3.7.0 works great! Thank you again!

mrousavy:
Awesome thanks for confirming!

PR #2391, opened 2024-01-15
feat: Add `minFocusDistance` prop to `CameraDevice`

### LINKED ISSUE #2246: iPhone 15 pro can't focus close up documents when using the scan utility.🐛
### What's happening?

Since purchased the new iPhone 15 pro, it seems that the iOS app can't automatically switch to the macro lens (the ultra wide lens), so it can't focus small documents (like tickets or barcodes).

This same error occurs in other third party apps.

Apple says is the "developer fault" because they are making the wrong call to the camera function, and so it only uses the main camera, that in the 14 and 15 pro models can't close focus.

This topic has been addressed in a session at WWDC 2021: https://developer.apple.com/videos/play/wwdc2021/10047/?time=133

Essentially the solution is to use the minimumFocusDistance property of the capture device to derive a videoZoomFactor to make sure the users hold the phone at a larger distance than the minimumFocusDistance. This is also illustrated in this sample code: [https://developer.apple.com/documentation/avfoundation/capture_setup/avcambarcode_detecting_barcodes...](https://developer.apple.com/documentation/avfoundation/capture_setup/avcambarcode_detecting_barcodes_and_faces)

### Reproduceable Code

```tsx
const device = useCameraDevice('back');
  const codeScanner = useCodeScanner({
    codeTypes: ['qr', 'ean-13'],
    onCodeScanned: onCodeScanned,
  });

  const _renderCamera = () => {
    if (device == null || !hasPermission || !isFocused) {
      return null;
    }

    return (
      <Camera
        style={styles.qrCamera}
        device={device}
        isActive={true}
        codeScanner={codeScanner}
        enableZoomGesture={true}
        torch={torchStatus}
      />
    );
  };
```

### Relevant log output

```shell
None
```

### Camera Device

```json
None
```

### Device

iPhone15 pro

### VisionCamera Version

3.6.12

### Can you reproduce this issue in the VisionCamera Example app?

I didn't try (⚠️ your issue might get ignored & closed if you don't try this)

### Additional information

- [ ] I am using Expo
- [ ] I have enabled Frame Processors (react-native-worklets-core)

mrousavy:
## What

Adds the `minFocusDistance` prop to the `CameraDevice` to expose how close the device can focus to.

## Related issues

- Resolves https://github.com/mrousavy/react-native-vision-camera/issues/2246

### DIFF
--- a/docs/docs/guides/ERRORS.mdx
+++ b/docs/docs/guides/ERRORS.mdx
@@ -55,7 +55,7 @@ See [the `CameraError.ts` file](https://github.com/mrousavy/react-native-vision-
 
 ### Runtime Errors
 
-The `CameraRuntimeError` represents any kind of error that occured while mounting the Camera view, or an error that occured during the runtime.
+The [`CameraRuntimeError`](/docs/api/classes/CameraRuntimeError) represents any kind of error that occured while mounting the Camera view, or an error that occured during the runtime.
 
 The `<Camera />` UI Component provides an `onError` function that will be invoked every time an unexpected runtime error occured.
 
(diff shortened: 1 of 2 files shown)


mrousavy:
whoops wrong branch

PR #2401, opened 2024-01-16
feat: Expose `PixelFormat` as native type

mrousavy:
## What

Exposes `PixelFormat` as a native Objective-C type for `Frame`.

Before:

```swift
let format = frame.pixelFormat // <-- String
if format == "yuv" { 
  // ...
```

After:

```swift
let format = frame.pixelFormat // <-- PixelFormat
if format == .yuv { 
  // ...
```


mrousavy:
lol - revisiting my old PRs really shows how much effort I tried to make to make it more typesafe and all of those PRs are just now baked into a single library: **Nitro.**

This is obsolete now with Nitro as Nitro natively has such typed enums

PR #2414, opened 2024-01-18
feat: Lower `minSdkVersion` from 26 to 23

mrousavy:
## What

Lowers the minimum required SDK version from 26 to 23.


Julesssss:
Thanks, we'll have no issues with this min version 👍

mrousavy:
This is going to be problematic for VisionCamera though, as multiple pieces of the codebase rely on `minSdk` 26:

- `VideoPipeline`: The entire video pipeline will not work anymore if not running on minSdk 26. Video recordings and Frame Processing will error out then.
- Creating a session: I need to add workarounds there as I use an API from SDK 24.
- HDR: The entire HDR pipeline needs to be disabled on <26.

Julesssss:
Yeah, sorry to be a pain about the min-supported version. 

I get that most people don't care about this and would much rather take the newer features. In our case though I'm not sure it's a trade-off worth making. I also wouldn't want us to be the sole reason for having to drop lib features, so maybe we should just seek an alternative? Let me know what you think

PR #1702, opened 2023-08-21
feat: Add `enableShutterSound` prop to `takePhoto()` 🔊

mrousavy:
## What

Adds option to control the shutter sound on both iOS and Android to the `takePhoto(..)` function.

Default value is `true`. 

This will play:

- [`MediaActionSound.SHUTTER_CLICK`](https://developer.android.com/reference/android/media/MediaActionSound#SHUTTER_CLICK) on Android
- [`photoShutter.caf` (1108)](https://github.com/TUNER88/iOSSystemSoundsLibrary#list-of-systemsoundids) on iOS

## Related issues

* resolves https://github.com/mrousavy/react-native-vision-camera/issues/1044
* resolves https://github.com/mrousavy/react-native-vision-camera/issues/311
* resolves https://github.com/mrousavy/react-native-vision-camera/issues/713
* resolves https://github.com/mrousavy/react-native-vision-camera/issues/565

PR #2433, opened 2024-01-25
feat: Silent and Vibrate mode, shutter sound is mute

quannq-vvt:
## What
when this PR appears: https://github.com/mrousavy/react-native-vision-camera/pull/1702 then android when taking a snap with silent mode and vibrate mode always has sound, this makes the app like an error on Android

## Tested on
Zflip 5
Samsung a54

### DIFF
--- a/package/android/src/main/java/com/mrousavy/camera/core/CameraSession.kt
+++ b/package/android/src/main/java/com/mrousavy/camera/core/CameraSession.kt
@@ -592,7 +592,7 @@ class CameraSession(private val context: Context, private val cameraManager: Cam
       orientation
     )
     Log.i(TAG, "Photo capture 1/3 - starting capture...")
-    val result = captureSession.capture(captureRequest, enableShutterSound)
+    val result = captureSession.capture(captureRequest, enableShutterSound, context)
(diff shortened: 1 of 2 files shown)


mrousavy:
Hey - thanks for this PR mate, what happens **without** your change though? Does the app crash? What error is being thrown?

quannq-vvt:
> Hey - thanks for this PR mate, what happens **without** your change though? Does the app crash? What error is being thrown?

It does not crash or throw an error, it's normal except if the user uses Android with silent or vibrate mode, it always has a shutter sound. It's so uncomfortable if your phone is silent but still makes a sound

mrousavy:
Ah, I see! Okay well this is more of a "how do we want it to behave question"...

quannq-vvt:
if you find it reasonable, hope you update soon! Thanks for your job

omerts:
Hi, would like to see this merged as well please. We got a complient about this, since if the phone is in silent mode camera should not make sounds (especially since it makes the shutter sound by default).

mrousavy:
I just fixed this in V4 and released it in the latest beta. Thanks @quannq-vvt, I took your approach and adjusted it a bit, but the idea is the same!

quannq-vvt:
done

PR #2435, opened 2024-01-25
feat: Don't play shutter sound if phone is on silent

mrousavy:
## What

Based off of https://github.com/mrousavy/react-native-vision-camera/pull/2433, author said there were errors if the phone is on silent.

I just restructured the code a bit to move the logic into CameraView (where it belongs imo) - it does the same thing.

I have yet to verify if this is actually an issue though.


mrousavy:
V5

PR #2441, opened 2024-01-26
feat: Use `yuv` by default if no `pixelFormat` is set

mrousavy:
## What

If no pixelFormat is set and the user is using a Frame Processor, use `yuv` as a default format.

### DIFF
--- a/package/src/Camera.tsx
+++ b/package/src/Camera.tsx
@@ -470,6 +470,7 @@ export class Camera extends React.PureComponent<CameraProps, CameraState> {
     }
 
     const shouldEnableBufferCompression = props.video === true && frameProcessor == null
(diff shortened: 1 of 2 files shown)


package/src/Camera.tsx
```diff
@@ -470,6 +470,7 @@ export class Camera extends React.PureComponent<CameraProps, CameraState> {
     }
 
     const shouldEnableBufferCompression = props.video === true && frameProcessor == null
+    const pixelFormat = props.pixelFormat ?? (frameProcessor == null ? 'yuv' : 'native')
```
bruno-centanaro, line 473: shouldn't this be the other way around? Unless i'm missing something this sets 'yuv' if there is no frame processor

mrousavy: Shit, nice catch man thank you!

PR #2454, opened 2024-01-29
fix: Fix `NullPointerException` in `initFrameProcessorPlugin` if plugin is not found

vanenshi:
## What

If you call the `VisionCameraProxy.initFrameProcessorPlugin('processor');` and the specific  processor were not available, the whole program crash with this error 
```
JNI DETECTED ERROR IN APPLICATION: JNI GetObjectRefType called with pending exception java.lang.NullPointerException: getPlugin(name, this, options) must not be null
```

this fix mark the getPlugin function as nullable

## Changes

mark the getPlugin function as nullable

## Tested on

Android simulator, and real android device

### DIFF
--- a/package/android/src/main/java/com/mrousavy/camera/frameprocessor/FrameProcessorPluginRegistry.java
+++ b/package/android/src/main/java/com/mrousavy/camera/frameprocessor/FrameProcessorPluginRegistry.java
@@ -36,7 +36,7 @@ public class FrameProcessorPluginRegistry {
 
     @DoNotStrip
     @Keep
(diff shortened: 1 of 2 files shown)


mrousavy:
Hey, thank you for your contribution!! This is perfect, nice catch.

I just added two annotations to make this even clearer (with those I would've caught the mistake as bridging Java -> Kotlin requires such annotations, otherwise non-null is assumed)

Once the CI goes green we can merge 🚀

mrousavy:
merged! 🚀

PR #2499, opened 2024-02-02
fix: Improve Templates type

LucasGarcez:
## What

The predefined template has a generic type, so the TypeScript and autocomplete can't get a typo issue. This PR defines a literal type to register all keys and help users when using `Templates`.   

## Changes

This PR changes the `Templates` const type. 

## Tested on

image

### DIFF
--- a/package/src/devices/Templates.ts
+++ b/package/src/devices/Templates.ts
@@ -1,9 +1,17 @@
 import { Dimensions } from 'react-native'
 import { FormatFilter } from './getCameraFormat'
 
-type TTemplates = {
-  [key: string]: FormatFilter[]
-}
+type PredefinedTemplates =
+  | 'Video'
+  | 'Video60Fps'
+  | 'VideoSlowMotion'
+  | 'VideoStabilized'
+  | 'Photo'
+  | 'PhotoPortrait'
+  | 'FrameProcessingYUV'
+  | 'FrameProcessingRGB'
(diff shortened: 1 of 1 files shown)


mrousavy (commented):
Ah, lol whoops, nice catch!! thanks for the PR.

One question; could we maybe use `as const` at the end of the `Templates` declaration instead of your approach?

I'm trying to find a way that would not require me to write the keys (`Video`, `Video60Fps`, ...) twice, but instead only once and have the value type (`FormatFilter[]`) still be typed...?

LucasGarcez:
> Ah, lol whoops, nice catch!! thanks for the PR.
> 
> One question; could we maybe use `as const` at the end of the `Templates` declaration instead of your approach?
> 
> I'm trying to find a way that would not require me to write the keys (`Video`, `Video60Fps`, ...) twice, but instead only once and have the value type (`FormatFilter[]`) still be typed...?

Thanks for the quick answer. It would be great to declare the keys just once, but as far as I know, there is no resource on TypeScript to do it. If you check the [TypeScript documentation for the Record utility](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) the example is exactly the same as our use case here. So, I assume `Record<Keys, Type>` is the best approach.

mrousavy:
Great, LGTM then, thank you Lucas! ❤️

PR #2509, opened 2024-02-05
fix: Fix crash in `HardwareBuffer.isSupported`

mrousavy:
## What

Might fix a crash reported by @bglgwyng


bglgwyng:
Since `GPU + CPU HardwareBuffers are supported!` is logged, Galaxy S21 returns `HardwareBuffer.isSupported(width, height, hardwareBufferFormat, 1, flags)` as true.

https://gist.github.com/bglgwyng/bf20a52175347f02e71966a7c60f8cae

mrousavy:
Okay, thanks!

PR #2519, opened 2024-02-06
fix: Fix PreviewView being stretched

mrousavy:
## What

**HUGE thanks to the [GitHub sponsors](https://github.com/sponsors/mrousavy) who sponsored me to fix this bug!** 💖

Fixes the PreviewView being stretched by awaiting the `SurfaceHolder`'s size change before configuring the Camera session on Android.

This will delay the Camera session startup time by a bit (130ms on my Huawei P10 for first init, 20ms afterwards), so https://github.com/mrousavy/react-native-vision-camera/issues/2518 is becoming even more relevant now.

I am not sure if there is another way to do this, but it seems like we really need to wait for the Surface View to resize properly before configuring the Camera with it. Weird.

Also prevents the display from going to sleep while the Camera is active, as an addition to https://github.com/mrousavy/react-native-vision-camera/pull/2421

## Changes

- Waits for the SurfaceView's Surface to actually change in size before configuring the Camera (fixes preview stretching)
- Flips `getMaximumPreviewSize()` dimensions to reflect sensor natural orientation
- Prevents the display from going to sleep while Camera is active (addition to https://github.com/mrousavy/react-native-vision-camera/pull/2421)

## Related issues

- Fixes https://github.com/mrousavy/react-native-vision-camera/issues/2311
- Fixes https://github.com/mrousavy/react-native-vision-camera/issues/452
- Fixes https://github.com/mrousavy/react-native-vision-camera/issues/2107
- Fixes https://github.com/mrousavy/react-native-vision-camera/issues/2142
- Fixes https://github.com/mrousavy/react-native-vision-camera/issues/2311
- Fixes https://github.com/mrousavy/react-native-vision-camera/issues/2380

PR #2522, opened 2024-02-07
Fix photo not saving in Android example app on Android 13+

Menardi:
## What

This PR fixes an issue in the example app which prevented the saving of photos on Android 13+. I came across this when testing #2519 and needed to save some example photos.

On Android 13+, requesting the `WRITE_EXTERNAL_STORAGE` permission immediately denies, without asking the user. Instead we need to use scoped storage. The `@react-native-camera-roll/camera-roll` plugin being used already supports using scoped storage for saving images on Android 13+, so all we need to do is skip the permission check in that case.

## Changes

* This PR skips the storage permission check in the example app on Android 13+

## Tested on

* Pixel 7a, Android 14


mrousavy:
Awesome, nice catch thank you!

PR #2564, opened 2024-02-15
fix: Fix `PreviewView` stretching on Android (now finally a real fix)

mrousavy:
## What

So in the previous PR (https://github.com/mrousavy/react-native-vision-camera/pull/2519) I found a solution to why the PreviewView was sometimes stretched and sometimes not - it was simply a race condition happening because `SurfaceHolder::setFixedSize` was updating the size in the background, and I was already trying to use it when it might not have updated yet.

While this PR fixed about 50% of the preview issues, some users were still experiencing stretched preview views. 
I now took another look at this, and found out this was happening because the Preview View was not listening to layout changes from the Android UI. So this now fixes this by using the proper `ViewGroup` layout with a hierarchy fitter listener, properly updating the preview view sizes now 💪 

## Related issues

- Fixes https://github.com/mrousavy/react-native-vision-camera/issues/2543
- Fixes https://github.com/mrousavy/react-native-vision-camera/issues/2541

PR #2565, opened 2024-02-15
fix: Take Orientation into account for `PreviewView`

mrousavy:
## What

An addition to https://github.com/mrousavy/react-native-vision-camera/pull/2564, we now also take Camera Sensor Orientation into account when computing Preview Size.


ismaelsousa (commented):
cool

PR #2574, opened 2024-02-16
fix: Double check HardwareBuffer creation if it is actually supported

mrousavy:
## What

When using `enableGpuBuffers={true}`, the pipeline tries to use GPU based buffers for the `ImageReader`/`ImageWriter` pipeline.

Previously we used [`HardwareBuffer.isSupported(...)`](https://developer.android.com/reference/android/hardware/HardwareBuffer#isSupported(int,%20int,%20int,%20int,%20long)) to check if we can use those GPU buffers (the `HardwareBuffer.USAGE_GPU_SAMPLED_IMAGE` flag), but for some reason on some Samsung devices this returned `true` and then crashed later on - which is either really stupid from Samsung's side (because they are simply lying in their APIs), or I am missing something on my side.

Either way, with this PR we now also additionally create a test HardwareBuffer using [`HardwareBuffer.create(...)`](https://developer.android.com/reference/android/hardware/HardwareBuffer#create(int,%20int,%20int,%20int,%20long)) to see if creation goes successfully. If this fails, then we can avoid crashing the app since it means GPU buffers are not supported, even though the `isSupported(...)` API said they are.

**I need help from the community** for this one though:

- If your app crashes with `enableGpuBuffers={true}` when starting a Video Recording while having a Frame Processor, can you try if this PR fixes the issue? 
- Can anyone create a bug report on Android or Samsung forums to see if this is a bug in Android, or maybe an error on my side? There's just no documentation on this available.

### DIFF
--- a/package/android/src/main/java/com/mrousavy/camera/core/VideoPipeline.kt
+++ b/package/android/src/main/java/com/mrousavy/camera/core/VideoPipeline.kt
@@ -238,7 +238,14 @@ class VideoPipeline(
   private fun supportsHardwareBufferFlags(flags: Long): Boolean {
     val hardwareBufferFormat = format.toHardwareBufferFormat()
     try {
-      return HardwareBuffer.isSupported(width, height, hardwareBufferFormat, 1, flags)
+      val isSupported = HardwareBuffer.isSupported(width, height, hardwareBufferFormat, 1, flags)
+      if (!isSupported) return false
+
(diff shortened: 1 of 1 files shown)


mrousavy:
According to @bglgwyng, this PR does not change anything - HardwareBuffer creation with that flags still succeeds on some Samsungs, but the actual `ImageReader` call later on fails/crashes. So we can close this unfortunately

PR #2599, opened 2024-02-21
Boolean.parse string values from gradle properties

jthure:
## What

The properties `VisionCamera_disableFrameProcessors` and `VisionCamera_enableCodeScanner ` when defined is returned as strings by `rootProject.ext.get`. So when for example setting `VisionCamera_enableCodeScanner=false` in `gradle.properties`, the read value of it is the string`"false"`. It is later used in a  conditional, but `if("false")` evaluates to `true` which is not the intended behaviour.

## Changes

Use `Boolean.parse` to parse props that are expected to have a boolean value

### DIFF
--- a/package/android/build.gradle
+++ b/package/android/build.gradle
@@ -46,6 +46,10 @@ def safeExtGet(prop, fallback) {
   rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
 }
 
+def safeExtGetBool(prop, fallback) {
+  Boolean.parseBoolean("${safeExtGet(prop, fallback)}")
+}
+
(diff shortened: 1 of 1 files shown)


mrousavy:
Hey - thanks for your PR, but the line is a bit messy - could you instead create a separate function below `safeExtGet` called `safeExtGetBool` that does this parsing? 

Thank you!

jthure:
> Hey - thanks for your PR, but the line is a bit messy - could you instead create a separate function below `safeExtGet` called `safeExtGetBool` that does this parsing?
> 
> Thank you!

Done!

mrousavy:
Awesome, thanks man LGTM!

PR #2624, opened 2024-03-01
feat: Use CameraX for device details

mrousavy:
## What

Uses CameraX instead of Camera2 for getting device details.

This makes VisionCamera future proof, if at some point a new Camera3 framework gets released as those device details should be generic (as long as CameraX still gets maintained, but AndroidX is a pretty safe bet).

Also, this adds support for FAKE cameras, and Camera1 cameras, meaning more devices will be supported now.

There are however a few things where I couldn't use CameraX but instead had to safely downcast to Camera2:

- `physicalDeviceIds`: CameraX doesn't have an abstraction over that, I needed to downcast to Camera2 for that.
- `sensorOrientation`: CameraX doesn't have an abstraction over that, I needed to downcast to Camera2 for that.
- `minFocusDistance`: CameraX doesn't have an abstraction over that, I needed to downcast to Camera2 for that.
- `isoRange`: CameraX doesn't have an abstraction over that, I needed to downcast to Camera2 for that.
- `maxFieldOfView`: CameraX doesn't have an abstraction over that, I needed to downcast to Camera2 for that.
- `formats`: CameraX doesn't have an abstraction over possible formats/stream resolution configurations/combinations, so I had to kinda manually tinker my way around here.

Would be great if they added those things to the `CameraInfo` API.


TheGhostFish:
Thanks for your great work 🙏

Version 4 camera device formats is missing high photo resolutions, that existed in versions up to 3.9.2
I think we need to use getSupportedHighResolutions along with getSupportedResolutions.
I created this patch to fix:
```
diff --git a/node_modules/react-native-vision-camera/android/src/main/java/com/mrousavy/camera/core/CameraDeviceDetails.kt b/node_modules/react-native-vision-camera/android/src/main/java/com/mrousavy/camera/core/CameraDeviceDetails.kt
index adf84c7..bace54f 100644
--- a/node_modules/react-native-vision-camera/android/src/main/java/com/mrousavy/camera/core/CameraDeviceDetails.kt
+++ b/node_modules/react-native-vision-camera/android/src/main/java/com/mrousavy/camera/core/CameraDeviceDetails.kt
@@ -115,7 +115,7 @@ class CameraDeviceDetails(private val cameraInfo: CameraInfo, extensionsManager:
     dynamicRangeProfiles.forEach { dynamicRange ->
       val qualities = videoCapabilities.getSupportedQualities(dynamicRange)
       val videoSizes = qualities.map { it as ConstantQuality }.flatMap { it.typicalSizes }
-      val photoSizes = cameraInfoInternal.getSupportedResolutions(ImageFormat.JPEG)
+      val photoSizes = cameraInfoInternal.getSupportedResolutions(ImageFormat.JPEG) + cameraInfoInternal.getSupportedHighResolutions(ImageFormat.JPEG)
       val fpsRanges = cameraInfo.supportedFrameRateRanges
       val minFps = fpsRanges.minOf { it.lower }
       val maxFps = fpsRanges.maxOf { it.upper }
```

mrousavy:
hm, but can we actually use such high res formats? can you try if you can actually start a session and capture a photo with high resolution formats?

TheGhostFish:
I tried it on two devices so far, it captures high resolution formats with no issues.
Samsung Galaxy A23
Redmi Note 12
Will try on more devices and let you know.

mrousavy:
Okay, if you can confirm that this actually works and captures higher resolution than before, could you please create a PR with that change? thank you :)

PR #2643, opened 2024-03-08
feat: Support orientation change on iOS

ldalzottomp:
## What

While integrating the react-native-vision-camera library in our application, we stumbled upon an issue.
When rotating the device, the camera feed were not folloing the interface orientation.

## Changes

I tried to add support for auto orientation on iOS by listening to the orientation change notification.
Then, update the connection input orientations based on UI interface orientation.
I also updated the input orientation on session start.

I have tried to integrate this feature as close as possible as the current architecture. As I am not familiar with camera and your project, feel free to tell me what to change.

I did not observed the initial faulty behaviour on different Android devices.

Before :

https://github.com/mrousavy/react-native-vision-camera/assets/91324462/483d190e-3218-40f8-99e9-3b9835bac51f

After : I entered in the device home screen at some point

https://github.com/mrousavy/react-native-vision-camera/assets/91324462/2b4e4ea2-b061-4f37-952c-1f3bc55011b4

## Tested on

iPad Pro (11-inch) (2nd generation). With both orientation locked screen and unlocked.
iPhone 11. No orientation change as the UI is orientation locked.

### DIFF
--- a/package/ios/Core/CameraSession.swift
+++ b/package/ios/Core/CameraSession.swift
@@ -8,6 +8,7 @@
 
 import AVFoundation
 import Foundation
+import UIKit
 
 /**
  A fully-featured Camera Session supporting preview, video, photo, frame processing, and code scanning outputs.
@@ -61,6 +62,10 @@ class CameraSession: NSObject, AVCaptureVideoDataOutputSampleBufferDelegate, AVC
                                            selector: #selector(audioSessionInterrupted),
                                            name: AVAudioSession.interruptionNotification,
                                            object: AVAudioSession.sharedInstance)
+    NotificationCenter.default.addObserver(self,
+                                           selector: #selector(deviceOrientationChanged),
+                                           name: UIDevice.orientationDidChangeNotification,
+                                           object: nil)
   }
 
   deinit {
@@ -73,6 +78,9 @@ class CameraSession: NSObject, AVCaptureVideoDataOutputSampleBufferDelegate, AVC
     NotificationCenter.default.removeObserver(self,
                                               name: AVAudioSession.interruptionNotification,
                                               object: AVAudioSession.sharedInstance)
+    NotificationCenter.default.removeObserver(self,
+                                              name: UIDevice.orientationDidChangeNotification,
+                                              object: nil)
   }
 
   /**
@@ -233,6 +241,7 @@ class CameraSession: NSObject, AVCaptureVideoDataOutputSampleBufferDelegate, AVC
     // Start/Stop session
     if configuration.isActive {
       captureSession.startRunning()
+      updateCaptureSessionInputOrientation()
       delegate?.onCameraStarted()
     } else {
       captureSession.stopRunning()
@@ -287,6 +296,36 @@ class CameraSession: NSObject, AVCaptureVideoDataOutputSampleBufferDelegate, AVC
       // restart capture session after an error occured
       CameraQueues.cameraQueue.async {
         self.captureSession.startRunning()
+        self.updateCaptureSessionInputOrientation()
+      }
+    }
+  }
+
+  @objc
+  func deviceOrientationChanged(notification _: Notification) {
+    updateCaptureSessionInputOrientation()
+  }
+
+  func updateCaptureSessionInputOrientation() {
+    // Lock the camera session, to be sure that it doesn't conflict with other session changes.
+    CameraQueues.cameraQueue.async {
+      // Because statusBarOrientation requires it, as it's an UI operation.
+      DispatchQueue.main.async {
+        let interfaceOrientation = UIApplication.shared.statusBarOrientation // Set up orientation for all connections.
+        for connection in self.captureSession.connections {
+          switch interfaceOrientation {
+          case .portrait:
+            connection.videoOrientation = .portrait
+          case .landscapeLeft:
+            connection.videoOrientation = .landscapeLeft
+          case .landscapeRight:
+            connection.videoOrientation = .landscapeRight
+          case .portraitUpsideDown:
+            connection.videoOrientation = .portraitUpsideDown
+          case .unknown:
+            connection.videoOrientation = .portrait
+          }
+        }
       }
     }
   }


mrousavy:
Thanks for your PR, but I wonder why you didn't use the `setOrientation` extension I already wrote? https://github.com/mrousavy/react-native-vision-camera/blob/6b71d4fd33a6f450fd70ddf8f077fe245de4624c/package/ios/Extensions/AVCaptureOutput%2Bmirror.swift#L24-L58

Is that different from your code?

ldalzottomp:
>  but I wonder why you didn't use the setOrientation extension I already wrote?

@mrousavy I did that as a first try !

The issue that I encountered is that it's updating the output.

In the case of a phone.
The user can put his phone on landscape mode.
And in this case, the developer may want the output in landscape mode with the "orientation" prop. But may not want to switch the input orientation because the UI orientation is locked.

Basically, that would be good for tablet, but not for phones or locked tablets.

Here, I feel that we really want something "UI wise". Which is : does the camera feed orientation follows the UI orientation or not.
Basically, I feel that we need to have a distinction between inputOrientation and outputOrientation ("orientation" prop). To overcome the issue I just described.

---

So the difference that I see is that the piece of code you highlighted apply to the output (AVCaptureOuput) of the photo capture. Which I don't think this is always what we want. Because the input (camera feed) and the output can be different.

But the logic of this extension can be extracted somewhere else to be reused.

ldalzottomp:
Oh sorry I misunderstood your comment. I thought that you were talking about the orientation in the config structure.

ajp8164:
This may be related.
https://github.com/mrousavy/react-native-vision-camera/pull/1619

zero2603:
Hi, I am using **react native v0.70.4** and **react-native-vision-camera v2.16.2**, I solved camera orientation for both iOS and Android. And I want to share this trick for everyone. Please follow below steps:

- Install `react-native-sensors`. Then create custom hook `useDeviceRotationSensor.js`:
```
import { useLayoutEffect, useRef } from 'react';
import { Platform } from 'react-native';
import { setUpdateIntervalForType, SensorTypes, gravity } from "react-native-sensors";

/**
 * 
 * Reference: https://github.com/mrousavy/react-native-vision-camera/issues/624
 */
export const useDeviceRotationSensor = (callback) => {
    setUpdateIntervalForType(SensorTypes.gravity, 1000);

    const callbackRef = useRef(callback);
    callbackRef.current = callback;

    useLayoutEffect(() => {
        // We use gravity sensor here because react-native-orientation
        // can't detect landscape orientation when the device's orientation is locked
        const subscription = gravity.subscribe(({ x, y }) => {
            const radian = Math.atan2(y, x);
            const degree = (radian * 180) / Math.PI;

            let rotation = 'landscapeLeft';

            if (Platform.OS === 'android') {
                if (degree > -135) rotation = 'portraitUpsideDown';
                if (degree > -45) rotation = 'landscapeRight';
                if (degree > 45) rotation = 'portrait';
                if (degree > 135) rotation = 'landscapeLeft';
            } else {
                if (degree > -135) rotation = 'portrait';
                if (degree > -45) rotation = 'landscapeLeft';
                if (degree > 45) rotation = 'portraitUpsideDown';
                if (degree > 135) rotation = 'landscapeRight';
            }

            callbackRef.current(rotation, degree);
        });

        return () => subscription.unsubscribe();
    }, []);
};
```
- Open `/node_modules/react-native-vision-camera/ios/CameraView+Orientation.swift`, then replace current code with below code (reference to comment: https://github.com/mrousavy/react-native-vision-camera/issues/1455#issuecomment-1742013679) :
```
//
//  CameraView+Orientation.swift
//  VisionCamera
//
//  Created by Marc Rousavy on 04.01.22.
//  Copyright © 2022 mrousavy. All rights reserved.
//

import Foundation
import UIKit

extension CameraView {
  /// Returns the current _interface_ orientation of the main window
  private var windowInterfaceOrientation: UIInterfaceOrientation {
    if #available(iOS 13.0, *) {
      return UIApplication.shared.windows.first?.windowScene?.interfaceOrientation ?? .unknown
    } else {
      return UIApplication.shared.statusBarOrientation
    }
  }

  /// Orientation of the input connection (preview)
  private var inputOrientation: UIInterfaceOrientation {
    return windowInterfaceOrientation
  }

  // Orientation of the output connections (photo, video, frame processor)
  private var outputOrientation: UIInterfaceOrientation {
    if let userOrientation = orientation as String?,
       let parsedOrientation = try? UIInterfaceOrientation(withString: userOrientation) {
      // user is overriding output orientation
      return parsedOrientation
    } else {
      // use same as input orientation
      return inputOrientation
    }
  }

  internal func updateOrientation() {
    // Updates the Orientation for all rotable connections (outputs) as well as for the preview layer
    DispatchQueue.main.async {
      // `windowInterfaceOrientation` and `videoPreviewLayer` should only be accessed from UI thread
      let isMirrored = self.videoDeviceInput?.device.position == .front

      self.videoPreviewLayer.connection?.setInterfaceOrientation(self.inputOrientation)

      let connectionOrientation = self.outputOrientation
      self.cameraQueue.async {
        // Run those updates on cameraQueue since they can be blocking.
        self.captureSession.outputs.forEach { output in
          output.connections.forEach { connection in
            if connection.isVideoMirroringSupported {
              connection.automaticallyAdjustsVideoMirroring = false
              connection.isVideoMirrored = isMirrored
            }
            connection.setInterfaceOrientation(connectionOrientation)
          }
        }
      }
    }
  }
}
```
- Install `patch-package`, then run command: `yarn patch-package react-native-vision-camera`
- In your file which is using <Camera>, you can code as below:
```
...
const [deviceOrientation, setDeviceOrientation] = useState('portrait');
useDeviceRotationSensor(rotation => {
    if (rotation != deviceOrientation && !isRecording) {
        setDeviceOrientation(rotation);
    }
});
...
return (
    <Camera
        // ...otherProps
        orientation={deviceOrientation}
    />
)
``` 

Hope this helps!!

IamHungry113:
any update here ? cause I encountered the same problem, in version 3.9.1, if the app orientation is locked, the output of the camera is not the same as expected， just like @ldalzottomp  said 
so any patch here ?

ldalzottomp:
@IamHungry113 The changes in this PR were enough to cover the use case that I mentioned. So you can grab it and apply it to a local fork. My public fork is here (https://github.com/ldalzottomp/react-native-vision-camera). However there is build issue regarding older iOS version. If your app don't target iOS lower than the requested verion, you can add if conditions to prevent build errors.

I discussed with @mrousavy about this PR and this is not going to be merged as the video feed/frame processor have not been thoroughly tested.

MCervenka:
Thank you, this fixed my problem.

nijatkazimliPG:
The similar thing happens also on Android.

mrousavy:

PR #2661, opened 2024-03-18
fix: Changes relative import paths in `usePreferredCameraDevice.ts`

dh031200:
## What
This PR fixes to reference 'react-native-vision-camera' instead of referencing a relative path.

## Changes
This PR changes import paths in `usePreferredCameraDevice.ts`.

Updated import paths for `CameraDevice` and `useCameraDevices`. 
It now correctly references 'react-native-vision-camera' instead of a relative path. 

Before
```typescript
import { CameraDevice } from '../../../src/CameraDevice'
import { useCameraDevices } from '../../../src/hooks/useCameraDevices'
```
After
```typescript
import { CameraDevice, useCameraDevices } from 'react-native-vision-camera'
```

## Tested on
iPhone SE (3rd generation) 17.3.1


mrousavy:
Nice catch, thank you!

PR #2664, opened 2024-03-18
feat: Basic read-only Frame Processors (`ImageAnalysis` Use-Case)

mrousavy:
## What

Adds support for basic read-only Frame Processors using the `ImageAnalysis` use-case.

By design, those Frame Processors are async, and Skia couldn't really be implemented ontop of this. But we'll figure it out!


ismaelsousa:
I like the use of ImageAnalysis 🎉

PR #2719, opened 2024-04-05
always supply NSLocationWhenInUseUsageDescription string in expo

bruno-centanaro:
## What

After upgrading to react-native-vision-camera 4.0.0-beta.12 and uploading a build to the app store connect I received the following warning:

> ITMS-90683: Missing purpose string in Info.plist - Your app’s code references one or more APIs that access sensitive user data, or the app has one or more entitlements that permit such access. The Info.plist file for the “*.app” bundle should contain a NSLocationWhenInUseUsageDescription key with a user-facing purpose string explaining clearly and completely why your app needs the data. If you’re using external libraries or SDKs, they may reference APIs that require a purpose string. While your app might not use these APIs, a purpose string is still required. For details, visit: https://developer.apple.com/documentation/uikit/protecting_the_user_s_privacy/requesting_access_to_protected_resources.

But I don't use the location at all

## Changes

This PR removes the need of the user to explicitly mention they use location to supply a default NSLocationWhenInUseUsageDescription string

### DIFF
--- a/package/src/expo-plugin/withVisionCamera.ts
+++ b/package/src/expo-plugin/withVisionCamera.ts
@@ -21,11 +21,9 @@ const withCamera: ConfigPlugin<ConfigProps> = (config, props = {}) => {
     config.ios.infoPlist.NSMicrophoneUsageDescription =
       props.microphonePermissionText ?? (config.ios.infoPlist.NSMicrophoneUsageDescription as string | undefined) ?? MICROPHONE_USAGE
   }
-  if (props.enableLocationPermission) {
-    // Location permission
-    config.ios.infoPlist.NSLocationWhenInUseUsageDescription =
-      props.locationPermissionText ?? (config.ios.infoPlist.NSLocationWhenInUseUsageDescription as string | undefined) ?? LOCATION_USAGE
-  }
+  // Location permission
+  config.ios.infoPlist.NSLocationWhenInUseUsageDescription =
+    props.locationPermissionText ?? (config.ios.infoPlist.NSLocationWhenInUseUsageDescription as string | undefined) ?? LOCATION_USAGE
   const androidPermissions = ['android.permission.CAMERA']
   if (props.enableMicrophonePermission) androidPermissions.push('android.permission.RECORD_AUDIO')
   if (props.enableLocationPermission) androidPermissions.push('android.permission.ACCESS_FINE_LOCATION')


mrousavy:
Hm, maybe we should make this even more optional. We don't want everybody to add location string to their app, do we? Does this do any harm?

fobos531:
> Hm, maybe we should make this even more optional. We don't want everybody to add location string to their app, do we? Does this do any harm?

I would say it depends on whether VisionCamera by default stores location EXIF data for each picture that is taken. If the answer is yes, I would say that in that case supplying this string is in fact mandatory.

There was a similar issue in iOS (unrelated to VisionCamera) where viewing very specific images which contain EXIF data about contacts triggered a crash if the NSContacts usage description string was not provided in Info.plist. Probably in such cases iOS tries to request the contacts description in order to check if it can match up the contact data in the image with the information in the user's contacts app. In that sense, I would say the motivation behind the "need" for this string would be similar.

mrousavy:
I think there's also a way to conditionally compile-out the location usage, I can look at that in a sec - can you help me with testing thoughg?

mrousavy:
Are you on the Margelo Community Discord? Please ping me over there in the vision camera channel

rwichmann:
I have this problem too and I think it would be great when this could be conditionally compiled out!

rwichmann:
> Hm, maybe we should make this even more optional. We don't want everybody to add location string to their app, do we? Does this do any harm?

I think an app should use as few permissions as possible. So it could do harm if the Apple Review Team doesn't like your app when you're app seems to use location services, but in reality it doesn't. If the review team is fine with it, I think it won't affect the app negative, however it might confuses other team members and it doesn't feel very clean.

mrousavy:
See https://github.com/mrousavy/react-native-vision-camera/pull/2751

mrousavy:
btw., is it enoughh to just compile out the `CLLocationManager.requestAccess` method, or do I also need to compile out stuff like `CLLocationManager` even though this code can never be reached?

Would be great if you could test that, as the first is pretty easy to do while the latter can be a bit hard to test.

bruno-centanaro:
Hello! Sorry I've been away, I was we a lot of stuff at work. On the weekend i can test that with no problem!

mrousavy:
This has been fixed in #2751 and will be released in the next beta

Issue #2347, opened 2024-01-04
Live Filter Preview✨ 


hashhirr:
### What feature or enhancement are you suggesting?

Do Anyone have any idea how can i apply live preview filters in react native vision camera i have implemented the camera and all the features but what i am missing is the instagram like filters. i tried different sdks like img.ly etc but i have no solution yet.

### What Platforms whould this feature/enhancement affect?

iOS, Android

### Alternatives/Workarounds

i tried few of the react native libraries and it didn't work i am open to even work in native modules if anyone have any solution

### Additional information

- [X] I agree to follow this project's [Code of Conduct](https://github.com/mrousavy/react-native-vision-camera/blob/main/CODE_OF_CONDUCT.md)
- [X] I searched for [similar feature requests in this repository](https://github.com/mrousavy/react-native-vision-camera/issues?q=is%3Aopen+is%3Aissue+label%3A%22✨+enhancement%22) and found none.

mrousavy:
There are multiple discussions like this in this repo, and I answered in each of those.

![Screenshot 2024-01-05 at 15 36 48](https://github.com/mrousavy/react-native-vision-camera/assets/15199031/c02c3035-7a41-43a4-b96f-c452c85243b9)

mrousavy:
See https://github.com/mrousavy/react-native-vision-camera/pull/2727 👀

Issue #2274, opened 2023-12-10
What is the recommended way to draw landmarks over the video❓ 


adrian-bobev:
### Question

What is the recommended way to draw position landmarks of body parts over the live video?

### What I tried

I tried to create a simple MLKit Pose detection FrameProcessor. At the moment I am successfully logged the detected pose landmarks in the console. However, I am wondering what is the recommended way to draw those landmarks.

I saw some mentions of Skia, but not sure what is the current state and how I should use it.

### VisionCamera Version

3.6.13

### Additional information

- [ ] I am using Expo
- [X] I have read the [Troubleshooting Guide](https://react-native-vision-camera.com/docs/guides/troubleshooting)
- [X] I agree to follow this project's [Code of Conduct](https://github.com/mrousavy/react-native-vision-camera/blob/main/CODE_OF_CONDUCT.md)
- [X] I searched for similar questions [in the issues page](https://github.com/mrousavy/react-native-vision-camera/issues?q=is%3Aopen+is%3Aissue+label%3A%22❓+question%22) as well as [in the discussions page](https://github.com/mrousavy/react-native-vision-camera/discussions) and found none.

rodgomesc:
i’m using react-native-skia with stable 30fps

adrian-bobev:
But how you do it? I saw that blog post: https://mrousavy.com/blog/VisionCamera-Pose-Detection-TFLite

However, the code is using ^3.0.0-rc.7 and from that page: https://react-native-vision-camera.com/docs/guides/skia-frame-processors I see that it is removed from the official release. 

This is the reason why I am searching now for the recommended way how this can be implemented

rodgomesc:
ik this is a bit confusing but vision-camera don't have integration with skia natively anymore, this is on hold for now, so you just collect those points you need and use rn skia shapes to draw them on a screen https://shopify.github.io/react-native-skia/docs/shapes/ellipses, it will require some math to normalize between the screen rendered image and the point coordinates but notting to much complex, you can ask for help on discord magelo comunity as well

adrian-bobev:
Thabjs for the answer. Just for future reference what I have done is the following:

```jsx
export default function App() {
  const [position, setPosition] = useState();
  const device = useCameraDevice('front');

  const onObjecteDetected = Worklets.createRunInJsFn(pos => {
    setPosition(pos);
  });

  const frameProcessor = useFrameProcessor(frame => {
    'worklet';

    const obj = detectObject(frame);
[...]

Issue #1755, opened 2023-09-02
❓ Applying color filters to the camera  


pTinosq:
### Question

I have a very simple system set up with a frame processor plugin set up to return a string (just to test communication between processor and ts is functioning).

I would like to somehow apply effects to my images (contrast, grain, brightness etc.) and I am not too sure how to go about doing this.

Would you do this in the frame processor plugin or is it not needed at all?

### What I tried

I have tried to return an Image from the frame processor with errors being thrown. Not too sure how to send an edited image back,

### VisionCamera Version

3.0.0

### Additional information

- [ ] I am using Expo
- [X] I have read the [Troubleshooting Guide](https://react-native-vision-camera.com/docs/guides/troubleshooting)
- [X] I agree to follow this project's [Code of Conduct](https://github.com/mrousavy/react-native-vision-camera/blob/main/CODE_OF_CONDUCT.md)
- [X] I searched for similar questions [in the issues page](https://github.com/mrousavy/react-native-vision-camera/issues?q=is%3Aopen+is%3Aissue+label%3A%22❓+question%22) as well as [in the discussions page](https://github.com/mrousavy/react-native-vision-camera/discussions) and found none.

mrousavy:
Hey there was some work to bring Skia into VisionCamera which allowed you to do all of that. I decided to remove it as it greatly increased the complexity of VisionCamera. If you want that for your business, reach out to my agency (margelo.io) and we can build a tailored solution for you :)

pTinosq:
Ah that's fair enough. So for hobby/free projects there is no support for this sort of thing?

mrousavy:
nope, not at the moment.

mrousavy:
See https://github.com/mrousavy/react-native-vision-camera/pull/2727 👀

Issue #1215, opened 2022-08-25
Is it possible to do LIVE TEXT with react-native-vision-camera❓ 


iamabdulhaseeb:
### Question

I would like to implement a functionality that the user can place a camera on an image and he can copy text from image at that point, I have researched a lot but could not find any resource to do like this,
Is it possible to do this with this package?

### VisionCamera Version

^2.14.1

### Additional information

- [ ] I am using Expo
- [X] I have read the [Troubleshooting Guide](https://mrousavy.github.io/react-native-vision-camera/docs/guides/troubleshooting)
- [X] I agree to follow this project's [Code of Conduct](https://github.com/mrousavy/react-native-vision-camera/blob/main/CODE_OF_CONDUCT.md)
- [X] I searched for similar questions [in the issues page](https://github.com/mrousavy/react-native-vision-camera/issues?q=is%3Aopen+is%3Aissue+label%3A%22❓+question%22) as well as [in the discussions page](https://github.com/mrousavy/react-native-vision-camera/discussions) and found none.

justindavies:
I'm looking at this too - but it should be possible with a frame processor and the new Live Text things with IOS 16

iamabdulhaseeb:
Hey justin
If you are free these days
Can we both look together on this thing because I have been stucked on this
thing for days now.
I am creating my own app in which I wanted to implement this

marcelxsilva:
HI @iamabdulhaseeb,

I am using [`vision-camera-ocr`](https://github.com/aarongrider/vision-camera-ocr) to recognize text by camera, I don't understand how I can copy text but text can be shown in terminal easily

iamabdulhaseeb:
Hi @marcelxsilva , thank you for recommending that package, I will try that very soon :)

sampadanaik:
> HI @iamabdulhaseeb,
> 
> I am using [`vision-camera-ocr`](https://github.com/aarongrider/vision-camera-ocr) to recognize text by camera, I don't understand how I can copy text but text can be shown in terminal easily

can we use for ios?

marcelxsilva:
> > HI @iamabdulhaseeb,
> > I am using [`vision-camera-ocr`](https://github.com/aarongrider/vision-camera-ocr) to recognize text by camera, I don't understand how I can copy text but text can be shown in terminal easily
> 
> can we use for ios?

My app is only to Android, I believe that can use in iOS, without problems!!

mrousavy:
[...]

PR #2727, opened 2024-04-08
feat: Skia for V4! 🥳 

mrousavy:
## What

Skia integration for V4! 🥳 

TL;DR: Allows you to "draw stuff onto the Camera Frame" in realtime.
This is different to just adding a view ontop of the `<Camera>`, because this actually draws inside the Frame buffer, meaning this can be extended to also be applied to captured photos and videos in the future. For now it's preview only.

For example, you can draw color filters, bounding boxes or circles for faces/qr-codes, face or license plate blurring, color corrections, animations on the frame, etc etc

### Backstory

After almost a year since the first idea and prototype of integrating Skia into VisionCamera, we can finally see it happen.
The main issues back then (July/August 2023) were that the react-native-skia dependency in VisionCamera's codebase was too harsh. It had a strong native dependency, we needed to add a bunch of native views including Metal/OpenGL Surface Views for preview, Choreographers for syncing, setting up Metal and OpenGL contexts, adding a native dependency and header include paths on libskia and all of the other skia companion libs, and we had a bunch of complicated nested branches in the Camera codebase where we switched logic between a normal Frame Processor and a Skia Frame Processor.
All of that is a thing from the past now, since I implemented a way to only expose a handle to the Frame ("platform buffer") to JS, and thanks to @wcandillon's amazing help (thank you so much for your time!) we got a `MakeImageFromPlatformBuffer(..)` function into react-native-skia.
Now we can create an `SkImage` from a `Frame`, straight from JS.

### API

In addition to `useFrameProcessor` ("readonly" Frame Processors) a new API has been introduced: `useSkiaFrameProcessor`.

In contrast to the `useFrameProcessor` hook, with the `useSkiaFrameProcessor` hook you now get `DrawableFrame`s instead of normal `Frame`s.

The `DrawableFrame` has a `render(..)` function to render itself (otherwise nothing gets rendered), and all other functions from `Frame` and `SkCanvas`.

For example, to render the image with a blur filter:

```tsx
const frameProcessor = useSkiaFrameProcessor((frame) => {
  'worklet'

  // 1. create blur filter
  const blurRadius = 10
  const blurFilter = Skia.ImageFilter.MakeBlur(blurRadius, blurRadius, TileMode.Repeat, null)

  // 2. wrap blur filter in Paint
  const paint = Skia.Paint()
  paint.setImageFilter(blurFilter)

  // 3. render image with the blur filter paint
  frame.render(paint)
}, [])
```

Or to draw a red rectangle on the middle of the Frame:

```tsx
const frameProcessor = useSkiaFrameProcessor((frame) => {
  'worklet'

  // 1. render frame normally
  frame.render() 

  // 2. create a red paint
  const redPaint = Skia.Paint()
  const red = Skia.Color('red')
  redPaint.setColor(red)

  // 3. draw a rectangle in the center of the screen
  const rect = Skia.XYWHRect(frame.width / 2, frame.height / 2, 150, 150)
  frame.drawRect(rect, redPaint)
}, [])
```

On Android, make sure to set the Camera's `pixelFormat={...}` to `"rgb"`. On iOS, you can use `"yuv"` or `"rgb"`, where `"yuv"` is typically more efficient.

> Note: Skia Frame Processors require react-native-reanimated (>= 3.x.x) and @shopify/react-native-skia (>= 1.2.0)

### Required PRs

1. ✅ The [`MakeImageFromPlatformBuffer(...)` PR in react-native-skia](https://github.com/Shopify/react-native-skia/pull/2344)
2. ⏳ The [`SkData` -> Metal Texture change PR in react-native-skia](https://github.com/Shopify/react-native-skia/pull/2356) (otherwise buffers are in wrong pixelformat and look blue)
3. ⏳ The [YUV CVPixelBuffers PR in react-native-skia](https://github.com/Shopify/react-native-skia/pull/2357) (otherwise we can only use `pixelFormat="rgb"`, no `pixelFormat="yuv"`/default)

### Example

Example shader running directly on the Camera Frame at 60 FPS:

https://github.com/mrousavy/react-native-vision-camera/assets/15199031/e324eed8-eb4b-4af2-94bf-7197d73d14ef

## Changes

- Adds a new API: `useSkiaFrameProcessor` (or `createSkiaFrameProcessor`)
- Marks the Java `FrameProcessorPlugin`'s `callback` function throwable (`throws Throwable`), you can now throw exceptions in Java Frame Processors
- Adds `getPlatformBuffer()` to `Frame`
- Adds a new prop to the `<Camera>`: `preview={boolean}`. This allows you to manually disable the Preview if not needed. (e.g. when using Skia it will use a Skia Canvas instead of the native preview output)
- Fixes an issue where the camera outputs/cameraId are being accessed before they are initialized (`device/no-device` or `session/no-outputs` error) by early aborting duplicate `configure { ... }` calls.
- Minor improvements around the native `Frame` object synchronizations
- Fixes an issue where the Frame Processor pipeline would just stall after a second on Android (forgot to `close()` the parent `ImageProxy`, whoops)
- Fixes an issue where `Frame.orientation` was wrong on iOS (it is always `portrait` since we rotate buffers accordingly in the pipeline)
- Removes the `"native"` pixel format, native does not exist in a Frame Processor - it's always RGB or YUV
- Removes the `supportedPixelFormats` prop on a Format, both iOS and Android always support YUV and RGB, you are free to choose.
- Throws more precise error messages when a YUV pixel format (HDR or compression) cannot be found on iOS, and automatically fallback to the best format to avoid blackscreens
- Adds a `jsi::Scope` to Frame Processor calls to maybe improve GC calls
- Adds lazy proxy dependencies on react-native-worklets-core, react-native-skia and react-native-reanimated

## Tested on

- iPhone 15 Pro (yuv + rgb)
- iPhone 8 (yuv + rgb)
- Huawei P10 (rgb)
- Samsung S22 (rgb)

## Related issues

- Resolves https://github.com/mrousavy/react-native-vision-camera/issues/2734
- Resolves https://github.com/mrousavy/react-native-vision-camera/issues/2347
- Resolves https://github.com/mrousavy/react-native-vision-camera/issues/1755
- Resolves https://github.com/mrousavy/react-native-vision-camera/issues/724
- Resolves https://github.com/mrousavy/react-native-vision-camera/issues/2274
- Resolves https://github.com/mrousavy/react-native-vision-camera/issues/1215


rodgomesc:
i'm very hyped for this feature!, fantastic work dude 🚀🚀 fun fact: i'm learning a lot with vision camera source code💜💜

mrousavy:
thanks man!

package/ios/Frame Processor/FrameHostObject.mm
```diff
@@ -94,6 +151,30 @@
...
+    auto getPlatformBuffer = JSI_FUNC {
+      // Lock Frame so it cannot be deallocated while we access it
+      std::lock_guard lock(this->_mutex);
+
+      // Box-cast to uintptr (just 64-bit address)
+      CMSampleBufferRef buffer = this->frame.buffer;
+      uintptr_t pointer = reinterpret_cast<uintptr_t>(buffer);
+      jsi::HostFunctionType deleteFunc = [=](jsi::Runtime& runtime, const jsi::Value& thisArg, const jsi::Value* args,
+                                             size_t count) -> jsi::Value {
+        // no-op
```
wcandillon: no     CFRelease(buffer); here?

package/src/dependencies/SkiaProxy.ts
```diff
@@ -0,0 +1,12 @@
...
+type TSkia = typeof Skia
+
+/**
+ * A proxy object that lazy-imports @shopify/react-native-skia as soon as the
+ * caller tries to access a property on {@linkcode SkiaProxy}.
+ *
+ * If @shopify/react-native-skia is not installed, accessing anything on
+ * {@linkcode SkiaProxy} will throw.
+ */
+export const SkiaProxy = createModuleProxy<TSkia>('@shopify/react-native-skia', () => require('@shopify/react-native-skia'))
```
wcandillon, line 12: omg this is cool, much better than we came up with for our lose Reanimated coupling.

package/ios/Frame Processor/FrameHostObject.mm
mrousavy: No, I agree that this might be a bit confusing but in my case the Frame is only valid as long as the `useFrameProcessor` didn't return yet. And I `.dispose()` the SkImage before it returns, so there's no way to use it outside of it's lifetime.

Originally I wanted to use HostObjects or NativeState for this so it is properly ref-counted, but that has a few downsides:
1. It requires explicit typings, so react-native-vision-camera needed to depend on react-native-skia in native code, or vice versa so I can use the C++ types.
2. It adds overhead since we'd be using shared_ptrs. (Locks/Mutexes + Ref count increments)
3. It is unnecessary since the Frame cannot live longer than the `useFrameProcessor` anyways (and shouldn't, because that'd stall the Camera pipeline)

package/src/dependencies/SkiaProxy.ts
mrousavy: Haha yea thanks I was also pretty excited when I figured out I could just use Proxies for that. I was a bit worried that Hermes force-included those `require`s but it works!

package/ios/Frame Processor/FrameHostObject.mm
wcandillon: Yes that makes sense.

mrousavy:
PR is ready 🎉 
I'll wait until https://github.com/Shopify/react-native-skia/pull/2356 and https://github.com/Shopify/react-native-skia/pull/2357 are merged so we can make use of metal texture conversion and YUV formats, then merge this PR and create a beta release 👍

rodgomesc:
i'm facing a weird bug in the preview for my **mi 9t pro**

-  changing the aspect ratio seems to affect the issue
-  if i comment `frameProcessor={frameProcessor}` everything works fine 
-  if aspectRatio==1 the frame rate seems to be very stable at 30fps

> note: this is a old device, my camera do not support 60fps

<table>
  <tr>
    <td align="center">
     <code>videoAspectRatio: SCREEN_HEIGHT / SCREEN_WIDTH</code>
      <a href="https://github.com/mrousavy/react-native-vision-camera/assets/4893591/35191ad1-e622-4cbd-8a5a-4f03c35965d0">
        
      </a>
    </td>
    <td align="center">
     <code>videoAspectRatio: 1                           </code>
      <a href="https://github.com/mrousavy/react-native-vision-camera/assets/4893591/b9558ff7-6f5c-43e9-9df0-27596a630bd5">
        
      </a>
    </td>
  </tr>
</table>

rodgomesc:
Tested on: `iPhone 14 Pro Max`  
Working like a charm. (Didn't merged the pending YUV stuff from Skia pr for the test.)

<a href="https://github.com/mrousavy/react-native-vision-camera/assets/4893591/8d097a5c-7064-41f3-a50f-b3e4f3e7910c">
  
</a>

llli0v0:
Cool!

mrousavy:
> if aspectRatio==1 the frame rate seems to be very stable at 30fps

Probably just uses a much lower resolution then for thhe Frame. @rodgomesc

hashhirr:
damn! took me three months to make cameras in native modules using kotlin and swift after i found out vision don't have filters...but i guess it will help me in the long run

meleffendi:
I'm trying to overlay a counter on the recorded video. What's the recommended way of doing this at the moment?

mrousavy:
Because of the issue described in https://github.com/Shopify/react-native-skia/pull/2368, I have to make sure to cache my `SkSurface`s on a per-thread basis, as my Frame Processor runs on an iOS DispatchQueue which by design uses multiple C++ Threads.

Unfortunately react-native-skia contexts are tied to C++ threads (`thread_local`), so that's the workaround I am now using: https://github.com/mrousavy/react-native-vision-camera/blob/2b16bf600d52dae7880d0666a3cf191bb26942b3/package/src/skia/useSkiaFrameProcessor.ts#L96-L135

In my short tests, a `DispatchQueue` uses between 8 and 12 Threads, so this creates between 8 and 12 `SkSurface`s:

```
 LOG  Creating new 540x960 Surface...
 LOG  Creating new 540x960 Surface...
 LOG  Creating new 540x960 Surface...
 LOG  Creating new 540x960 Surface...
 LOG  Creating new 540x960 Surface...
 LOG  Creating new 540x960 Surface...
 LOG  Creating new 540x960 Surface...
 LOG  Creating new 540x960 Surface...
```

I agree that the ideal solution is not the one proposed in https://github.com/Shopify/react-native-skia/pull/2368 (the `dispatch_queue_set_specific` approach), but rather an approach where we can create and manage a `SkiaContext` in JS, which I can then pass down to the `SkImage` and `SkSurface` functions (as proposed in https://github.com/Shopify/react-native-skia/pull/2373).
However, this is a **huge** refactor and requires careful construction to make non-breaking, and also I am simply not sure how to get the Context into the DOM/View APIs.

Anyways, my approach works for now, but I'd love to find a better solution for this at some point in the future so I can remove the 7-11 unnecessary `SkSurface`s and get down to just 1 `SkSurface` in memory (cleaner code, and more importantly, much less RAM usage).

mrousavy:
Here's a performance comparison between the two approaches:

|                               | CPU usage | RAM usage |
|-------------------------------|-----------|-----------|
| `thread_local`                | 47%       | 411 MB    |
| `dispatch_queue_set_specific` | 45%       | 334 MB    |

As you can see, the `dispatch_queue_set_specific` approach uses 77 MB less than the approach where I create one `SkSurface` per Thread, and uses 2% CPU less.

So I think we definitely want to avoid making SkContext `thread_local`, and instead re-use it throughout the DispatchQueue.

package/src/dependencies/ModuleProxy.ts
```diff
@@ -0,0 +1,23 @@
...
+          target.module = getModule() as TModule
+        } catch (e) {
+          throw new Error(`${name} is not installed!`)
+        }
+      }
+      return target.module[property as keyof typeof holder.module]
+    },
+  })
+  return proxy as unknown as TModule
+}
```
wcandillon, line 23: this is cool 😁

mrousavy:
released in latest V4 beta :)

everyone can try `useSkiaFrameProcessor` now in their apps

efstathiosntonas:
@mrousavy I installed beta 15 and it requires Skia to be installed. Can you please consider on making it optional if possible? I guess not everyone is planning on using Skia processors and bloating the app with [6MB](https://shopify.github.io/react-native-skia/docs/getting-started/installation/#bundle-size) is not optimal. 

~edit: by looking at the changes it seems impossible to avoid it~

mrousavy:
Skia is optional: https://github.com/mrousavy/react-native-vision-camera/blob/b751f2da0b1d1543e417a04f0e3b2b077eaa6cc0/package/package.json#L120-L122

Are you seeing an error or anything?

mrousavy:
I have react-native-skia in `devDependencies`: https://github.com/mrousavy/react-native-vision-camera/blob/b751f2da0b1d1543e417a04f0e3b2b077eaa6cc0/package/package.json#L92

Maybe this is why it gets included in your app as well then? I just need the types, that's why I have it in `devDependencies`...

I don't think this ends up in your release app though, as the `devDependencies` are compiled out. Not sure on this.

efstathiosntonas:
after installation I get red screen that skia is missing

efstathiosntonas:
more info:

<details>
  <summary>Click me</summary>
  
Simulator Screenshot - iPhone 13 - 2024-04-17 at 18 56 57

</details>

geekabhinavp:
Hey @mrousavy, can this be merged now, i see all the required PR being merged in Skia.

mrousavy:
@efstathiosntonas fixed in https://github.com/mrousavy/react-native-vision-camera/pull/2750 👍

geekabhinavp:
Hey @mrousavy , I was also getting the skia missing error, i installed it manually in the package. Now that error is gone but I am getting stuck on app loader itself on launch, in the logs i am just getting:
```
iOS Bundling complete 5644ms
[appname] VisionCameraProxy: Creating Worklet Context...
[appname] VisionCameraProxy: Worklet Context Created!
```

mrousavy:
Are there any native logs?

geekabhinavp:
These are the complete logs:
```
[UIKitCore] Class CKBrowserSwitcherViewController overrides the -traitCollection getter,
which is not supported. If you're trying to override traits, you must use the
appropriate API.
iOS ./index.js ░░░░░░░░░░░░░░░░  0.1% ( 2/57)[CoreFoundation] AddInstanceForFactory: No factory registered for id <CFUUID
0x6000006c06a0> F8BB1C28-BAE8-11D6-9C31-00039315CD46
iOS Bundling complete 5644ms
[appname] VisionCameraProxy: Creating Worklet Context...
[appname] VisionCameraProxy: Worklet Context Created!
```

PR #2746, opened 2024-04-17
Fix ResolutionSelector+forSize.kt

versusvoid:
## What

This PR fixes resolution selection on android

## Changes

This PR fixes computation of distance between resolutions

## Tested on

POCO M3


mrousavy:
ah nice spot, thank you!

PR #2750, opened 2024-04-18
fix: Make dependencies (`ModuleProxy`) actually optional

mrousavy:
## What

I created the `createModuleProxy` function to generate lazy-, optional-dependencies for us.

This was used for react-native-worklets-core (Frame Processors) and Skia + Ranimated (Skia Frame Processors), but unfortunately it crashed when the dependency was not installed in JS.

This was because we were using try/catch in `createModuleProxy`, and apparently the JS parser is not smart enough to catch require errors thrown in a function passed in.
So we need to try/catch at the actual `require` call to fix this.

Maybe this is a bug in Hermes or Metro or something?

This breaks: 

```ts
const getModule = () => require('bla') // <-- error thrown here already

let module = null
try {
  module = getModule()
} catch (e) {
  console.log('module not available.')
}
```

This works:

```ts
const getModule = () => {
  try {
    return require('bla')
  } catch (e) {
    console.log('module not available.')
    return null
  }
}

let module = getModule()
```

### DIFF
--- a/package/example/ios/Podfile.lock
+++ b/package/example/ios/Podfile.lock
@@ -467,7 +467,7 @@ PODS:
     - RCT-Folly (= 2021.07.22.00)
     - React-Core
   - SocketRocket (0.6.1)
-  - VisionCamera (4.0.0-beta.14):
+  - VisionCamera (4.0.0-beta.15):
     - React
     - React-callinvoker
     - React-Core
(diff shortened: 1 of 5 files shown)


mrousavy:
Crated a metro bug report: https://github.com/facebook/metro/issues/1256

We'll see if this is actually intended or a bug

wcandillon:
thanks I'll update too

Issue #2753, opened 2024-04-18
🐛 [4.0.0-beta.15] - Recording video with useSkiaFrameProcessor freezes my app


Ballonek:
### What's happening?

When I try to record video during frame processing using the new useSkiaFrameProcessor everything works fine, the problem occurs when I want to stop the video recording. At that point my application freezes and in some cases crashes.

<img width="1172" alt="Snímek obrazovky 2024-04-18 v 15 40 16" src="https://github.com/mrousavy/react-native-vision-camera/assets/61647090/c324055e-b537-4406-9803-51f39f689e87">

All of this did work until I updated to beta.15 so i asume that is caused by this update.

I tried to reproduce it on example app but getting `error: Error: Unable to resolve module @babel/runtime/helpers/interopRequireDefault from /Users/XYZ/Git/react-native-vision-camera/package/src/Camera.tsx: @babel/runtime/helpers/interopRequireDefault could not be found within the project.`

### Reproduceable Code

```tsx
const frameProcessor = useSkiaFrameProcessor(frame => {
  'worklet';
  const detectedPose = poseDetection(
    frame,
    frameNumber.current,
  ) as SkeletonPose;
  frame.render();
  if (isRecording.current) {
    if (Object.values(detectedPose).length > 0) {
      const poseArray = getFramePoseData(
        detectedPose,
        frame.width,
        frame.height,
      );

      const pointsPaint = Skia.Paint();
      pointsPaint.setColor(Skia.Color('green'));
      pointsPaint.setAntiAlias(true);
      pointsPaint.setStyle(PaintStyle.Fill);
      for (let i = 0; i < poseArray.length; i++) {
        frame.drawCircle(
          poseArray[i].x * frame.width,
          poseArray[i].y * frame.height,
          10,
          pointsPaint,
        );
      }
    
      frame.restore();
      frameNumber.current++;
    } else {
      frame.render();
    }
  }
}, []);

<ReanimatedCamera
  ref={cameraRef}
  frameProcessor={frameProcessor}
  style={StyleSheet.absoluteFill}
  device={device}
  video={true}
  pixelFormat="rgb"
  isActive
  fps={30}
  format={format}
  preview={false}
/>
```

### Relevant log output

```shell
15:35:41.555: [info] 📸 VisionCamera.startRecording(options:onVideoRecorded:onError:): Starting Video recording...
[...]

PR #2754, opened 2024-04-18
fix: Clean up Skia resources on Worklet context to avoid race condition

mrousavy:
## What

Cleans up Skia context resources (all the `SkSurface`s, and all the `SkImage`s (textures)) on the VisionCamera Frame Processor Worklet Context, instead of the default React JS Context.

Before, we cleaned it up on the JS context, which could introduce a race condition as the Frame Processor might still be executing. This can then lead to a hard-crash when trying to access some things, as the memory has been deleted elsewhere.

So instead, we now "schedule" that cleanup to run on the Worklet Context, which will make sure that it always runs after the Frame Processor.

## Related issues

- Fixes https://github.com/mrousavy/react-native-vision-camera/issues/2753

### DIFF
--- a/package/android/src/main/cpp/frameprocessor/VisionCameraProxy.cpp
+++ b/package/android/src/main/cpp/frameprocessor/VisionCameraProxy.cpp
@@ -42,6 +42,7 @@ std::vector<jsi::PropNameID> VisionCameraProxy::getPropertyNames(jsi::Runtime& r
   result.push_back(jsi::PropNameID::forUtf8(runtime, std::string("setFrameProcessor")));
   result.push_back(jsi::PropNameID::forUtf8(runtime, std::string("removeFrameProcessor")));
   result.push_back(jsi::PropNameID::forUtf8(runtime, std::string("initFrameProcessorPlugin")));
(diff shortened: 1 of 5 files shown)


package/android/src/main/cpp/frameprocessor/VisionCameraProxy.cpp
```diff
@@ -97,6 +96,9 @@ jsi::Value VisionCameraProxy::get(jsi::Runtime& runtime, const jsi::PropNameID&
 
           return this->initFrameProcessorPlugin(runtime, pluginName, options);
         });
+  } else if (name == "workletContext") {
+    std::shared_ptr<RNWorklet::JsiWorkletContext> context = _javaProxy->cthis()->getWorkletContext();
```
efstathiosntonas, line 100: Hi Marc, 

I get this while building beta.16:

`error: use of undeclared identifier 'RNWorklet'`

Screenshot 2024-04-19 at 19 12 41

mrousavy: ah whoops, pushed a fix.

PR #2769, opened 2024-04-21
feat: Rebuild FPS Graph as JS component

mrousavy:
## What

Has the benefit of much more flexibility, and looks the same on both platforms now.


efstathiosntonas:
Hi Marc,

I got  a case where user takes photo from Camera -> navigate to image cropper -> user taps cancel and navigates back to Camera throws `Timer already cancelled`:

<details>
  <summary>Click me</summary>
  
Screenshot_20240422_114052

</details>

maybe convert `timer` to this on `FpsSampleCollector.kt`? 🤔 

```kotlin
  private var timer: Timer? = null

   init {
      timer = Timer("VisionCamera FPS Sample Collector")
   }


    fun start() {
        if (timer == null || timer!!.equals(null)) {
            timer = Timer()
        }
        timer?.schedule(1000, 1000) {
            callback.onAverageFpsChanged(averageFps)
        }
    }

    fun stop() {
        timer?.cancel()
        timer = null
    }

```

after applying the patch above it worked as a charm, don't know if there are any side effects.

mrousavy:
```js
 if (timer == null || timer!!.equals(null))
 ```
what is that lol

but yea ok you're right apparently you cannot resume canceled timers

efstathiosntonas:
haha I thought it was obvious that this is chat gpt 😅 

Screenshot 2024-04-22 at 12 20 09

PR #2788, opened 2024-04-24
CameraX upgrade

SveBo:


mrousavy:
hey - just upgrading from 0.72 to 0.73 without any changes at all is not going to work in the example app.
we might need to migrate using the RN Upgrade Helper.

PR #2811, opened 2024-04-27
feat: Added option to consider ringer mode in Android

imjihun:
## What
When using the takePhoto function in Android, ringerMode is not applied by default, but is applied when a value is received as an option

If isAppliedRingerMode option is set to true in takePhoto function,
When ringerMode is RINGER_MODE_NORMAL, the shutterSound option is applied.
When the ringer mode is RINGER_MODE_SILENT or RINGER MODE VIBRATE, the shutter Sound option is ignored.

### DIFF
--- a/package/android/src/main/java/com/mrousavy/camera/core/CameraSession.kt
+++ b/package/android/src/main/java/com/mrousavy/camera/core/CameraSession.kt
@@ -492,7 +492,7 @@ class CameraSession(private val context: Context, private val callback: Callback
     }
   }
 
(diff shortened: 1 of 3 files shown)


mrousavy:
Hey - thank you for your contribution, this is much appreciated. I always prioritize PRs over issues to discuss such things.

In this case, I think the prop can be a bit confusing. The naming, especially combined with the description, isn't very obvious.

I think that the current behaviour is perfectly fine - it is exactly how it is on iOS. When the phone is on silent, we don't want any sound.
What is your use-case that you need shutter sound anyways? Is that really what users expect to happen?

mrousavy:

PR #2819, opened 2024-04-29
fix: Fix CodeScanner crash

mrousavy:
## What

Fixes an issue where the CodeScanner would crash because it threw an error that didn't go into the `Task<T>`, but instead threw already before creating the `Task<T>`. This is Google's fault, but whatever we can catch it.


mrousavy:
actually I don't know if this changes anything. The errors people were seeing were C++ SIGSEV errors, which are memory access errors and cannot be "caught" with a try/catch.

PR #2840, opened 2024-05-03
feat: Use `worklet(...)` to automatically detect `useFrameProcessor` dependencies

mrousavy:
## What

Uses the new `worklet(..)` function from react-native-worklets-core to automatically detect dependencies for `useFrameProcessor` and `useSkiaFrameProcessor`.

This now means that the second argument (`DependencyList`) for `useFrameProcessor` is no longer required and will be automatically generated:

```ts
const [state, setState] = useState(false)
useFrameProcessor(() => {
  'worklet'
  console.log(state)
}) // <-- no more [...] here!
```

Additionally, `useFrameProcessor` and `useSkiaFrameProcessor` will now throw more helpful error messages when the given function is not a proper worklet.

> [!NOTE]
> This requires react-native-worklets-core 1.3.0 or higher.


mrousavy:
V5

PR #2880, opened 2024-05-15
Fix component did update if statement

nmajumder12:
Currently the if statement is always true since ReadonlyFrameProcesser != FrameProcessor

## What

Fixes if boolean


mrousavy:
Nice catch, thank you!

Issue #893, opened 2022-03-12
✨ Add record / save path


hirbod:
### What feature or enhancement are you suggesting?

I think it would be nice to define where the file should be saved. Currently it's being saved in the /tmp/ folder.
While this is not an issue but more a convenience and speed thing,  (I am using FileSystem.copyAsync from expo to create a copy of it, since I am not allowed to "moveAsync" (permission issue?)

I would love to be able to define a save path, which Expo's FileSystem has access to.

Why I am asking for it? Because it's unreliable when the OS clears the tmp folder. I currently generated 16 GB of video files in my tests and there is no way for me to delete those. I think the end user will eventually delete our app if they can't free up space. 

### What Platforms whould this feature/enhancement affect?

iOS, Android

### Alternatives/Workarounds

There are no workarounds. I tried to delete the files with `FileSystem.deleteAsync` but it always ends up with "file is not deleteable/moveable".

The copies of the files I create though, which live in FileSystems documentDirectory, are move- and deleteable.

I am using a video editor sdk which also has the ability to define the save path and its working pretty good and reliable. 

### Additional information

- [X] I agree to follow this project's [Code of Conduct](https://github.com/mrousavy/react-native-vision-camera/blob/main/CODE_OF_CONDUCT.md)
- [X] I searched for [similar feature requests in this repository](https://github.com/mrousavy/react-native-vision-camera/issues?q=is%3Aopen+is%3Aissue+label%3A%22✨+enhancement%22) and found none.

mrousavy:
Thanks for the feature request, that's a pretty valid enhancement. I'd imagine the user being responsible for creating all of the paths and making sure VisionCamera has permission to create a file there?

hirbod:
Here are the comments the other SDK I am using is making:

```
* The filename for the exported data .
     * The correct filename extension will be automatically added
     * based on the selected export format. It can be an absolute path or file URL or a relative path.
     * If some relative path is chosen it will be created in a temporary system directory and overwritten
     * if the corresponding file already exists. If the value is `null` an new temporary file will be
     * created for every export.
     * @note Please make sure that the provided `filename` is valid for the different devices and that
     * your application has the corresponding access rights to write to the desired location. For Android,
     * you will want to make sure to set this inside one of the directories conforming to scoped storage:
     * - DCIM/
     * - Pictures/
     * 
     * For Videos you can additionally use:
     *  - Movies/
```

And I am supplying something like this:
```
[...]

PR #2900, opened 2024-05-22
feat: Add custom save path

j-jonathan:
## What
This PR adds a feature allowing users to define the output path when taking a photo/snapshot or recording a video.
The idea was presented in https://github.com/mrousavy/react-native-vision-camera/issues/893.

Additionally, this PR fixes an issue on Android where the quality parameter for takeSnapshot was mandatory, resulting in an error [EUNSPECIFIED: quality] if not provided. (https://github.com/mrousavy/react-native-vision-camera/issues/2804)

## Changes
Added a new parameter "path" to specify the output directory for photos, snapshots, and recorded videos.
Updated the relevant functions to handle the custom save path.

## Tested on
iPhone 12, iOS 17.4.1
Huawei ELE-L29, Android 12

## Related issues
- https://github.com/mrousavy/react-native-vision-camera/issues/893
- https://github.com/mrousavy/react-native-vision-camera/issues/2804

### DIFF
--- a/package/android/src/main/java/com/mrousavy/camera/core/CameraError.kt
+++ b/package/android/src/main/java/com/mrousavy/camera/core/CameraError.kt
@@ -208,6 +208,8 @@ class FrameInvalidError :
       "- If you want to use runOnJS, increment it's ref-count: `frame.incrementRefCount()`"
   )
 class InvalidImageTypeError : CameraError("capture", "invalid-image-type", "Captured an Image with an invalid Image type!")
+class InvalidPathError(message: String) :
+  CameraError("capture", "invalid-path", "The given path is invalid! $message")
 
 class CodeTypeNotSupportedError(codeType: String) :
   CameraError(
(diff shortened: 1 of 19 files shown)


mrousavy:
Hi! Thank you so much for this contribution, this is good stuff.
I have a few code-organization/style concerns which I will think about and get back to you soon.

To share what's going on inside my head;
- I am not a big fan of having one function for getting a path or creating a temp file.
- I think maybe we can get the `path` at high-level using a sync native function, and basically always assume the path is non-null and writable on native side. In the JS func for startRecording/takePhoto/takeSnapshot we get a temp path from native if the current path is null (or not writable). This also reduces the amount of logic/branching on the native side and moves it to JS, where code is more flexible in general
- `message` is a bit weird for the error parameter, I'd call it `cause` or something and pass an error I guess? Just a minor nitpick

Don't worry about changes right now, I'll get back to you in a few days about this. But I definitely think this should go into main, this is good nice work 👍

j-jonathan:
Thank you for taking the time to review my PR and for your positive feedback.

I agree that it could be a good idea to have a function for getting a temporary path and managing the path logic directly in the JS code. This would indeed simplify the native side and make the code more flexible overall.

I'll wait for your detailed feedback before making any changes. Looking forward to refining this feature based on your suggestions :)

PR #2913, opened 2024-05-24
Attach the VisionCameraProxy runtime to FrameProcessorPluginHostObject

nmajumder12:
## What

Since Frame Processor Plugin will always be called within the context of the RNVC, shouldn't it be attached to the vision camera runtime instead of the runtime of the initializer.

## Related issues

Potentially related to https://github.com/mrousavy/react-native-vision-camera/issues/2820 and https://github.com/margelo/react-native-worklets-core/issues/173


mrousavy:
Hi - thank you for your PR, unfortunately that is not safe to do.
The Pointer cannot be "captured" like this, and this can now no longer be called from multiple Runtimes.

Issue #2838, opened 2024-05-02
🐛 Camera FPS throttles on low light, and torch doesn't always work


wuguishifu:
### What's happening?

I am using a frame processor on a Pixel 6a/Note 9. I have simplified the frame processor to just not actually perform any calculations. The camera normally runs at 30 FPS according to the graph, but when I cover the camera (with a cloth, my finger, putting it on a table, etc.) the FPS drops to 15. I have tried to disable as much auto stuff as possible on the camera.

Another issue is that the torch does not turn on, even when specified "on" in the camera's props. This happens regardless of the other props. If I edit the file and hot update my app while it's running (in dev mode) it turns on the flashlight. However, if I close and reopen the app, the flashlight once again does not turn on.

These systems worked in v3.9.2 on Android and are currently working in v4 on iOS (iPhone 15pm, iOS 17.x and iPhone 11, iOS 17.x), so I believe these issues are exclusive to Android and v4.

### Reproduceable Code

```tsx
const device = useCameraDevice('back');

const frameProcessor = useFrameProcessor(() => {
  'worklet'
}, []);

const format = useMemo(() => {
  return getCameraFormat(device, [{
    fps: 30
  }]);
}, [device]);

return (
  <Camera
    isActive
    style={StyleSheet.absoluteFill}
    device={device}
    torch='on'
    frameProcessor={frameProcessor}
    enableFpsGraph
    pixelFormat='yuv'
    fps={30}
    format={format}
  />
);
```

### Relevant log output

```shell
N/A
```

### Camera Device

```json
{
  "formats": [],
  "sensorOrientation": "landscape-left",
  "hardwareLevel": "full",
  "maxZoom": 8,
  "minZoom": 1,
  "maxExposure": 20,
  "supportsLowLightBoost": false,
  "neutralZoom": 1,
  "physicalDevices": [
    "wide-angle-camera"
  ],
  "supportsFocus": true,
  "supportsRawCapture": false,
  "isMultiCam": false,
  "minFocusDistance": 10,
  "minExposure": -20,
  "name": "0 (BACK) androidx.camera.camera2",
  "hasFlash": true,
  "hasTorch": true,
  "position": "back",
  "id": "0"
}
```

### Device

Pixel 6A (Android 14), Note 9 (Android 10)

### VisionCamera Version

4.0.1

### Can you reproduce this issue in the VisionCamera Example app?

Yes, I can reproduce the same issue in the Example app here

### Additional information

[...]

PR #2925, opened 2024-05-31
feat: Set `fps` for Frame Processor on Android

mrousavy:
## What

Actually set the target FPS on the ImageAnalysis usecase.

Inspired by @wuguishifu's comment in https://github.com/mrousavy/react-native-vision-camera/issues/2838


wuguishifu:
This is awesome, thanks for taking the time to get this working!!

PR #2931, opened 2024-06-03
fix: Fix Skia Frame Processor transformation matrix

mrousavy:
## What

Fixes the rotation/orientation of the Frame in a Skia Frame Processor.

When we rotate the frame, the original origin (0, 0) is no longer at the top left corner, so we also need to post-translate by the width- and height-difference (normalized to the new aspect ratio)

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>

<tr>
<td></td>
<td></td>
</tr>
</table>

## Tested on

- Samsung S21

## Related issues

- Fixes https://github.com/mrousavy/react-native-vision-camera/issues/2762

PR #2932, opened 2024-06-03
feat: Implement Orientation

### LINKED ISSUE #1891: ✨ Implement Orientation ($3,000)

mrousavy:
## What

Implements orientation for VisionCamera.

1. A new prop is introduced (`outputOrientation`), which specifies how VisionCamera should treat orientation:
    - `device`; use physical device orientation, even if screen is locked
    - `preview`; use preview orientation, respects screen-lock
    - `portrait`, `landscape-left`, `landscape-right`, `portrait-upside-down`; locked to a specific orientation
2. The current `orientation` prop has been removed to avoid confusion. (it never really worked anyways)
3. Adds documentation for **Orientation** to explain how it works
4. Make `Frame.orientation` actually reflect it's relative orientation.
5. Stops rotating `CMSampleBuffer`s on iOS. This makes the video and frame processor pipeline much more lightweight and use less energy, as it now no longer physically rotates pixels in a buffer.
    - For Videos, rotation will be passed to the AVAssetWriter (EXIF flag)
    - For Frame Processors, `Frame.orientation` now represents it's rotation relative to the target rotation (like on Android)
6. In a previous PR, Skia Frame Processors now properly rotate their transform matrix to support orientation (https://github.com/mrousavy/react-native-vision-camera/pull/2931)

## Related issues

- Resolves https://github.com/mrousavy/react-native-vision-camera/issues/1891


rodrigobertolottizinsp:
Hi! Thanks for the updates! Looks great so far. I was able to rotate preview and capture pictures perfectly in iOS. On Android, images seems to be correctly orientated but preview does not rotate. I have outputOrientation="device", testing on a Google Pixel 2. if I explicitly set an orientation it works fine.

mrousavy:
Does the screen/view itself rotate?

jkaufman:
I appreciate that you're using `UIDevice` and `SensorManager` APIs, specifically, for device orientation. These rely on accelerometer, alone.

rodrigobertolottizinsp:
> Does the screen/view itself rotate?

https://github.com/mrousavy/react-native-vision-camera/assets/112571554/785c2e44-5d99-49c0-b502-3506c0c53591

Yes it does, but it's not correctly orientated.

docs/docs/guides/ORIENTATION.mdx
```diff
@@ -0,0 +1,110 @@
...
+:::info
+This will always happen automatically according to the screen's rotation.
+:::
+
+### The `outputOrientation` prop
+
+The orientation in which photos and videos are captured can be adjusted via the [`outputOrientation`](/docs/api/interfaces/CameraProps#outputorientation) property:
+
+```tsx
+<Camera {...props} outputOrientation="device" />
```
kirillzyusko, line 52: Would it make sense to combine these two properties into a single object, and pass it as config?

So it will look as:

```tsx
const orientation = {
  sensor: "...",
  output: "...",
};

// ...

<Camera orientation={orientation} />
```

Asking from curiosity 🙂

mrousavy: Hey Kiryll - not really - sensor orientation is something fixed by the Camera (it is relative to the phone's natural orientation), and the `outputOrientation` just specifies where to rotate towards (either follow `device`/`preview` orientation, or fixed to something like `landscape-left`).

Hope that makes sense :)

kirillzyusko: @mrousavy ah, yeah, makes sense then 🙂

mrousavy:
@rodrigobertolottizinsp huh - I just noticed that this is because of the `frameProcessor` - if you remove it, it rotates fine... It might be a CameraX bug?

ISnowFoxI:
Tested for our use case, which does not involve `frame processors`. Orientation is working nicely for both Android and iOS, on my devices. The `Selfie` and `back` cameras show the correct orientation. 

Thank you all for the great work.

The example app needs a couple of updates, so it can show the video recordings correctly. I'll update the codebase for the example app if I can make time for it. But first I need to get your permission to edit the example app.  

The list below shows what I'd change on the example app 
- Lock the orientation of the camera preview to `portrait`, as this is how most native camera apps do it UI-wise. This would be done with the `options` prop for the preview screen in the navigation logic. 
- Fix video playback so it shows the videos in a `contained` style and fixes overflow issue on `landscape` videos. This would be done by editing `resizeMode` for `react-native-video`.

mrousavy:
I'll build a new example app the coming days, don't worry bout it

xulihang:
> Stops rotating CMSampleBuffers on iOS. This makes the video and frame processor pipeline much more lightweight and use less energy, as it now no longer physically rotates pixels in a buffer.

This seems like a breaking change for frame processors. For text recognition, the frame processor has to rotate the image to match the preview in portrait mode if it does not accept passing the orientation.

Maybe adding an option to enable rotation is a good idea to make old frame processors compatible with the new version.

mrousavy:
Hm, I am not sure - I think for portrait nothing changed, but whereas frame processors previously didnt work when the phone is rotated, they now work.

Not entirely sure though, need to test it...

laxminarayana-capgemini:
Hi @mrousavy, we trying to record the video in Landscape, but the output video is in Portrait. Whatever the orientation we are recording it is in Portrait only. We are running react-native-vision-camera in Android application, but it is not working. Hereby I have attached a sample code. Kindly help us with a solution.
`<Camera
            style={StyleSheet.absoluteFill}
            device={device}
            isActive={isAppForeground}
            video={true}
            audio={true}
            ref={camera}
            format={format}
            exposure={0}
            zoom={0}
            orientation='landscape-left'
          />`

pke:
thanks @mrousavy! It seems `onCodeScanned` of `useCodeScanner` has not had its `frame.bounds` fixed with this PR? The bounds should be aligned in React space always at 0,0 being top left of the detected code.

jkaufman:
@laxminarayana-capgemini The name of the orientation prop changed with this PR.

laxminarayana-capgemini:
@mrousavy and @jkaufman. thanks, the issue is resolved.

PR #2960, opened 2024-06-10
feat: Expose `Frame.matrix`

mrousavy:
## What

Exposes camera intrinsic frame matrix ([`kCMSampleBufferAttachmentKey_CameraIntrinsicMatrix`](https://developer.apple.com/documentation/coremedia/kcmsamplebufferattachmentkey_cameraintrinsicmatrix?language=objc)) to JS via a 3x3 Matrix.


mrousavy:
Hm, actually this is not what I wanted. I thought this would be a 3x3 matrix representing rotations, scalings and transformations to apply to convert between coordinate systems (e.g. for tap to focus), but apparently that's wrong.

PR #2962, opened 2024-06-11
fix: onUIRotationChanged uiRotation

j-jonathan:
## What

This PR fixes the incorrect UI rotation returned by onUIRotationChanged on certain devices.

For example, on my ELE-L29, all rotations were inverted.

Indeed, when the previewOrientation and outputOrientation are identical, the uiRotation should be 0. This was not the case with the current formula.

## Changes

This PR changes the formula used to calculate uiRotation.

## Tested on

- iPhone 12, iOS 17.4.1
- Huawei ELE-L29, Android 12

### DIFF
--- a/package/src/RotationHelper.ts
+++ b/package/src/RotationHelper.ts
@@ -30,6 +30,6 @@ export class RotationHelper {
   get uiRotation(): number {
     const previewDegrees = orientationToDegrees(this.previewOrientation)
     const outputDegrees = orientationToDegrees(this.outputOrientation)
-    return ((previewDegrees + outputDegrees) % 360) - 180
+    return (outputDegrees - previewDegrees) % 360
   }
 }


mrousavy:
Can you confirm that this still looks correct on an iPhone in every rotation? I remember having some problems with that formula..

j-jonathan:
I've just tested the modification (with app orientation locked & unlocked) on :
- iPhone 12, iOS 17.4.1
- iPad Air 2, iOS 15.8.2 (everything was reversed without the fix on this device)
- Huawei ELE-L29, Android 12 (everything was reversed without the fix on this device)
- Samsung A22 5G, Android 13 (everything was reversed without the fix on this device)

And everything seems to be ok!

mrousavy:
I remember why I did that - it was because sometimes it returns -270, and that looks weird when animating as it should just be +90 instead. I'll add a fix after that PR, thank you for your contribution LGTM! ❤️

PR #2964, opened 2024-06-11
feat: Android preview aspect ratio

j-jonathan:
## What

Currently on iOS, the previewer ratio matches the video capture ratio (or the photo ratio if video mode is disabled).
This ensures that the preview corresponds to the captured result.

The idea of this pull request is to have the same behavior on Android.

## Changes

The previewer uses the resolution with the aspect ratio closest to the video capture resolution (or photo resolution if video mode is disabled).

Maybe someone has a better idea, such as adding the preview resolution to the format (but this would significantly increase the number of formats) or adding a new prop to define the previewer resolution.

## Tested on

- iPhone 12, iOS 17.4.1
- Huawei ELE-L29, Android 12

### DIFF
--- a/package/android/src/main/java/com/mrousavy/camera/core/CameraConfiguration.kt
+++ b/package/android/src/main/java/com/mrousavy/camera/core/CameraConfiguration.kt
@@ -51,6 +51,22 @@ data class CameraConfiguration(
   data class Audio(val nothing: Unit)
   data class Preview(val surfaceProvider: SurfaceProvider)
 
+  val targetPreviewAspectRatio: Float?
+    get() {
+      val format = format ?: return null
+      val video = video as? Output.Enabled<Video>
+      val photo = photo as? Output.Enabled<Photo>
+      return if (video != null) {
+        // Video capture is enabled, use video aspect ratio
+        format.videoWidth.toFloat() / format.videoHeight.toFloat()
+      } else if (photo != null) {
+        // Photo capture is enabled, use photo aspect ratio
+        format.photoWidth.toFloat() / format.photoHeight.toFloat()
+      } else {
+        null
+      }
+    }
+
   @Suppress("EqualsOrHashCode")
   sealed class Output<T> {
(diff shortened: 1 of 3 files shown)


mrousavy:
Hi - thanks for your PR, maybe I am missing something but doesn't the PreviewView automatically match the aspect ratio of other outputs?

mrousavy:
Also, I think the Preview View will automatically find a size that matches the aspect ratio of other outputs, is that not the case? 🤔

j-jonathan:
No, unfortunately the preview does not match the aspect ratio of the other outputs, which means that depending on the device, there can be a significant difference between the photo/video captured and what is seen in the preview at the moment of capture/recording.

mrousavy:
I pushed a commit that cleans up the code a bit, `CameraSession+Configuration` is already way too big 😅 

Otherwise that's cool - I am a bit hesitant about merging this because it differs from how iOS works (and how the formats documentation works).

On iOS, the Preview View size (and aspect ratio) is always bound to the `format`'s `videoWidth` / `videoHeight`.
On Android, this is now different; it only uses the `format`'s video size if `video` is enabled, otherwise it uses the `format`'s photo size.

I think in practise it should be fine, but it is different behaviour now between iOS and Android.

I have code in ShadowLens that finds a format with a video size where the aspect ratio matches the screen aspect ratio, and this also has the assumption that the preview view uses that aspect ratio. I have `video` enabled so it's fine, but for people who don't have `video` enabled it's gonna be a different case..

Maybe I will change this in the future to always use the `format`'s video aspect ratio so it stays consistent.

mrousavy:
Anyways - thank you for your contribution, LGTM!!! :)

j-jonathan:
Thanks for the review and the modifications!

It's true that I adapted the code to my test case by taking the photo format when video mode is deactivated.
I will check tomorrow if all the ratios are available in video on the various Android devices I have. If so, we could have the same behavior as on iOS.

PR #2970, opened 2024-06-12
fix: auto focus/exposure by default

j-jonathan:
## What

On some iOS devices (e.g. on my iPod) the auto focus/exposure is not enabled by default. The focus/exposure is done when the camera is loaded and does not change afterwards.

## Changes

Enable the auto focus/exposure when a format is loaded.

## Tested on

- iPhone 12, iOS 17.4.1
- iPad Air 2, iOS 15.8.2
- iPod, iOS 12.5.7

### DIFF
--- a/package/ios/Core/CameraSession+Configuration.swift
+++ b/package/ios/Core/CameraSession+Configuration.swift
@@ -195,6 +195,14 @@ extension CameraSession {
     // Set new device Format
     device.activeFormat = format
 
+    // Set auto focus/exposure
+    if device.isFocusModeSupported(.continuousAutoFocus) {
+      device.focusMode = .continuousAutoFocus
+    }
+    if device.isExposureModeSupported(.continuousAutoExposure) {
+      device.exposureMode = .continuousAutoExposure
+    }
+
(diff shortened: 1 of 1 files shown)


mrousavy:
oh wow interesting! is that really the case? I thought it was always continuous by default...

mrousavy:
Does that need to be done in `configureFormat`? What if no Format is selected?

j-jonathan:
Yeah :( At least that's the case on my iPod.

And once, I had the case where the continuous autofocus/autoexposure (i don't remember which one) was not available, so it's important to test first if it's available otherwise it crashs.

> Does that need to be done in `configureFormat`? What if no Format is selected?

Yeah, i don't know if i made the fix at the right place ^^

mrousavy:
I just pushed another commit - can you test if this still works fine on your iPod?

j-jonathan:
I will test that tomorrow :)
Thank you!

j-jonathan:
I just tested on my iPod and it works perfectly fine 😊

mrousavy:
nice!

rsainiWin:
@mrousavy Still not working for me Ipad Pro 11inch 2nd Gen

PR #2972, opened 2024-06-12
fix: orientation on iOS

j-jonathan:
## What

Upon testing the new orientation implementation, I found some issues on iOS.

On iOS 17:
- onPreviewOrientationChanged and onOutputOrientationChanged do not correctly return the orientation. For example, when the phone is in portrait mode, the returned mode is "landscape-left".

On iOS < 17:
- The video orientation is not good. The recorded video is not rotated correctly (it doesn't use the camera sensor orientation).
- When the orientation is locked on the phone, the output orientation is not updated (e.g. onOutputOrientationChanged / onUIRotationChanged).

Upon analyzing the code, I realized that depending on the iOS version, we use videoRotationAngleForHorizonLevelPreview/videoRotationAngleForHorizonLevelCapture and on the other side UIDevice.current.orientation/UIApplication.shared.statusBarOrientation.
videoRotationAngleForHorizonLevelPreview/videoRotationAngleForHorizonLevelCapture already takes into account the camera orientation, whereas the other solution takes into account the phone orientation.

Additionally, when portrait orientation lock is on, UIDevice.current.orientation always returns .up. Therefore, I had to modify the LegacyCameraOrientationCoordinator to function correctly even when the orientation is locked (to achieve the behavior described in the documentation and the same as with iOS 17).

## Changes

- Fix preview / output orientation (same on every iOS version)
- Usage of the accelerometer in LegacyCameraOrientationCoordinator to support locked orientation.

## Tested on

- iPhone 12, iOS 17.4.1
- iPad Air 2, iOS 15.8.2
- iPod, iOS 12.5.7


mrousavy:
Hi - thank you for your PR! Much appreciated.

I will review this asap in the coming days. Thanks!

mrousavy:
> onPreviewOrientationChanged and onOutputOrientationChanged do not correctly return the orientation. For example, when the phone is in portrait mode, the returned mode is "landscape-left".

That is intended though, no? It is relative to the sensor orientation, so if the sensor is `landscape-left` this will return a landscape left orientation..

I am not sure if I am missing something, do you maybe have screenshots or graphics?

package/ios/Core/CameraSession+Orientation.swift
```diff
@@ -63,10 +66,14 @@ extension CameraSession: OrientationManagerDelegate {
...
 
+    guard let device = videoDeviceInput?.device else {
+      return
+    }
+
     // update the orientation for each preview layer that is connected to this capture session
     let previewConnections = captureSession.connections.filter { $0.videoPreviewLayer != nil }
     for connection in previewConnections {
-      connection.orientation = previewOrientation
+      connection.setOrientation(newOrientation: previewOrientation, device: device)
```
mrousavy: I don't think this is the right way to implement this.

I want `orientation` to be a simple property, not a setter/getter function that takes a `device`.

If we do any operations on `device` inside `setOrientation`, why do we not do that beforehand (e.g. in `OrientationManager` when we create `previewOrientation`)?

This is hard to read, so I think this should be the `OrientationManager`'s responsibility.

Can you refactor it that way?

Thank you!

package/ios/Core/CameraSession.swift
```diff
@@ -266,7 +266,10 @@ class CameraSession: NSObject, AVCaptureVideoDataOutputSampleBufferDelegate, AVC
   public final func captureOutput(_ captureOutput: AVCaptureOutput, didOutput sampleBuffer: CMSampleBuffer, from connection: AVCaptureConnection) {
     switch captureOutput {
     case is AVCaptureVideoDataOutput:
-      onVideoFrame(sampleBuffer: sampleBuffer, orientation: connection.orientation)
+      guard let device = videoDeviceInput?.device else {
+        return
+      }
+      onVideoFrame(sampleBuffer: sampleBuffer, orientation: connection.getOrientation(device: device))
```
mrousavy: Same story here - I think `connection.orientation` should not need a device. Either we do the rotation here (`let relativeOrientation = connection.orientation.rotatedBy(...)`), or we just pass in `connection.orientation` directly.

I don't really get this part either - why can we not just pass `connection.orientation` here? Isn't that what you want in a Frame Processor? cc @thomas-coldwell

package/ios/Core/Extensions/AVCaptureConnection+orientation.swift
```diff
@@ -9,27 +9,24 @@ import AVFoundation
...
-   Gets or sets the target orientation of the video connection.
-   */
-  var orientation: Orientation {
-    get {
-      if #available(iOS 17.0, *) {
-        return Orientation(degrees: videoRotationAngle)
-      } else {
-        return Orientation(videoOrientation: videoOrientation)
-      }
+  func getOrientation(device: AVCaptureDevice) -> Orientation {
```
mrousavy: As I said before; I'd prefer a property (getter + setter) and remove the `device` logic from here. I don't think it belongs in here. All `sensorOrientation` logic should be handled before this.

package/ios/Core/Extensions/AVCaptureOutput+orientation.swift
```diff
@@ -9,20 +9,16 @@ import AVFoundation
...
-  /**
-   Gets or sets the target orientation of the video output.
-   */
-  var orientation: Orientation {
-    get {
-      guard let connection = connections.first else {
-        return .portrait
-      }
-      return connection.orientation
+  func getOrientation(device: AVCaptureDevice) -> Orientation {
```
mrousavy: As I said before; I'd prefer a property (getter + setter) and remove the `device` logic from here. I don't think it belongs in here. All `sensorOrientation` logic should be handled before this.

package/ios/Core/Utils/LegacyCameraOrientationCoordinator.swift
```diff
@@ -6,41 +6,78 @@
 //
 
 import AVFoundation
-import Foundation
-import UIKit
```
mrousavy: Let's keep those imports. `UIKit` -> `UIDevice`, `Foundation` -> anything else we might use

package/ios/Core/Utils/LegacyCameraOrientationCoordinator.swift
```diff
@@ -6,41 +6,78 @@
...
+        let zNorm = abs(acceleration.z)
+
+        // If the z-axis is greater than the other axes, the orientation does not change
+        if zNorm > xNorm && zNorm > yNorm {
+          return
+        }
+
+        let orientation: UIDeviceOrientation
+        if xNorm > yNorm {
+          orientation = acceleration.x > 0 ? .landscapeRight : .landscapeLeft
```
mrousavy, line 63: Shouldn't we use `45` here? Isn't `0` very sensitive?

package/ios/Core/Utils/LegacyCameraOrientationCoordinator.swift
```diff
@@ -6,41 +6,78 @@
...
+        } else {
+          orientation = acceleration.y > 0 ? .portraitUpsideDown : .portrait
```
mrousavy, line 65: Same here, `45` or `0`? Is that degrees? Radians? Maybe add a few comments if necessary. I'd call it `degreesX` instead of `xNorm` if it is degrees please.

package/ios/Core/Utils/ModernCameraOrientationCoordinator.swift
```diff
@@ -14,18 +14,20 @@ class ModernCameraOrientationCoordinator: CameraOrientationCoordinator {
   private weak var delegate: CameraOrientationCoordinatorDelegate?
   private var previewObserver: NSKeyValueObservation?
   private var outputObserver: NSKeyValueObservation?
+  private var device: AVCaptureDevice?
 
   var previewOrientation: Orientation {
-    let degrees = rotationCoordinator.videoRotationAngleForHorizonLevelPreview
+    let degrees = (rotationCoordinator.videoRotationAngleForHorizonLevelPreview - getSensorOrientation().degrees + 360).truncatingRemainder(dividingBy: 360)
```
mrousavy: Way too much logic in one line, this is impossible to debug. Please split it across multiple lines and create temporary variables in between with good naming :)

package/ios/Core/Utils/ModernCameraOrientationCoordinator.swift
```diff
@@ -14,18 +14,20 @@ class ModernCameraOrientationCoordinator: CameraOrientationCoordinator {
   private weak var delegate: CameraOrientationCoordinatorDelegate?
   private var previewObserver: NSKeyValueObservation?
   private var outputObserver: NSKeyValueObservation?
+  private var device: AVCaptureDevice?
```
mrousavy:
```suggestion
  private let device: AVCaptureDevice
```

Device cannot be nil here.

package/ios/Core/Utils/ModernCameraOrientationCoordinator.swift
```diff
@@ -46,4 +48,14 @@ class ModernCameraOrientationCoordinator: CameraOrientationCoordinator {
...
+
+  private func getSensorOrientation() -> Orientation {
+    let sensorOrientation: Orientation
+    if let device = device {
+      sensorOrientation = device.sensorOrientation
+    } else {
+      sensorOrientation = Orientation(degrees: 0)
+    }
+    return sensorOrientation
+  }
```
mrousavy: This func is unnecessary since we can just use `device.sensorOrientation` directly - it cannot be nil.

mrousavy (changes requested):
a few thoughts :)

j-jonathan:
I don't know what's the best way to look at it.

Currently, in "portrait" mode (let's say you hold the phone normally in your hands) :
- on a device with iOS 17, output/preview will be "landscape-left"
- on a device with an older version of iOS, output/preview will be "portrait"

Given that the phone is in portrait mode, I think it's more appropriate to set it to portrait.
If, for example, we read the previewOrientation value, we'd like it to be the phone's orientation, wouldn't we? Why should the orientation of the photo sensor be taken into account in the value returned?

Finally, whatever decision is made, I think it's preferable to have the same value for the same orientation regardless of the iOS version.

package/ios/Core/CameraSession.swift
j-jonathan: Because orientation is not treated in the same way on iOS 17 as it was before.
On one of them, you have to take sensor orientation into account.

I don't really know how to solve this problem easily while still using device.sensorOrientation (because I could just add a +90 when using videoRotationAngleForHorizonLevelPreview/videoRotationAngleForHorizonLevelCapture, but that wouldn't be very clean). 

For example, in the orientation setter:
videoRotationAngle (iOS 17) and videoOrientation (iOS < 17) require for the same result a different orientation (difference of the sensor orientation)

mrousavy:
> I think it's preferable to have the same value for the same orientation regardless of the iOS version.

Absolutely! Agreed.

> Given that the phone is in portrait mode, I think it's more appropriate to set it to portrait.

I get what you're saying, but that's why I have a `onUIRotationChanged` event - this is what you actually want to rotate your UI elements by. The other events (`onPreviewOrientationChanged` and `onOutputOrientationChanged`) are kinda low-level events, and they should hold the exact value that I will internally also use and set on the `AVCaptureConnection`.

If I set `.landscapeLeft` on the `AVCaptureConnection` and it is in the proper orientation then, then I want to also send `landscape-left` to the user. It's the actual value that it uses. No need to change that once again.

So in short; `onUIRotationChanged` is for the user to actually high-level rotate UI, and the other two events are what is internally actually being used. We can get the difference in the orientations to figure out UI rotation, no matter if they are both landscapeLeft or both portrait- UI rotation will be `0`.

mrousavy:
Btw what's your handle on Discord? Are you on our community discord? Some thing can be discussed quicker over chat, and I also want to give you the contributor role now :)

package/ios/Core/CameraSession.swift
mrousavy: The Modern Rotation thingy is what is actually the truth - if it says `landscapeLeft` when you hold your phone in portrait, then we apply landscape left. Internally it knows what the Camera's sensor orientation is.

In the Legacy Rotation thingy we should mimic the Modern one, so also take sensor orientation into account. I think this is the place where we should handle that then.
Same for mirroring (`if device.position == .front`), flip it 180.

j-jonathan:
Yes, you're right, so it makes more sense to use the true orientation (the one that takes into account the sensor's orientation).
As long as the rotations are the same regardless of the iOS version, I'm fine with it ^^

I'll change the code to change that and apply your comment!

Is it acceptable for you to hard-code the +90 degrees in package/ios/Core/Extensions/AVCaptureConnection+orientation.swift for the "legacy" mode (iOS < 17)? (otherwise, I'll think of something else)

> Btw what's your handle on Discord? Are you on our community discord? Some thing can be discussed quicker over chat, and I also want to give you the contributor role now :)

I sent you a dm ;)

j-jonathan:
**About the accelerometer**

The values reported by the accelerometers are measured in increments of the gravitational acceleration, with the value 1.0 representing an acceleration of 9.8 meters per second (per second) in the given direction.

image

(Source: https://developer.apple.com/documentation/coremotion/getting_raw_accelerometer_events)

The acceleration measured by a phone’s accelerometer includes gravity. This allows us to determine the phone’s orientation.

Left:
X = -1
Y = 0

Right:
X = 1
Y = 0

Portrait
X = 0
Y = -1

Upside down
X = 0
Y = 1

So we can take the larger absolute value between the x and y axes to determine if the phone is vertical or horizontal, and then check if the value is positive or negative.
The value on the Z-axis is used to detect when the phone is parallel to the ground (in either direction). In that case, we don't want to rotate the camera.

package/ios/Core/Utils/LegacyCameraOrientationCoordinator.swift
j-jonathan: see my comment on the usage of the accelerometer ;)

package/ios/Core/Utils/LegacyCameraOrientationCoordinator.swift
j-jonathan: see my comment on the usage of the accelerometer ;)

j-jonathan:
I updated my pull request to include the various changes and to fix other issues I hadn’t noticed yesterday.

I tested on different iOS devices (iOS 17 or earlier), with orientation locked or not, using the back and front cameras to take photos and videos.

imanderson:
I had an issue, but only with useSkiaFrameProcessor, where the orientation was on "landscape-right" but I was holding the phone in "portrait". Now with this fix the camera is displayed correctly, but the orientation string is "landscape-right" 😄 

Im on iOS 17.5.1

imanderson:
This is what I get inside useFrameProcessor / useSkiaFrameProcessor. Is this intended? 

PS: the real image orientation is correct though

IMG_1102
IMG_1100
IMG_1101
IMG_1103

imanderson:
This is the frame.orientation inside useFrameProcessor / useSkiaFrameProcessor. Should it be like this?

IMG_1102
IMG_1100
IMG_1101
IMG_1103

j-jonathan:
Hello @imanderson,
This is the expected orientation because the preview/output orientations take into account the orientation of the camera sensor.
If you want to have the "real" orientation of the phone, you need to use the uiOrientation (if the orientation is locked in portrait mode)

imanderson:
> Hello @imanderson, This is the expected orientation because it takes into account the orientation of the camera sensor. If you want to have the orientation of the phone, you need to use the uiOrientation

Ah ok, understood. then this PR fixed my issues too. Thanks!

danielshamburger:
This looks like working fix for an issue I am also having on a project. If it could get merged in would be hugely appreciated. Thanks!

mrousavy:
Hey - I took another stab at this to implement this in a clean and effective way. Can you check out https://github.com/mrousavy/react-native-vision-camera/pull/3039 to see if that works for you and if it solves the issues you encountered?

PR #3010, opened 2024-06-19
fix: iOS FPS

j-jonathan:
## What

Currently, the FPS of a video recorded by the vision camera library on iOS are not consistent.
The FPS vary from one video to another even when the “fps” prop is used.

Examples:
image
image

## Changes

Modification of recorder parameters to take into account the desired FPS.

## Tested on

- iPhone 12, iOS 17

### DIFF
--- a/package/ios/Core/CameraSession+Video.swift
+++ b/package/ios/Core/CameraSession+Video.swift
@@ -121,7 +121,7 @@ extension CameraSession {
 
         // Init Video
         let videoSettings = try videoOutput.recommendedVideoSettings(forOptions: options)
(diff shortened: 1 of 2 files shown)


mrousavy:
Hmm, as per documentation the asset writer will choose an appropriate value automatically:

> The default value is 0, which indicates that the input chooses an appropriate value, if applicable. 

Do we really want to overwrite this?

j-jonathan:
After some research, I found that phones don't necessarily capture video at a constant bitrate, so it's a "normal" behaviour.
So, this PR is not necessary :)

PR #3020, opened 2024-06-25
fix: Android ResolutionSelector.Builder.forSize

j-jonathan:
## What

This PR fixes an issue with the resolution selection logic where two formats with the same difference metric resulted in selecting the incorrect resolution.

I encountered a problem with a phone that supports the following two formats:
- "photoHeight": 2592, "photoWidth": 4608  --> (2592*4608 = 11943936)
- "photoHeight": 3456, "photoWidth": 3456  --> (3456*3456 = 11943936)

When the first format was selected, the resulting picture was taken with the resolution of 3456 x 3456 instead of the intended 2592 x 4608.

## Changes

This PR updates the difference function in the resolution selection code to consider both width and height differences in addition to the area difference.

## Tested on

- Samsung SM-G525F

### DIFF
--- a/package/android/src/main/java/com/mrousavy/camera/core/extensions/ResolutionSelector+forSize.kt
+++ b/package/android/src/main/java/com/mrousavy/camera/core/extensions/ResolutionSelector+forSize.kt
@@ -4,8 +4,7 @@ import android.util.Size
 import androidx.camera.core.resolutionselector.ResolutionSelector
 import kotlin.math.abs
 
-private fun difference(left: Size, right: Size): Int =
-  abs(left.width * left.height - right.width * right.height) + abs(left.width - right.width) + abs(left.height - right.height)
+private fun difference(left: Size, right: Size): Int = abs(left.width - right.width) + abs(left.height - right.height)
 
 /**
  * Gets a [ResolutionSelector] that finds a resolution closest to the given target size.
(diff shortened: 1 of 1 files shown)


mrousavy:
Nice!

mrousavy:
> to consider both width and height differences in addition to the area difference.

Why not just width and height differences? Why do we need the total area difference as well?

j-jonathan:
> Why not just width and height differences? Why do we need the total area difference as well?

You’re right; using just the width and height differences simplifies the logic and is sufficient for selecting the closest resolution.

I have made the change accordingly.

mrousavy:
> ... and is sufficient for selecting the closest resolution.

Okay are you 100% sure with this code we now choose the right format? I didn't test this, so I am fully relying on your testing here to make sure this **always** chooses the expected format.

j-jonathan:
> Okay are you 100% sure with this code we now choose the right format? I didn't test this, so I am fully relying on your testing here to make sure this **always** chooses the expected format.

I have tested this on 2 or 3 phones, and everything works fine.

The resolution contained in the format is supposed to exist, so the correct resolution will be chosen. However, sometimes the selected format cannot be used (I found a case where if you set the FPS to the value of format.maxFps, certain high resolutions are not accepted). The chosen resolution will then be the next valid one in the list sorted by forSize.

The change to just using the difference between abs(left.width - right.width) + abs(left.height - right.height) could alter the order slightly compared to using the area difference. However, there isn’t one solution that is definitively better than the other; the behavior will potentially change depending on the available resolutions.

The only potential issue is that the width and height might be inverted. If you think it could happen , I can add a min/max check to ensure that we are correctly comparing width and height.
Something like:
```kt
private fun difference(left: Size, right: Size): Int {
    // Use min and max to handle cases where width and height might be inverted
    val leftMin = min(left.width, left.height)
    val leftMax = max(left.width, left.height)
    val rightMin = min(right.width, right.height)
    val rightMax = max(right.width, right.height)
    return abs(leftMin - rightMin) + abs(leftMax - rightMax)
}
```

PR #3021, opened 2024-06-25
fix: unfocus on iOS

j-jonathan:
## What

Currently, the focus is never reset on iOS (even when the camera is restarted).

This pull request aims to fix this issue.

## Changes

- Fix for resetting autofocus/autoexposure

Info:
focusPointOfInterest / exposurePointOfInterest should be set to CGPoint(x: 0.5, y: 0.5) to be reset.

## Tested on

- iPhone 12, iOS 17

### DIFF
--- a/package/ios/Core/CameraSession+Configuration.swift
+++ b/package/ios/Core/CameraSession+Configuration.swift
@@ -272,9 +272,15 @@ extension CameraSession {
 
     // Configure auto-focus
     if device.isFocusModeSupported(.continuousAutoFocus) {
+      if device.isFocusPointOfInterestSupported {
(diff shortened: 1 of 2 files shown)


mrousavy:
Nice!

mrousavy:
Can you 100% confirm that before this change it didn't reset focus properly, and after this change it now does?

If that's the case, I'm confident in merging it - thanks! :)

j-jonathan:
I tested it on my iPhone 12 and i can 100% confirm you that it works on it (and without the fix, the focus/autoexposure is never reset even if i close / reopen the screen with the camera component).

(i tested both the subjectAreaDidChange event and the camera init changes)

mrousavy:
perfect, thank you jonathan!

PR #3026, opened 2024-06-25
fix: Fix `ResolutionSelector.forSize` not respecting aspect ratio

mrousavy:
## What

When passing a format, make sure that if we need to fall back to another format we use a similar aspect ratio.

## Related issues

- Overrides https://github.com/mrousavy/react-native-vision-camera/pull/3020


j-jonathan:
Hello,
I tested it on 4 different phones/tablets (with different Android versions, etc...) and everything is ok on my side :)

mrousavy:
Awesome!

PR #3039, opened 2024-07-02
fix: Fix Orientation edge-cases on iOS

mrousavy:
## What

On iOS, I built two implementations - one for iOS 17+, and one for below iOS 17. 

The iOS 17+ one was a bit of a black-box since it internally handles sensor orientation, mirror modes, and preview settings, and the iOS 17 and below one was a bit dumber and didn't handle those cases.

So now I nuked both of those and rewrote it into a single `OrientationManager` which does all of it manually - it handles sensor orientation, uses device orientation (CMMotionManager) for updates, and should work the same on all iOS versions.

## Related issues

- Closes https://github.com/mrousavy/react-native-vision-camera/pull/2972


xulihang:
On iOS 15, iPhone SE, the preview orientation is rotated in v4.4.0. It works fine using v4.3.2

fleuverouge:
> On iOS 15, iPhone SE, the preview orientation is rotated in v4.4.0. It works fine using v4.3.2

I got the same issue on iPhone 7 plus iOS 15 and iphone 6S iOS 13.
The orientation is correct on iPhone 13 iOS 17.

mrousavy:
What does rotated mean? Please share screenshots and logs.

PR #3046, opened 2024-07-03
feat: Dynamically determine `sensorOrientation` based on default output-connection-orientation

mrousavy:
## What

Previously, we just assumed a default sensor orientation of `landscapeLeft` on iOS.

While this worked on my iPhone 15 Pro and was also posted as a safe-to-assume default value in Apple Developer Forums, it is apparently not safe to assume that this is the default value. 🤦 

On my iPhone 8, the preview was rotated by 90deg.

So instead, this PR now attempts to dynamically get the sensor orientation value by creating a dummy capture session, adding an output to it, and getting it's default orientation value. 

This then needs to be rotated by 90deg (I don't know why), and then we have a default orientation value for all phones - I think.

I'd need to test this on iPads and other iPhones first though

### DIFF
--- a/package/ios/Core/CameraSession+Configuration.swift
+++ b/package/ios/Core/CameraSession+Configuration.swift
@@ -113,8 +113,8 @@ extension CameraSession {
       if configuration.isMirrored {
         // 2.1. If mirroring is enabled, mirror all connections along the vertical axis
         videoOutput.isMirrored = true
-        if videoOutput.orientation.isPortrait {
-          // 2.2. If we have a portrait orientation, we need to flip it upside down so it is mirrored on horizontal axis as well
+        if videoOutput.orientation.isLandscape {
+          // 2.2. If we have a landscape orientation, we need to flip it to counter the mirroring on the wrong axis.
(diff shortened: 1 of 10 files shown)


mrousavy:
Something interesting I found out is that the new iOS 17 API `videoRotationAngle` does not do the same thing as the old API `videoOrientation`.

`videoRotationAngle` is by default always `0`, and setting it to `90` will rotate the output by 90 degrees.

`videoOrientation` has a default value, for example `landscapeRight` for back-, and `landscapeLeft` for front-cameras. Setting this from `landscapeRight` to `portrait` will cause it to rotate the output by 90 degrees.

The reason why this is a problem is because with `videoRotationAngle` we don't know what it's default orientation is. It's like a black-box. We can't just set `videoRotationAngle = 90`, because what if the output is actually naturally `landscapeLeft`? Then it would need to be set to `270`, not `90`. **But there is no API to find that out!** 🤦 

So right now I removed those new APIs, since they won't work with our case. I keep using `videoOrientation` at least until Apple adds an API for getting the device's native sensor orientation.

Issue #3048, opened 2024-07-03
🐛 [iOS] TypeError: Cannot read property 'CameraDevices' of undefined


short-dsb:
### What's happening?

Trying to invoke `useCameraDevice` on an iOS simulator in 4.4.1 results in a JS exception:

> Error: Exception in HostObject::get: for prop 'CameraDevices': <unknown exception>, js engine: hermes
> TypeError: Cannot read property 'CameraDevices' of undefined

This issue was not present in 4.4.0.

### Reproduceable Code

```tsx
const device = useCameraDevice("back", {
  physicalDevices: [
    "ultra-wide-angle-camera",
    "wide-angle-camera",
    "telephoto-camera",
  ],
});

<Camera
  ref={camera}
  device={device}
  enableZoomGesture={true}
  isActive={isActive}
  photo={true}
  resizeMode="contain"
  style={styles.camera}
/>
```

### Relevant log output

```shell
13:06:01.742: [error] 📸 VisionCamera.physicalDeviceDescriptor: Unknown AVCaptureDevice.DeviceType ()! Falling back to wide-angle-camera..
```

### Camera Device

```json
// N/A the exception prevents this from being logged.
```

### Device

iPhone 15 simulator (iOS 17.2)

### VisionCamera Version

4.4.1

### Can you reproduce this issue in the VisionCamera Example app?

I didn't try (⚠️ your issue might get ignored & closed if you don't try this)

### Additional information

- [ ] I am using Expo
- [ ] I have enabled Frame Processors (react-native-worklets-core)
- [X] I have read the [Troubleshooting Guide](https://react-native-vision-camera.com/docs/guides/troubleshooting)
- [X] I agree to follow this project's [Code of Conduct](https://github.com/mrousavy/react-native-vision-camera/blob/main/CODE_OF_CONDUCT.md)
- [X] I searched for [similar issues in this repository](https://github.com/mrousavy/react-native-vision-camera/issues) and found none.

rubbishCoder:
@short-dsb  try downgrading to the below version, it worked for me

"react-native-vision-camera": "4.4.0"

fleuverouge:
> @short-dsb try downgrading to the below version, it worked for me
> 
> "react-native-vision-camera": "4.4.0"

Version 4.4.0 has issues with orientations in older iOS so it's not an option.

Dat-Mobile:
I'm using this as the workaround, wait for the crash to fix in the next version
```
[...]

PR #3064, opened 2024-07-10
fix: Make `AVCaptureDevice.sensorOrientation` safer

mrousavy:
## What

Makes the getter for `sensorOrientation` safer by checking if we can set a preset, add the input, and add the output before actually doing it.

This could prevent some crashes with external/continouity cameras.


mrousavy:
This fixes https://github.com/mrousavy/react-native-vision-camera/issues/3048.

Issue #3051, opened 2024-07-05
🐛 Frame orientation is in landscape-right when phone is held in portrait mode


Marius-Adam:
### What's happening?

I want to scan text in portrait mode but the camera does not pick up anything only if the text is sideways or the phone is in landscape mode. 

When logging out the frame.orientation it comes back as landscape-right when the phone is held in portrait

https://github.com/mrousavy/react-native-vision-camera/assets/63927286/a1e9f58d-27f5-47bf-848a-f579c03e6b49

### Reproduceable Code

```tsx
import React from "react";
import { StyleSheet } from "react-native";
import { Camera, useCameraDevice, useFrameProcessor } from "react-native-vision-camera";
import { useTextRecognition } from "react-native-vision-camera-text-recognition";

function App() {
  const device = useCameraDevice("back");
  const options = { language: "latin"};
  const { scanText } = useTextRecognition(options);
  const frameProcessor = useFrameProcessor((frame) => {
    "worklet";
    console.log(frame.orientation)
    const data = scanText(frame);
   
  }, []);
  return (
    <>
      {!!device && (
        <Camera
          style={StyleSheet.absoluteFill}
          device={device}
          isActive
          frameProcessor={frameProcessor}
        />
      )}
    </>
  );
}
export default App;
```

### Relevant log output

```shell
Running "main" with {"rootTag":1,"initialProps":{"concurrentRoot":false}}
Loading react-native-worklets-core...
Worklets loaded successfully
VisionCameraProxy: Creating Worklet Context...
VisionCameraProxy: Worklet Context Created!
🟢 Creating JS object for module 'ExpoKeepAwake'
Looking up Frame Processor Plugin "scanText"...
Frame Processor Plugin "scanText" found! Initializing...
Initialized TensorFlow Lite runtime.
INFO: Initialized TensorFlow Lite runtime.
13:27:21.231: [info] 📸 VisionCamera.didSetProps(_:): Updating 22 props: [onInitialized, cameraId, position, enableBufferCompression, onOutputOrientationChanged, onStarted, preview, top, onCodeScanned, right, isActive, isMirrored, onViewReady, onError, onStopped, onPreviewOrientationChanged, onPreviewStopped, enableFrameProcessor, onPreviewStarted, left, bottom, onShutter]
13:27:21.236: [info] 📸 VisionCamera.configurePreviewOrientation(_:): Updating Preview rotation: landscapeLeft...
[...]

PR #3077, opened 2024-07-12
fix: Fix `Frame.orientation`, it should always be fixed

mrousavy:
## What

So this has been a topic for a while, people constantly asked questions about this.

@xulihang also pointed out that `Frame.orientation` is a bit confusing.

So I am making this (kinda breaking) change (sorry!) to get this right.

First we need to understand a few things:

1. Preview will always be rotated accordingly to screen/view rotation. User can lock this using the 🔒 button in control center.
2. Photo and Video can either be the same as the Preview (`outputOrientation="preview"`), **or** be allowed to rotate even if the phone is locked (`outputOrientation="device"`, this follows the gyroscope/accelerometer). This allows building UIs like the native iOS Camera app, where the view all stays the same, but only the buttons can actually rotate (again, taking gyro/accelerometer into consideration).
3. Code Scanner should follow the Preview orientation, since the coordinates it returns are always respective to screen coordinates (**this is now fixed with this PR!** previously it was not the case)
4. Frame Processors allow full control, and their Frame buffers are always in **native sensor orientation**, meaning it is always some fixed orientation (e.g. `landscape-left`), no matter if the phone is in portrait or landscape.

Now for Point 4 we really need to understand that the Frame cannot by physically rotated (rotating pixels in a 4k buffer is expensive), so that's why it is always in a fixed orientation, which is just coming from the native device sensor. (e.g. `landscape-left`)

In order to make Frame Processor Plugins (e.g. MLKit face detection) work, we need to tell the plugin in what orientation the Frame is, so it can internally know how to read the pixels in the Frame.
If it is portrait, it can assume that a Face is upright.

This graphic explains it quite well:

<table>
<tr>
<th>Portrait</th>
<th>Landscape</th>
</tr>
<tr>
<td>  </td>
<td>  </td>
</tr>
</table>

The blue frame is always in the same orientation as you can see. The phone changes orientation, and the face inside the Frame changes orientation, but the Frame buffer itself is the same.

So now with this PR, `Frame.orientation` is always just the native device sensor orientation, and the user can then either just rotate it by this orientation (then it will just always be "up-right" in the phone/iPad's natural portrait orientation), or also rotate it more by `previewOrientation` or `outputOrientation` - this is fully up to the user depending on the use-case.

For example to detect faces you might want to rotate by `outputOrientation`, because otherwise the face is sideways in a portrait Frame and MLKit cannot detect faces sideways - only "up-right" faces.

### For Frame Processor Plugin authors

If you are a Frame Processor Plugin author, make sure that you use `Frame.orientation` correctly now.

For example, if you use Apple's Vision-SDK for Face/Image/Text recognition, set the `VisionImage.orientation` exactly to the `Frame.orientation`.

```swift
public override func callback(_ frame: Frame, withArguments arguments: [AnyHashable: Any]?) -> Any {
  let image = VisionImage(buffer: frame.buffer)
  image.orientation = frame.orientation
```

This is now in it's **native portrait orientation**.

If the user holds his phone sideways (aka _not_ in it's native portrait orientation), then you would also need to rotate that Orientation.

**VisionCamera does not do that automatically**, because each use case is different - for example in a Skia Frame Processor you want to lock the orientation to portrait, otherwise the canvas would rotate within itself.

So if you also want to support orientation in your Frame Processor, you need to rotate the `Frame.orientation` by whatever the current output orientation is.

To get the current output orientation you can just add it as a parameter to your Frame Processor Plugin - then the user can also decide whether he wants to support it or not. 

In your plugin:

```swift
public override func callback(_ frame: Frame, withArguments arguments: [AnyHashable: Any]?) -> Any {
  let image = VisionImage(buffer: frame.buffer)
  image.orientation = frame.orientation

  if let outputOrientation = arguments["outputOrientation"] as? String {
    // parse outputOrientation String and then rotate image.orientation by that
  }
```

From the user perspective:

```ts
function App() {
  const outputOrientation = useSharedValue('portrait')

    'worklet'
    myCustomFpPlugin(frame, { outputOrientation: outputOrientation.value })
  })

  return <Camera {...props} onOutputOrientationChanged={(o) => outputOrientation.value = o} />
```

**Note:** In a future version of VisionCamera I might add `outputOrientation` to `Frame` directly, so you don't have to ask the user for it.
Then you can decide whether to use the raw orientation (e.g. for skia plugins), the preview's orientation, or the output orientation.

### DIFF
--- a/docs/docs/guides/ORIENTATION.mdx
+++ b/docs/docs/guides/ORIENTATION.mdx
@@ -42,6 +42,8 @@ Frame Processors will stream frames in a potentially "wrong" orientation, and th
 :::info
 This needs to be handled manually, see [`Frame.orientation`](/docs/api/interfaces/Frame#orientation).
 For example, in MLKit just pass the `Frame`'s `orientation` to the `detect(...)` method.
+
+Instead of always rotating up-right to portrait, you might also want to rotate the Frame to either preview-, or output-orientation, depending on your use-case.
 :::
 
 ## Implementation
@@ -97,9 +99,12 @@ For a smoother user experience, you should animate changes to the UI rotation. U
 In a Frame Processor, frames are streamed in their native sensor orientation.
 This means even if the phone is rotated from portrait to landscape, the Frame's [`width`](/docs/api/interfaces/Frame#width) and [`height`](/docs/api/interfaces/Frame#height) stay the same.
 
-The Frame's [`orientation`](/docs/api/interfaces/Frame#orientation) represents it's orientation relative to the current target orientation.
+The Frame's [`orientation`](/docs/api/interfaces/Frame#orientation) represents the **image buffer's orientation, relative to the device's native portrait mode**.
 
-For example, if the phone is held in `portrait` mode and the Frame's `orientation` is `landscape-right`, it is 90° rotated and needs to be counter-rotated by -90° to appear "up-right".
+For example, if the Frame's `orientation` is `landscape-right`, it is 90° rotated and needs to be counter-rotated by -90° to appear "up-right".
+
+On an iPhone, "up-right" means portrait mode (the home-button is at the bottom). On an iPad, "up-right" might mean a landscape orientation.
+
 Instead of actually rotating pixels in the buffers, frame processor plugins just need to interpret the frame as being rotated.
 
 MLKit handles this via a `orientation` property on the `MLImage`/`VisionImage` object:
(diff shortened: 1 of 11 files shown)


mrousavy:
btw; In a future version of VisionCamera (maybe 5.x.x), `Frame` will be a Swift-based class that has better accessors (for PixelFormat and Orientation), and will also hold preview- and output-orientation.

This will also be more efficient, as I will skip the Objective-C route.

I am working on building this foundation in a private repo elsewhere, the idea is that this will be much more efficient 😄

Flocurry:
Hello @mrousavy 

Can you explain me what is the problem please ?

https://github.com/mrousavy/react-native-vision-camera/issues/3051#issuecomment-2243708666

i7soft:
i test on my device, android (Google Pixel2 Android Version 11) ,ios (iPad mini 6 iPadOS Version 17.0.3), the "frame.orientation" value is diffent between them
iShot_2025-01-04_10 57 54

so "src/skia/useSkiaFrameProcessor.ts function withRotatedFrame()" not work for me, i must commented code this code, and change the translate and rotate for diffent platform

and "src/skia/useSkiaFrameProcessor.ts  function getSurfaceSize()"return wrong value on ios, 
because frame orientation always is landscape, when deivice is portrait,return width is frame.height is correct,
but when deivice is landscape,the frame is landscape too,return width is frame.height is wrong,the return width should be frame.width

PR #3079, opened 2024-07-12
fix: Move `Frame` into VisionCamera `Core` (and change namespace)

mrousavy:
### Android

Moves `Frame` into the VisionCamera `Core` namespace and rewrites it to a Kotlin class.

⚠️ This might be a breaking change for Frame Processor Plugins, this is what they need to change:

```diff
- import com.mrousavy.camera.frameprocessors.Frame;
+ import com.mrousavy.camera.core.Frame;
```

### iOS

Moves `Frame` into the VisionCamera `Core` subspec, and rewrites it to a Swift class.

This simplifies a few occasions where `CMSampleBuffer` was used in addition to `orientation` and `isMirrored` props, and also makes it a bit more efficient as it is now a Swift class instead of an NSObject.


mrousavy:
V5

PR #3089, opened 2024-07-17
feat: Support tap to focus without a Preview View

mrousavy:
## What

Currently, focusing only works when using the normal Preview View.
When using Skia, there is no Preview View and we cannot focus.

My idea was to move the layer point conversion method to JS, so that we can do it synchronously & instantly - either convert a view point to PreviewView coordinates, or Skia video output stream coordinates.

Unfortunately both iOS and Android do not expose matrixes that allow me to convert `x`/`y` coordinates myself.
You have to use their functions for it, and on Android it only works on the UI Thread (why do you need to dispatch to the UI thread for some number multiplication you ask? I dont know. 🤦‍♂️)

So now I think the only solution we have is to actually pass a prop to `focus` which is `coordinateSystem: 'preview' | 'video'`... No way around it 🤦‍♂️


mrousavy:
Exciting news: this is now obsolete, as VisionCamera V5 now supports this! 🥳

PR #3098, opened 2024-07-24
feat: Use `setExternalMemoryPressure` to notify JS VM about `Frame`'s actual size (on GPU)

mrousavy:
## What

Uses `setExternalMemoryPressure` to notify the JS VM about the `Frame`'s actual size (the byte buffer on the GPU, which is `bytesPerRow * height`).

If the JS VM (e.g. Hermes) knows the true size of the `Frame` in memory, it will delete old stale `FrameHostObject`s sooner instead of just keeping them floating around in memory because it thinks it's just a few bytes in size. (right now around 400 of those pile up until GC hits, with this PR only 3 pile up at a time)

This doesn't really matter for us since we need to eagerly delete the Frame after the Frame Processor is done, because the Camera pipeline stalls otherwise.

So we still need to keep the manual ref counting stuff. Make sure to use it properly!

### DIFF
--- a/package/VisionCamera.podspec
+++ b/package/VisionCamera.podspec
@@ -1,8 +1,10 @@
 require "json"
 
 package = JSON.parse(File.read(File.join(__dir__, "package.json")))
+$config = find_config()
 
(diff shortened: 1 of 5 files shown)


mrousavy:
V5 now supports this thanks to Nitro!

PR #3139, opened 2024-08-14
Fix: clear orientation listener on close

ldalzottomp:
## What

This PR cleans android orientation listener on camera session close.
If this is not done, the events are still fired and they accumulate until you kill the app.

## Changes

Added a `clearListeners` functions in the OrientationManager which is then called when CameraSession.close() is called.

## Tested on

Samsung Galaxy tab active4 pro. Android 14.

### DIFF
--- a/package/android/src/main/java/com/mrousavy/camera/core/OrientationManager.kt
+++ b/package/android/src/main/java/com/mrousavy/camera/core/OrientationManager.kt
@@ -67,13 +67,17 @@ class OrientationManager(private val context: Context, private val callback: Cal
     }
   }
 
+  // remove previous listeners if attached
+  fun clearListeners() {
+    displayManager.unregisterDisplayListener(displayListener)
+    orientationListener.disable()
+  }
+
   fun setTargetOutputOrientation(targetOrientation: OutputOrientation) {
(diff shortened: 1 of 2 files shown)


Kamiss7 (approved):

package/android/src/main/java/com/mrousavy/camera/core/OrientationManager.kt
```diff
@@ -67,13 +67,17 @@ class OrientationManager(private val context: Context, private val callback: Cal
     }
   }
 
+  // remove previous listeners if attached
```
mrousavy: Can you move this comment down again to line 80 where it was before?

package/android/src/main/java/com/mrousavy/camera/core/OrientationManager.kt
```diff
@@ -67,13 +67,17 @@ class OrientationManager(private val context: Context, private val callback: Cal
...
+  fun clearListeners() {
```
mrousavy: Let's rename this to `stopUpdates`

mrousavy (changes requested):
Hey - thanks for your PR! This is good, just a few code style nitpicks.

ldalzottomp:
Thanks for handling the requested changes

PR #3175, opened 2024-09-07
fix: codeScanner unsupported ios version error handing

mgefimov:
## What
Enabling `codabar` format in code scanner is dangerous

Scenario: You set up the code scanner with several formats, including `codabar`. On Android and iOS 15.4 and above, everything works well. However, on iOS 15.3 or below, you get a black screen, and the scanner doesn't work at all.

## Changes

On iOS 15.3 or below, if you enable an unsupported code scanner format (e.g., `codabar`), the scanner works without this format, instead of a black screen.

## Tested on
- iphone SE (second rev.), iOS 15.0
- iphone X, ios 15.5

### DIFF
--- a/package/ios/Core/Types/CodeScannerOptions.swift
+++ b/package/ios/Core/Types/CodeScannerOptions.swift
@@ -16,7 +16,7 @@ struct CodeScannerOptions: Equatable {
 
   init(fromJsValue dictionary: NSDictionary) throws {
     if let codeTypes = dictionary["codeTypes"] as? [String] {
-      self.codeTypes = try codeTypes.map { value in
+      self.codeTypes = try codeTypes.compactMap { value in
         return try AVMetadataObject.ObjectType(withString: value)
       }
     } else {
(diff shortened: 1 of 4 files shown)


package/ios/Core/Types/CodeScannerOptions.swift
```diff
@@ -16,8 +16,18 @@ struct CodeScannerOptions: Equatable {
...
     if let codeTypes = dictionary["codeTypes"] as? [String] {
-      self.codeTypes = try codeTypes.map { value in
-        return try AVMetadataObject.ObjectType(withString: value)
+      self.codeTypes = try codeTypes.compactMap { value in
+        do {
+          return try AVMetadataObject.ObjectType(withString: value)
+        } catch let CameraError.codeScanner(codeScannerError) {
+          switch codeScannerError {
+          case .iosVersionNotSupported:
+            VisionLogger.log(level: .warning, message: codeScannerError.message)
```
mrousavy: There's no point in throwing an error if we just ignore it later.

Please add the `ios-version-not-supported` error to the TypeScript error types, and then in here we just let the error propagate upwards for the user to handle.

mrousavy: but then hmm he will just ignore it later anyways as well

mgefimov: Is it okay to propagate the error to the JS side? In that case, the `onError` callback will fire, and the user can handle the error. However, the user might assume that this callback fired because the camera failed to start and cannot operate, leading them to show an error screen, for instance. But the scanner can still operate.

I can refactor the solution without throwing an error

mgefimov: @mrousavy 

I've refactored without throwing an error. Also i've added supported code types table to the docs

mgefimov:
Hi @mrousavy . I hope you're doing well! I have updated PR, can you review it please?

mrousavy:

PR #3222, opened 2024-10-03
working prototype of orientation normalisation

Szymon20000:


mrousavy:
thanks but closing since V5 makes this out of sync now

PR #3249, opened 2024-10-17
chore: monorepo setup with bun 

hannojg:


hannojg:
All thats left is to update the vercel build step to just call bun install

PR #3250, opened 2024-10-17
[WIP] feat: new arch support

hannojg:
## Changes

based on the following PR which should be merged first:

- https://github.com/mrousavy/react-native-vision-camera/pull/3249


hannojg:
Currently blocked by:

- https://github.com/facebook/react-native/issues/47113

Looking into the issue, but leaning towards shipping a manual write of the FabricComponent classes with VisionCamera (just like Rn TextInput or ScrollView does it). Because it will take time until the fix is landing in react native and we can't ask users to upgrade codegen or apply a patch

PR #3390, opened 2025-01-30
fix: Simplify logic and also catch extension manager error

mrousavy:
## What

Also catches ExtensionManager initialization error


mrousavy:

PR #3416, opened 2025-02-15
feat: Add the option to enable idle timer on iOS

rithik-b:
## What

Added support for enabling the idle timer while the camera is active on iOS. Use case is I am using this for barcode scanning and I want the phone to go to sleep if the user prefers it, in case they leave their phone on and leave it alone. On android it seems like the phone does go to sleep when left alone so I have left it as is.

## Changes

Added `enableIdleTimer` prop (defaults to false) and implemented it for iOS. 

## Tested on

* iPhone 15 Pro Max, iOS 18.3

Tested both with enableIdleTimer set to true as well as false to see if everything works as expected.

### DIFF
--- a/package/ios/Core/CameraConfiguration.swift
+++ b/package/ios/Core/CameraConfiguration.swift
@@ -50,6 +50,8 @@ final class CameraConfiguration {
   // isActive (Start/Stop)
   var isActive = false
 
+  var enableIdleTimer = false
+
   // Audio Session
   var audio: OutputConfiguration<Audio> = .disabled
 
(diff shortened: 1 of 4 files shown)


rithik-b:
Hey @mrousavy did you have a chance to take a look at this yet? Since my use case is for timing out for barcode scanning, would it make sense to disable idle timer only when a video is being recorded (and remove this prop?) I think I saw the disable idle timer was only added for not sleeping when a video is being recorded so let me know what you think.

mrousavy:
Hey thanks for the PR - in V5 we have removed this and expect the user to manually disable the idle timer if he wants. We rely on system policy now

PR #3423, opened 2025-02-17
fix: Fix orientation updating to `portrait` when phone is flat

mrousavy:
## What

Previously, when the phone is flat, `outputOrientation` would be updated to `'portrait'` - which is not correct since it can also be landscape before the phone is being laid flat.

Instead, we now return `nil` if the orientation is flat, since it is not known how the user is holding the phone. In other words, we skip an orientation update for flat now.


mrousavy:

PR #3429, opened 2025-02-20
fix: Don't send `onOrientationChanged` event if phone is flat on Android

mrousavy:
## What

Same as https://github.com/mrousavy/react-native-vision-camera/pull/3423, but for Android.


mrousavy:

PR #3549, opened 2025-06-12
Feature | Support onBytesWrittenVideo to receive bytes on JS while recording Video

felire:
## What

This PR Adds the `onBytesWrittenVideo` callback in order to receive the bytes of the video while it is being recorded.

### IOS

- Implementing `DispatchSourceTimer` on Ios in order to read the file size from the FileSystem every 400 ms, that will be executed after the video has started and when finished the timer is cleaned.
-  Adding `onBytesWrittenVideoEvent` as `RCTDirectEventBlock` in order to send events to JS in `RCTViewManager`

### Android

- Creating BytesWrittenVideoEvent class
- Adding `onBytesWrittenVideo` on the `getExportedCustomDirectEventTypeConstants` and all the logic to treat that as an Event and send data to JS
- Send the data directly from `VideoRecordEvent.Status` as it is available there

### TSX/TS

- Adding types and JS Docs
- Adding the onBytesWrittenVideo and destructuring the `NativeEvent` to just expose `(bytes: number) => void`

### Docs

- Add an example on video page
## Tested on

- iPhone 8
- Samsung Galaxy S24 FE

### DIFF
--- a/package/ios/Core/CameraSession+Video.swift
+++ b/package/ios/Core/CameraSession+Video.swift
@@ -143,10 +143,7 @@ extension CameraSession {
           let path = session.url.path
           if let size = try? FileManager.default.attributesOfItem(atPath: path)[.size] as? NSNumber {
             let bytes = size.doubleValue
-
-            DispatchQueue.main.async {
-              onBytesWritten(bytes)
-            }
+            onBytesWritten(bytes)
           }
         }
         self.recordingSizeTimer = timer
--- a/package/ios/Core/CameraSession+Video.swift
+++ b/package/ios/Core/CameraSession+Video.swift
@@ -18,7 +18,8 @@ extension CameraSession {
    */
   func startRecording(options: RecordVideoOptions,
                       onVideoRecorded: @escaping (_ video: Video) -> Void,
-                      onError: @escaping (_ error: CameraError) -> Void) {
+                      onError: @escaping (_ error: CameraError) -> Void,
+                      onBytesWritten: @escaping (_ bytes: Double) -> Void) {
     // Run on Camera Queue
     CameraQueues.cameraQueue.async {
       let start = DispatchTime.now()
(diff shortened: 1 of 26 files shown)


mrousavy:
cool!

package/ios/Core/CameraSession+Video.swift
```diff
@@ -128,6 +131,26 @@ extension CameraSession {
...
+            return
+          }
+
+          let path = session.url.path
+          if let size = try? FileManager.default.attributesOfItem(atPath: path)[.size] as? NSNumber {
+            let bytes = size.doubleValue
+
+            DispatchQueue.main.async {
+              onBytesWritten(bytes)
+            }
```
mrousavy: Does this have to run on the main queue?

package/src/types/CameraProps.ts
```diff
@@ -416,5 +416,17 @@ export interface CameraProps extends ViewProps {
    * ```
    */
   codeScanner?: CodeScanner
+  /**
+   * Used to get the bytes written to the video file on real time while it is being recorded
```
mrousavy:
```suggestion
   * Fires every few hundred milliseconds to notify how many
   * total bytes are currently written to the video file.
```

package/src/types/VideoFile.ts
```diff
@@ -54,3 +54,7 @@ export interface VideoFile extends TemporaryFile {
    */
   height: number
 }
+
+export interface OnBytesWrittenVideoEvent {
+  bytesWritten: number
```
mrousavy, line 62:
```suggestion
  /**
   * Represents the amount of total bytes written to the video file.
   */
  bytesWritten: number
```

mrousavy (approved):

package/ios/Core/CameraSession+Video.swift
felire: It wasn't needed, I removed the execution on the main queue

package/src/types/CameraProps.ts
felire: Done

package/src/types/VideoFile.ts
felire: Done

PR #3583, opened 2025-07-18
fix:- Fix build for RN 0.81

riteshshukla04:
## What
Fixes build issue on RN 0.81.0

## Changes
This PR changes Mutable Map to Map . I think the type has been changed when this was migrated to Kotlin (https://github.com/facebook/react-native/pull/51626)

Also changes `currentActivity` to `reactApplicationContext.currentActivity` to fix build

## Tested on

This is build issue
## Related issues
N/A


mrousavy:
cool thanks! if it builds before RN 0.81 then we can use it.

PR #3619, opened 2025-09-02
Do not use deprecated `MapBuilder` but use `mapOf` instead

cortinico:
## What

While looking at the source code of this library, I've realized you still use a deprecated `MapBuilder` class.
If you're using Kotlin, the reccomendation is to use `mapOf` from Kotlin instead:
https://github.com/facebook/react-native/blob/8f0713fd4bfdb8ece5616c9aa47a20ee381b792e/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/common/MapBuilder.kt#L11


cortinico:
This is also a nice-to-have, build would keep on working with the old API still

mrousavy:

I believe this is now obsolete, as VisionCamera V5 is on Nitro now - but this would've been a nice fix for the time. I dont know why I didnt see this PR back in September

PR #3650, opened 2025-10-18
feat(shutdown): implement graceful shutdown for camera session

EthanDM:
## What
Fix the navigation freeze on React Navigation/Fabric by shutting the camera session down deterministically when the view unmounts.

## Changes
- Add CameraSession.shutdown, which disables photo/video/audio/code-scanner/location/torch before flipping isActive = false, and clears delegates while removing outputs.  
- Stop both capture sessions asynchronously during CameraSession.deinit as a backstop.  
- Have CameraView call the shutdown helper on unmount/deinit, log if teardown exceeds 2 s, and guard against concurrent shutdowns.

## Tested on
- iPhone 14 Pro (iOS 26.0.1) with code scanner enabled, navigating away from the camera screen (no freeze).

## Related issues
- Fixes #3636

PR #3664, opened 2025-11-06
fix: implement more robust config and tear down bookkeeping

rustle:
## What

This is a revision pass on the work done in https://github.com/mrousavy/react-native-vision-camera/pull/3650

Fix the navigation freeze on React Navigation/Fabric by shutting the camera session down deterministically when the view unmounts.

## Changes

Instead of introducing a new shutdown method revise use of configure

* Cleaning up capture sessions in deinit likely indicates some bookkeeping isn't completing and is likely to obscure rather than fix bugs
* Adding a completion handler to config adds a lot of complexity we can avoid and replace with a more correct DispatchGroup (the group doesn't notify until any async audio and location work complete)
* Switch existing DispatchTime usage for dropping redundant config calls to a simple monotonic counter that uses an unfair lock
* remove didScheduleShutdown as deactivateCameraSession() is safe to call repeatedly

## Tested on

iPhone 12 mini running 18.7.1

### DIFF
--- a/package/ios/Core/CameraSession+Configuration.swift
+++ b/package/ios/Core/CameraSession+Configuration.swift
@@ -61,21 +61,6 @@ extension CameraSession {
 
     // Remove all outputs
     for output in captureSession.outputs {
-      // The following delegates are weak references
-      // so cleaning them up is not strictly necessary
-      // from a safety standpoint.
-      // It is good to mark that between this point
-      // and logic further down configureOutput()
-      // messages to delegate would be in an undefined
-      // state and with delegate == self the weak 
-      // references wouldn't be cleared during that 
-      // window.
-      if let metadataOutput = output as? AVCaptureMetadataOutput {
-        metadataOutput.setMetadataObjectsDelegate(nil, queue: nil)
-      }
-      if let videoOutput = output as? AVCaptureVideoDataOutput {
-        videoOutput.setSampleBufferDelegate(nil, queue: nil)
-      }
       captureSession.removeOutput(output)
     }
     photoOutput = nil
--- a/package/ios/Core/CameraSession+Configuration.swift
+++ b/package/ios/Core/CameraSession+Configuration.swift
@@ -61,6 +61,12 @@ extension CameraSession {
 
     // Remove all outputs
     for output in captureSession.outputs {
+      if let metadataOutput = output as? AVCaptureMetadataOutput {
+        metadataOutput.setMetadataObjectsDelegate(nil, queue: nil)
+      }
+      if let videoOutput = output as? AVCaptureVideoDataOutput {
+        videoOutput.setSampleBufferDelegate(nil, queue: nil)
+      }
       captureSession.removeOutput(output)
     }
     photoOutput = nil
--- a/package/ios/Core/CameraSession+Configuration.swift
+++ b/package/ios/Core/CameraSession+Configuration.swift
@@ -61,6 +61,15 @@ extension CameraSession {
 
     // Remove all outputs
     for output in captureSession.outputs {
+      // The following delegates are weak references
+      // so cleaning them up is not strictly necessary
+      // from a safety standpoint.
+      // It is good to mark that between this point
+      // and logic further down configureOutput()
+      // messages to delegate would be in an undefined
+      // state and with delegate == self the weak 
+      // references wouldn't be cleared during that 
+      // window.
       if let metadataOutput = output as? AVCaptureMetadataOutput {
         metadataOutput.setMetadataObjectsDelegate(nil, queue: nil)
       }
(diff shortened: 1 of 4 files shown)


package/ios/React/CameraView.swift
```diff
@@ -135,10 +135,15 @@ public final class CameraView: UIView, CameraSessionDelegate, PreviewViewDelegat
...
       }
     } else {
+      deactivateCameraSession()
       fpsSampleCollector.stop()
     }
   }
 
+  deinit {
+    deactivateCameraSession()
+  }
```
mrousavy, line 148: As I was saying in #3650, I don't like running (async) stuff in `deinit`. We have it on view unmount at the top anyways.

Can you try removing this and check if it still gracefully shuts down for your case?

package/ios/Core/CameraSession+Configuration.swift
```diff
@@ -61,6 +61,12 @@ extension CameraSession {
 
     // Remove all outputs
     for output in captureSession.outputs {
+      if let metadataOutput = output as? AVCaptureMetadataOutput {
+        metadataOutput.setMetadataObjectsDelegate(nil, queue: nil)
+      }
+      if let videoOutput = output as? AVCaptureVideoDataOutput {
+        videoOutput.setSampleBufferDelegate(nil, queue: nil)
+      }
```
mrousavy: Same as in #3650 - I really don't think this is necessary as they store a `weak` ref to delegate - this feels like monkeypatching.

Did you try removing it?

And even if it has to stay, please add a short comment above explaning why this needs to be unset.

package/ios/Core/CameraSession.swift
```diff
@@ -113,8 +113,22 @@ final class CameraSession: NSObject, AVCaptureVideoDataOutputSampleBufferDelegat
 
     VisionLogger.log(level: .info, message: "configure { ... }: Waiting for lock...")
 
+    let slowConfigurationWarning = DispatchWorkItem {
+      VisionLogger.log(level: .warning, message: "configure { ... }: is still running after 2 seconds.")
+    }
+    CameraQueues.cameraQueue.asyncAfter(deadline: .now() + .seconds(2), execute: slowConfigurationWarning)
+
+    let completionGroup = DispatchGroup()
```
mrousavy, line 121: This is a good change, I like it.

package/ios/Core/CameraSession.swift
```diff
@@ -113,8 +113,22 @@ final class CameraSession: NSObject, AVCaptureVideoDataOutputSampleBufferDelegat
 
     VisionLogger.log(level: .info, message: "configure { ... }: Waiting for lock...")
 
+    let slowConfigurationWarning = DispatchWorkItem {
+      VisionLogger.log(level: .warning, message: "configure { ... }: is still running after 2 seconds.")
```
mrousavy, line 117: Has this ever been logged? I'm guessing most of configure is pretty fast apart from initial start right

mrousavy (changes requested):
Thanks for your PR!

Left some comments

package/ios/Core/CameraSession+Configuration.swift
rustle: I inherited that from the previous PR and was a bit suspicious of it at first but the argument I'd make for keeping and could add in a comment is that: 
this doesn't effect if the delegate is a valid reference (which weak will handily manage for us) but rather the book keeping of saying this delegate shouldn't be messaged after this point in time is desirable.

All that being said, I don't have a reproducible issue this addresses and can omit it entirely.

package/ios/Core/CameraSession.swift
rustle: I haven't managed to instigate it to log myself, but like having it tracked for the pathological cases that the docs encouraging config off the main queue imply.

package/ios/React/CameraView.swift
rustle: Async in init certainly makes me nervous as well. I believe it is correctly architected in that it doesn't rely on self in any of its logic after returning from the synchronous call. A different approach could be too pull out just the front half of deactivate and let session handle it's own tear down as it deinits:
```
    UIApplication.shared.isIdleTimerDisabled = false

    currentConfigureCall.increment()
```

rustle: You're right about view init as well. In all of my testing we'd see deactivate called twice on teardown, which is safe, but if we're confident in unmount, redundant.

rustle:
I've added a comment to explain the manual delegate cleanup and made the work in deinit much smaller and synchronous, but if your preference is still to drop those neither are strictly necessary for this PR to work well.

package/ios/Core/CameraSession+Configuration.swift
mrousavy: I believe if you remove the output from the session it will not call the delegate anymore at all. so this might not be needed

rustle: Removed

notfound-yet:
I have a problem with the library; the phone no longer enters sleep mode after passing over a screen with a camera. Testing this PR solved my problem. Do you have an estimated date for accepting this PR or creating a test version? @mrousavy

nakshatrasinghh:
This PR works! Thank you for this fix 🚀

jkaufman:
Would love to upstream this. We're here to address any feedback.

mrousavy:

PR #3677, opened 2025-12-03
fix: iPhone 17 Pro/Pro Max crashes app when capturePhoto

herarya:
Fix iPhone 17 Pro/Pro Max crashes app when capturePhoto

## What

This PR fixes a crash on iPhone 17 Pro and Pro Max devices when attempting to capture photos. The crash was caused by calling `AVCapturePhotoOutput.capturePhoto(with:delegate:)` when the video connection was not yet active or enabled, resulting in the error: "No active and enabled video connection".

## Changes

* Added readiness check to ensure `AVCaptureSession.isRunning` is `true` before attempting photo capture
* Added validation to verify the video connection is both enabled and active before capture
* Implemented auto-repair logic that re-adds `AVCapturePhotoOutput` and resets the session preset when the video connection is not ready
* Added polling mechanism (max 0.5s, 50ms intervals) to handle timing races after session reconfiguration
* Session automatically starts if not running, with a brief wait period (200ms) for initialization

## Tested on

* iPhone 17 Pro,  iOS 26.x (verified fix works as expected)

https://github.com/user-attachments/assets/d4849409-8529-4de2-8883-cb0bdd1a11a5

* iPhone 16 verified
* iPhone 17 verified

### DIFF
--- a/package/ios/Core/CameraSession+Photo.swift
+++ b/package/ios/Core/CameraSession+Photo.swift
@@ -17,6 +17,46 @@ extension CameraSession {
   func takePhoto(options: TakePhotoOptions, promise: Promise) {
     // Run on Camera Queue
     CameraQueues.cameraQueue.async {
+      // Ensure AVCaptureSession is running
+      if !self.captureSession.isRunning {
+        VisionLogger.log(level: .error, message: "Capture session is not running. Attempting to start...")
+        self.captureSession.startRunning()
+        usleep(200_000) // 200ms
+        if !self.captureSession.isRunning {
+          promise.reject(error: .session(.cameraNotReady))
+          return
+        }
+      }
+
+      // Check video connection readiness
+      if let photoOutput = self.photoOutput,
+         let videoConnection = photoOutput.connection(with: .video),
+         !(videoConnection.isEnabled && videoConnection.isActive) {
+        VisionLogger.log(level: .error, message: "No active/enabled video connection. Attempting auto-repair...")
+        self.captureSession.beginConfiguration()
+        self.captureSession.removeOutput(photoOutput)
+        let newPhotoOutput = AVCapturePhotoOutput()
+        self.captureSession.addOutput(newPhotoOutput)
+        self.photoOutput = newPhotoOutput
+        if self.captureSession.canSetSessionPreset(.photo) {
+          self.captureSession.sessionPreset = .photo
+        }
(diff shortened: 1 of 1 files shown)


pouyaemami:
We just tried this and it looks promising but we are still seeing a greyed out video feed when taking a picture the first time around:
Screenshot 2025-12-03 at 15 39 35

When a picture is taken and we go back to the camera to re-take another one, the image looks right:
Screenshot 2025-12-03 at 15 39 52

Tebrynn:
I've got same problem. Grayed image and crash when take photo.

danielshamburger:
Hi there - are there any updates on the initial grey image? If so, is there a possibility of getting the iPhone 17 Pro fix included in a release? Thanks so much!

mrousavy:

yinminqian:
For anyone still on VisionCamera 4.x hitting this on iPhone 17 (and now iPhone 18, which reproduces the same grey preview + `takePhoto` crash): we worked around both symptoms without touching native code.

Root cause on our side: any custom `format` from `useCameraFormat` (we asked for screen aspect ratio + max resolution) makes 4.x pick a bad format on these devices. Passing no `format` at all fixes both the grey preview and the crash. The trade-off is the device default 4:3, so we switch `resizeMode` to `contain` to keep preview == captured photo.

```js
// Apple identifiers lag one behind marketing names: iPhone 17 = iPhone18,x, iPhone 18 = iPhone19,x
const isNewMultiCamIphone = () =>
  Platform.OS === 'ios' && Number((/^iPhone(\d+),/.exec(DeviceInfo.getDeviceId()) || [])[1]) >= 18

function usePhotoFormat(device) {
  const { width, height } = Dimensions.get('window')
  const defaultFormat = useCameraFormat(device, [
    { videoAspectRatio: height / width },
    { videoResolution: 'max' },
    { photoAspectRatio: height / width },
    { photoResolution: 'max' },
  ])
  return isNewMultiCamIphone() ? undefined : defaultFormat
}

<Camera
  device={device}
  format={format}
  resizeMode={isNewMultiCamIphone() ? 'contain' : 'cover'}
  ...
/>
```

Verified on iPhone 17 Pro and iPhone 18 with 4.3.1 + iOS 26. Older iPhones and Android keep the custom format and full-screen `cover`, unchanged. Not a real fix, but it unblocks 4.x users until they can move to V5.

PR #3691, opened 2026-01-08
fix: [Android] No available camera on some Samsung devices

moonjava2005:
## What

* On some Samsung devices, the Camera cannot be opened because `useCameraDevices()` always returns an empty device list.
* This issue happens **frequently and intermittently**, making the Camera unusable.
* The problem occurs **only on Android**, and is **especially common on certain Samsung models**.

---

## Why

* This is caused by a **race condition** during module initialization.
* When React Native binds constants via `getConstants()`, the `cameraProvider` may **not be fully initialized yet**, causing the exported device list to be empty.
* Although previous PRs attempted to mitigate this by initializing earlier, the issue is **not fully resolved**, and still happens frequently on newer Samsung devices.
* Additionally, native events related to camera device changes may be **emitted before the React Native JS runtime is ready**.
  If such events are sent too early, the JS layer will **never receive them**, resulting in a permanently empty device list.

---

## How

* Add a **new exported native method** to retrieve the current camera device list directly from `cameraProvider`, instead of relying solely on the one-time value from `getConstants()`.
* Update the JS hook logic on **Android only**:

  * If `useCameraDevices()` returns an empty list, the hook will **manually invoke** this native method to fetch the latest available devices.
* This ensures the device list is always retrieved from the **latest and fully initialized camera state**, eliminating the race condition.

---

### DIFF
--- a/package/android/src/main/java/com/mrousavy/camera/react/CameraDevicesManager.kt
+++ b/package/android/src/main/java/com/mrousavy/camera/react/CameraDevicesManager.kt
@@ -6,6 +6,7 @@ import android.util.Log
 import androidx.camera.extensions.ExtensionsManager
 import androidx.camera.lifecycle.ProcessCameraProvider
 import com.facebook.react.bridge.Arguments
+import com.facebook.react.bridge.Promise
 import com.facebook.react.bridge.ReactApplicationContext
 import com.facebook.react.bridge.ReactContextBaseJavaModule
 import com.facebook.react.bridge.ReactMethod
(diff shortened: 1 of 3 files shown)


Thomas-Vargas:
I’ve been running into the same issue on Android, mainly on Samsung devices, where useCameraDevices() intermittently returns an empty list on cold start and the camera never becomes available.

I tested this fix by patching it onto react-native-vision-camera@4.7.3 in an Expo SDK 51 / React Native 0.74.5 app, and it reliably resolves the problem across repeated launches.

For reference, this is happening on production devices (Galaxy A53 5G / SM-A536U, Android 16). The fallback to querying the camera provider at runtime instead of relying on getConstants() makes a lot of sense given how intermittent the race condition is.

Thanks for putting this together — this has been a real pain point on these devices.

soportetb:
Good day, I will apply the suggested patch; it doesn't seem like the maintainer is going to fix it anytime soon.

bowedcontainer2:
I have applied this fix as patch onto react-native-vision-camera@4.7.3 and it fixed my issue! Thanks for the sleuthing!

mrousavy:

Issue #3648, opened 2025-10-17
🐛 'UIWindowScene' is only available in iOS 13.0 or newer


mushthak02:
### What's happening?

'UIWindowScene' is only available in iOS 13.0 or newer

<img width="744" height="549" alt="Image" src="https://github.com/user-attachments/assets/109ee892-469f-4d7e-87dc-19e6de642e6d" />

### Reproduceable Code

```tsx
'UIWindowScene' is only available in iOS 13.0 or newer
```

### Relevant log output

```shell
'UIWindowScene' is only available in iOS 13.0 or newer
```

### Camera Device

```json
'UIWindowScene' is only available in iOS 13.0 or newer
```

### Device

iphone 17 pro

### VisionCamera Version

latest

### Can you reproduce this issue in the VisionCamera Example app?

I didn't try (⚠️ your issue might get ignored & closed if you don't try this)

### Additional information

- [ ] I am using Expo
- [x] I have enabled Frame Processors (react-native-worklets-core)
- [x] I have read the [Troubleshooting Guide](https://react-native-vision-camera.com/docs/guides/troubleshooting)
- [x] I agree to follow this project's [Code of Conduct](https://github.com/mrousavy/react-native-vision-camera/blob/main/CODE_OF_CONDUCT.md)
- [x] I searched for [similar issues in this repository](https://github.com/mrousavy/react-native-vision-camera/issues) and found none.

sofianegargouri:
Seems like a duplicate of https://github.com/mrousavy/react-native-vision-camera/issues/3040

mrousavy:
Upgrade your Xcode version. Also make sure your min iOS target is above 13.0

Issue #3404, opened 2025-02-09
🔧 Facing: 'UIWindowScene' is only available in iOS 13.0 or newer


praveenflymeout:
### How were you trying to build the app?

Hello supprt team,

I am facing issue like 'UIWindowScene is only available in iOS 13.0 or newer' at the time of app running from Xocde, I am using version 4.6.3, i try to updated my ios Deployment Target 14.2 and also i done changes inside pod file like 'config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'] = '14.2''. But still same issue is occure, Please check attached screenshot.

<img width="1496" alt="Image" src="https://github.com/user-attachments/assets/424a1f46-2392-495b-ad3d-4572e997a168" />

### Full build logs

```tsx
Showing All Errors Only

Prepare build
error: Build service could not create build operation: unknown error while handling message: MsgHandlingError(message: "unable to initiate PIF transfer session (operation in progress?)")


Build service could not create build operation: unknown error while handling message: MsgHandlingError(message: "unable to initiate PIF transfer session (operation in progress?)")



Build failed    09/02/25, 3:12 PM    0.4 seconds
```

### Project dependencies

```json
"dependencies": {
    "react": "18.2.0",
    "react-native": "0.72.3",
    "react-native-fast-image": "^8.6.3",
    "react-native-fbsdk-next": "^13.3.0",
    "react-native-flipper": "0.212.0",
    "react-native-gesture-handler": "^2.13.1",
    "react-native-video": "6.1.2",
    "react-native-view-shot": "^3.8.0",
    "react-native-vision-camera": "4.6.3",
  }
```

### VisionCamera Version

4.6.3

### Target platforms

iOS

### Operating system

MacOS

### Can you build the VisionCamera Example app?

I didn't try (⚠️ your issue might get ignored & closed if you don't try this)

### Additional information

- [ ] I am using Expo
- [ ] I have enabled Frame Processors (react-native-worklets-core)
- [x] I have read the [Troubleshooting Guide](https://react-native-vision-camera.com/docs/guides/troubleshooting)
- [x] I agree to follow this project's [Code of Conduct](https://github.com/mrousavy/react-native-vision-camera/blob/main/CODE_OF_CONDUCT.md)
- [x] I searched for [similar issues in this repository](https://github.com/mrousavy/react-native-vision-camera/issues) and found none.

Issue #3040, opened 2024-07-02
🐛 'UIWindowScene' is only available in iOS 13.0 or newer


Romick2005:
### What's happening?

![image](https://github.com/mrousavy/react-native-vision-camera/assets/17779660/52dd534b-308b-4aa6-9797-2a1e462e83f0)

Please add version checks.

### Reproduceable Code

```tsx
No code needed.
```

### Relevant log output

```shell
It does not build on iOS for relevant output, but works on Android.
```

### Camera Device

```json
{
  device: "Any iPhone device with iOS > 13"
}
```

### Device

iPhone XS

### VisionCamera Version

4.4.0

### Can you reproduce this issue in the VisionCamera Example app?

I didn't try (⚠️ your issue might get ignored & closed if you don't try this)

### Additional information

- [ ] I am using Expo
- [ ] I have enabled Frame Processors (react-native-worklets-core)
- [X] I have read the [Troubleshooting Guide](https://react-native-vision-camera.com/docs/guides/troubleshooting)
- [X] I agree to follow this project's [Code of Conduct](https://github.com/mrousavy/react-native-vision-camera/blob/main/CODE_OF_CONDUCT.md)
- [X] I searched for [similar issues in this repository](https://github.com/mrousavy/react-native-vision-camera/issues) and found none.

rodrigobertolottizinsp:
Project is set to run for iOS 13.4 and higher, maybe you are modifying your pods target iOS deployment version in the Podfile?

Romick2005:
> Project is set to run for iOS 13.4 and higher, maybe you are modifying your pods target iOS deployment version in the Podfile? 

I already have that. Latest react native do that for me in:
/node_modules/react-native/scripts/cocoapods/helpers.rb
min_ios_version_supported 13.4

geekabhinavp:
+1 facing same issue!

geekabhinavp:
I fixed it!
Step 1: Check the deployment versions using this command:
```
grep -r 'IPHONEOS_DEPLOYMENT_TARGET' ios/
```
Step 2: If any version lower than 13 is found, run the following:
```
find ios/Pods/Pods.xcodeproj -name project.pbxproj -print0 | xargs -0 sed -i '' 's/IPHONEOS_DEPLOYMENT_TARGET = [0-9]*\.[0-9]*/IPHONEOS_DEPLOYMENT_TARGET = 13.0/g'
```

moigamijunior:
[...]

PR #3694, opened 2026-01-13
fix(ios): add @available(iOS 13.0, *) to UIApplication extension

msageryd:
UIWindowScene and connectedScenes APIs require iOS 13+, but the extension was missing the availability annotation, causing build errors when deployment target is below iOS 13.

## Summary

  Adds missing `@available(iOS 13.0, *)` annotation to `UIApplication+interfaceOrientation.swift`.

  ## Problem

  The extension uses `UIWindowScene` and `connectedScenes` APIs which require iOS 13+, but the extension is missing the availability annotation. This causes build errors:

  'UIWindowScene' is only available in iOS 13.0 or newer

  While the VisionCamera podspec declares iOS 12.4 as minimum, this file uses iOS 13+ APIs without proper guards. When CocoaPods compiles the pod, it uses the pod's declared minimum (12.4), not the app's deployment target, causing the build to fail.

  ## Why this matters

  Users may be running older Xcode versions (e.g., 16.x) because they're not yet ready to migrate to React Native's new architecture and need to stay on older RN versions. These setups are more likely to surface this build error.

  ## Alternative solution

  Instead of adding `@available`, the podspec minimum could be bumped to iOS 13.0. However, adding the annotation is more conservative and maintains backwards compatibility for any iOS
   12 users.

  ## Changes

  - `UIApplication+interfaceOrientation.swift`: Added `@available(iOS 13.0, *)` to the extension

## Related issues

#3648, #3404, #3040

### DIFF
--- a/package/ios/Core/Extensions/UIApplication+interfaceOrientation.swift
+++ b/package/ios/Core/Extensions/UIApplication+interfaceOrientation.swift
@@ -8,6 +8,7 @@
 import Foundation
 import UIKit
 
+@available(iOS 13.0, *)
 extension UIApplication {
   private var windowScene: UIWindowScene? {
     // Get all active scenes


mrousavy:

PR #3704, opened 2026-01-28
fix: Make sure that camera is initialized when querying devices

ldalzottomp:
## What

This PR fixes user reports where android app shows no camera at all by changing how android initialization is done.

## Changes

Note: RNVC = React Native Vision Camera

The root cause was:

* RNVC native module were initializing camera related structure asynchronously on module load. ([code](https://github.com/mrousavy/react-native-vision-camera/blob/main/package/android/src/main/java/com/mrousavy/camera/react/CameraDevicesManager.kt#L64))
* RNVC rely on a initial single call to getConstants() and initialize() to list devices. (next calls to getConstants() are cached by RN) ([code](https://github.com/mrousavy/react-native-vision-camera/blob/main/package/android/src/main/java/com/mrousavy/camera/react/CameraDevicesManager.kt#L107))

This leads to a scenario where:

* The first call to getConstants() is done without camera structures being initialised. Leading the function returning empty as if no devices were detected. Subsequent calls to getConstants() are cached by react-native and always returns the first result. ([code](https://github.com/mrousavy/react-native-vision-camera/blob/main/package/android/src/main/java/com/mrousavy/camera/react/CameraDevicesManager.kt#L91))
* initialize() being called before getConstants() there is still a way that the devices registered are empty due to asynchronicity of the init (see the attached image).

This PR fixes this behaviour and makes sure that camera structure are initialized when demanded by awaiting the initialization task. We keep starting the initialzation as early as possible.

Here is an image of init log with additional lines:

Screenshot 2026-01-30 at 11 31 20

The issue is super easy to reproduce. On any app, introduce an artificial delay before [extensionsManager = ExtensionsManager.getInstanceAsync(reactContext, cameraProvider!!).await(executor)](https://github.com/mrousavy/react-native-vision-camera/blob/main/package/android/src/main/java/com/mrousavy/camera/react/CameraDevicesManager.kt#L69) 

## Tested on

Google Pixel 7 : Android 16.

### DIFF
--- a/package/android/src/main/java/com/mrousavy/camera/react/CameraDevicesManager.kt
+++ b/package/android/src/main/java/com/mrousavy/camera/react/CameraDevicesManager.kt
@@ -13,20 +13,26 @@ import com.facebook.react.bridge.ReadableArray
 import com.facebook.react.modules.core.DeviceEventManagerModule.RCTDeviceEventEmitter
 import com.mrousavy.camera.core.CameraDeviceDetails
 import com.mrousavy.camera.core.CameraQueues
-import com.mrousavy.camera.core.extensions.await
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.asCoroutineDispatcher
-import kotlinx.coroutines.launch
+import java.util.concurrent.Callable
+import java.util.concurrent.Future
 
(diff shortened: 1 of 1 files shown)


mrousavy:

PR #3724, opened 2026-03-24
[ios] Use runtime camera checks on simulator for camera availability

kmagiera:
## What
Replace the unconditional simulator camera block with a runtime `AVCaptureDevice` availability check so simulator fallbacks only apply when no runtime video device exists.

## Changes
- Replace the unconditional `#if targetEnvironment(simulator)` throw in `configureDevice(...)` with a simulator-only runtime check.
- Keep throwing `device/camera-not-available-on-simulator` when `AVCaptureDevice.default(for: .video)` is unavailable.
- Allow the normal camera configuration path when a runtime video device is present.
- Clarify the inline simulator comment to describe the runtime source-availability behavior.

## Tested on
- Run example app on Simulator and iPhone, verified the error is thrown on Simulator and the normal code path works as expected on the device.

## Related issues
- Inspired by [https://github.com/expo/expo/pull/44159](https://github.com/expo/expo/pull/44159).

### DIFF
--- a/package/ios/Core/CameraSession+Configuration.swift
+++ b/package/ios/Core/CameraSession+Configuration.swift
@@ -25,8 +25,10 @@ extension CameraSession {
     videoDeviceInput = nil
 
     #if targetEnvironment(simulator)
-      // iOS Simulators don't have Cameras
-      throw CameraError.device(.notAvailableOnSimulator)
+      // On Simulator, only throw if no runtime video source is available.
+      if AVCaptureDevice.default(for: .video) == nil {
+        throw CameraError.device(.notAvailableOnSimulator)
+      }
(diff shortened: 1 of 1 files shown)


mrousavy:
Thanks @kmagiera for this PR, really appreciate it!

I believe this has now been fixed in VisionCamera V5, as we don't have any special handling for Simulators.

Did you manage to provide a virtual Camera to the Simulator? This is something we can recommend in the docs

hirbod:
@mrousavy yes. https://simcam.swmansion.com/

https://github.com/software-mansion/simcam.app/tree/main/patches/react-native-vision-camera seems to only need a patch for the 4.* branch.

PR #3749, opened 2026-04-20
WIP: test request ios app permission through appium/webdriver

hannojg:


mrousavy:
This Pr will have to be re-opened / re-created since #3793 - I disabled the iOS AWS Device Farm steps for now entirely, and only build iOS for now.

In the future we will want to run iOS AWS Device Farm test once Harness supports granting permissions (CAMERA, MICROPHONE) on a real device (cc @V3RON)

PR #3761, opened 2026-04-21
fix: isActive not working on android

### LINKED ISSUE #3756: `Camera` does not seem to respect `isActive` on V5
Seems like bug template was removed too with the migration.

# Description

`Camera` component does not seem to respect `isActive`, I've set `isActive={false}` and it still displays the preview instead of pausing. This causes `CodeScanner` component on another screen to fail to mount because the `cameraDevice` is still being used by the other screen.

# What I tried

As specified above, I tried setting `isActive={false}` but it does nothing. I also tried unmounting the component when the screen is not focused and it still does not fix the issue, `CodeScanner` displays black.

# Device

Pixel 7a

# VisionCamera Version

5.0.1

# Additional information

I am using Expo.

I can provide any other info if deemed necessary.

riteshshukla04:
fixes https://github.com/mrousavy/react-native-vision-camera/issues/3756


mrousavy:
Let's merge this for now but maybe in the future we can simplify `CustomLifecycle.kt` a bit as it's super hard to understand, compared to iOS' simple start + stop.

Maybe CameraX can also add imperative start + stop methods, that would be awesome.

PR #3774, opened 2026-04-23
feat(recorder): configurable MP4/MOV container via `VideoOutputOptions.fileType`

radko93:
## Summary

Adds an optional `fileType: 'mp4' | 'mov'` to `VideoOutputOptions`, letting callers pick the output container for the entire `CameraVideoOutput`. The default is unchanged (`'mov'` on iOS, `'mp4'` on Android) so this is additive and backward-compatible.

## Motivation

In v4, `startRecording({ fileType: 'mp4' | 'mov' })` let the app choose its output container. V5's `HybridVideoRecorder` hardcodes `AVFileType.quickTimeMovie`, which means every iOS recording now comes out as `.mov` with a `video/quicktime` container. That's fine for on-device playback, but downstream consumers (HTTP upload pipelines, CDNs, FFmpeg/MP4-only tooling) often expect MPEG-4 Part 14. Shipping the v5 upgrade currently forces either:

- uploading `.mov` bytes labelled with a lying `video/mp4` content-type, or
- a local yarn patch flipping `.quickTimeMovie` → `.mp4` in `HybridVideoRecorder.swift`.

Neither is great. The feature is also already hinted at in the codebase — `HybridCameraVideoFrameOutput.swift` had a `// TODO: Allow configuring this` next to its hardcoded `AVFileType.mov`.

## Why output-level, not per-recorder

The persistent recorder path binds `fileType` to the output because codec negotiation (`availableVideoCodecTypesForAssetWriter(writingTo:)`), output settings validation, and `AVAssetWriter` track init all happen at output setup, before any `Recorder` is created. Putting `fileType` on `VideoOutputOptions` keeps both recorder implementations (persistent and non-persistent) on one code path with no branching or preconditions. If callers need a different container for a subsequent recording, they can create a new `CameraVideoOutput`.

## Changes

**Spec**
- `RecorderFileType = 'mp4' | 'mov'`
- `VideoOutputOptions.fileType?: RecorderFileType` (documented as `@platform iOS`, default `'mov'` to preserve current behavior)
- `useVideoOutput(...)` forwards the new option
- Regenerated nitrogen bindings (`bun camera specs`)

**iOS**
- `HybridCameraVideoOutput`: stores `fileType: AVFileType` derived from `options.fileType?.toAVFileType() ?? .quickTimeMovie`, passes it to `HybridVideoRecorder` in `createRecorder`.
- `HybridCameraVideoFrameOutput` (persistent path): replaces the hardcoded `AVFileType.mov` with the option-derived `fileType`. All existing `self.fileType` uses (`availableVideoCodecTypesForAssetWriter(writingTo:)`, `recommendedVideoSettingsForAssetWriter(writingTo:)`, audio-track init) now honor the caller's choice.
- `HybridVideoRecorder`: takes `fileType: AVFileType = .quickTimeMovie` on init and threads it into `URL.createTempURL(fileType:)`. No stored property (only used in init).
- `ios/Extensions/Converters/AV+RecorderFileType.swift`: new extension with `RecorderFileType.toAVFileType()` so both call sites use `options.fileType?.toAVFileType() ?? .quickTimeMovie` instead of a switch.

**Android**
- `HybridVideoOutput.kt#createRecorder` already writes `.mp4` unconditionally — CameraX's `Recorder` doesn't expose a QuickTime output. `fileType` is documented as iOS-only; Android ignores it without throwing to keep cross-platform code simple.

**Docs**
- `docs/content/docs/recorder.mdx` — added a short section under Recorder showing `useVideoOutput({ fileType: 'mp4' })`.

## Usage

```ts
const videoOutput = useVideoOutput({ fileType: 'mp4' })
const recorder = await videoOutput.createRecorder({})
```

## Backward compatibility

No breaking changes. `VideoOutputOptions.fileType` is optional; omitting it preserves the v5.0.x default (`.mov` on iOS, `.mp4` on Android). `RecorderSettings` is untouched aside from its existing `location` field.

## Test plan

- [x] `bun camera typecheck` passes
- [x] `bun camera build` passes
- [x] `bun biome check packages/react-native-vision-camera/src` passes
- [x] Hand-verified `nitrogen` output for `RecorderFileType` and `VideoOutputOptions` on iOS + Android
- [ ] Record a clip with `fileType: 'mp4'` on iOS via `apps/simple-camera`, confirm the produced file is `.mp4` and plays in a strict MP4 player
- [ ] Record a clip without `fileType` on iOS, confirm existing `.mov` default is preserved
- [ ] Record with `enablePersistentRecorder: true` + `fileType: 'mp4'` on iOS, confirm the persistent path honors the container
- [ ] Confirm Android behavior is unchanged

Local typecheck/build/biome are green. On-device verification I'd appreciate guidance on — happy to iterate.


packages/react-native-vision-camera/ios/Hybrid Objects/Outputs/HybridCameraVideoFrameOutput.swift
```diff
@@ -118,6 +118,16 @@ class HybridCameraVideoFrameOutput: HybridCameraVideoOutputSpec, NativeCameraOut
...
+      // Persistent recorders are bound to the output's preconfigured fileType
+      // (used for both codec negotiation and track setup). Customizing
+      // `fileType` per-recorder on this path would require a deeper refactor,
+      // so for now we reject it explicitly instead of silently ignoring it.
+      if settings.fileType != nil {
+        throw RuntimeError.error(
+          withMessage:
+            "`fileType` is not supported on persistent recorders. Disable `enablePersistentRecorder` to configure the container format per-recording."
+        )
+      }
```
mrousavy: That's something I really want to avoid. Branching, special cases, preconditions, etc.

If we want to support this, we should support both persistent and non-persistent. And we 100% can, via `videoSettings`.

packages/react-native-vision-camera/ios/Hybrid Objects/Outputs/HybridCameraVideoOutput.swift
```diff
@@ -106,7 +106,16 @@ class HybridCameraVideoOutput: HybridCameraVideoOutputSpec, NativeCameraOutput {
         let metadataItem = try location.location.toAVMutableMetadataItem()
         self.setMetadataTag(metadataItem)
       }
-      return try HybridVideoRecorder(videoOutput: self.output, queue: self.queue)
+      let fileType: AVFileType
```
mrousavy: Something like 

```swift
let fileType = settings.fileType?.toAVFileType() ?? .quickTimeMovie
```

would be cleaner than a vertically growing `switch`. We can add an Extension for this

packages/react-native-vision-camera/ios/Hybrid Objects/Recording/HybridVideoRecorder.swift
```diff
@@ -12,12 +12,14 @@ class HybridVideoRecorder: HybridRecorderSpec {
   private let videoOutput: AVCaptureMovieFileOutput
   private let queue: DispatchQueue
   private let fileURL: URL
+  private let fileType: AVFileType
```
mrousavy: Do we need this?

packages/react-native-vision-camera/src/specs/outputs/CameraVideoOutput.nitro.ts
```diff
@@ -128,6 +135,25 @@ export interface RecorderSettings {
...
+   *   most HTTP video consumers and CDNs.
+   * - `'mov'`: Apple QuickTime (`.mov`). The historical iOS default; still an
+   *   ISO base media container but advertised with a `video/quicktime` MIME.
+   *
+   * On Android, {@linkcode CameraVideoOutput} always writes `.mp4` regardless
+   * of this setting (CameraX's `Recorder` does not expose a QuickTime output).
+   *
+   * @note Currently only honored for the default (non-persistent) recorder.
+   * Persistent recorders (created when {@linkcode VideoOutputOptions.enablePersistentRecorder}
+   * is `true`) always write `.mov`.
```
mrousavy: Yea this should be gone it should work for both.

mrousavy (changes requested):
Thanks - I really appreciate this, this is good stuff - but we should support the persistent recorder here too to avoid too much branching, which will 100% end up as technical debt.

radko93:
@mrousavy thanks a lot of for quick feedback. I updated it be available via `VideoOutputOptions` so it will work in both scenarios

mrousavy:
Great stuff, thank you @radko93 !

PR #3801, opened 2026-04-26
fix(android): preserve Photo file metadata in memory

Aniketiitk21:
## Summary
- make Android `Photo.getFileData()` and `getFileDataAsync()` reuse the saved-file pipeline so in-memory encoded bytes preserve the same metadata-bearing representation as `saveToFile*()`
- remove the native encoded-bytes cache per review feedback
- clarify the public `Photo.getFileData*()` contract around saved-file parity and EXIF/container metadata

## Testing
- `npm run typecheck`
- `./gradlew.bat :react-native-vision-camera:compileDebugKotlin --no-daemon --console=plain`

### DIFF
--- a/packages/react-native-vision-camera/android/src/main/java/com/margelo/nitro/camera/hybrids/instances/HybridPhoto.kt
+++ b/packages/react-native-vision-camera/android/src/main/java/com/margelo/nitro/camera/hybrids/instances/HybridPhoto.kt
@@ -35,6 +35,7 @@ class HybridPhoto(
   val location: Location?,
 ) : HybridPhotoSpec() {
   private val ioScope = CoroutineScope(Dispatchers.IO)
+  private val encodedFileDataLock = Any()
 
   override val width: Double
     get() = image.width.toDouble()
@@ -71,9 +72,12 @@ class HybridPhoto(
     super.dispose()
     image.close()
     cachedPixelBuffer?.dispose()
+    cachedEncodedFileDataBytes = null
   }
 
   private var cachedPixelBuffer: DisposableArrayBuffer? = null
+  @Volatile
+  private var cachedEncodedFileDataBytes: ByteArray? = null
 
   override fun getPixelBuffer(): ArrayBuffer {
     cachedPixelBuffer?.let {
@@ -85,15 +89,34 @@ class HybridPhoto(
     return pixelBuffer.arrayBuffer
   }
 
+  private fun getRawImageBytes(): ByteArray {
+    return when (image.format) {
+      android.graphics.ImageFormat.JPEG -> {
+        // Always read from a duplicate so callers don't consume ImageProxy state.
+        val buffer = image.planes.single().buffer.duplicate()
+        buffer.rewind()
+        ByteArray(buffer.remaining()).also { bytes -> buffer.get(bytes) }
+      }
+      else -> {
+        throw Error(
+          "Photos with ImageFormat \"${image.format}\" cannot be encoded to bytes " +
+            "until https://issuetracker.google.com/u/3/issues/482079661 is implemented!",
+        )
+      }
+    }
+  }
+
   private fun saveToFile(file: File) {
     when (image.format) {
       android.graphics.ImageFormat.JPEG -> {
-        // JPEG Images have a single plane of image data.
-        val plane = image.planes.single()
-        val buffer = plane.buffer
-        val bytes = ByteArray(buffer.remaining()).also { bytes -> buffer.get(bytes) }
+        cachedEncodedFileDataBytes?.let { encodedBytes ->
+          FileOutputStream(file).use { stream ->
+            stream.write(encodedBytes)
+          }
+          return
+        }
         FileOutputStream(file).use { stream ->
-          stream.write(bytes)
+          stream.write(getRawImageBytes())
         }
         attachExifData(file)
       }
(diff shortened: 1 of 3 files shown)


mrousavy:
Hey - thanks for the PR, but I don't fully understand why we need this?

1. Caching is not required here imo. We shouldn't cache this call. It's explicitly marked as an asynchronous method, callers should worry about caching in JS side.
2. Why didn't wrapping the JPEG plane in an ArrayBuffer work?

Aniketiitk21:
> Hey - thanks for the PR, but I don't fully understand why we need this?
> 
> 1. Caching is not required here imo. We shouldn't cache this call. It's explicitly marked as an asynchronous method, callers should worry about caching in JS side.
> 2. Why didn't wrapping the JPEG plane in an ArrayBuffer work?

Thanks, that makes sense.

You're right about the cache. I added it because the implementation moved away from a direct buffer wrapper to a metadata-preserving file-data path, and I wanted repeated sync/async reads on the same `Photo` instance to stay consistent without depending on `ImageProxy` buffer state. But I agree native caching is not essential here, and JS callers can own caching if they need it.

On the second point: wrapping the JPEG plane directly returns the raw encoded JPEG bytes from the in-memory `ImageProxy`, but on Android, that is not necessarily the same representation as VisionCamera's saved-file path. In `HybridPhoto.saveToFile()`, VisionCamera writes the JPEG bytes and then applies EXIF/location/timestamp/orientation metadata via `attachExifData(file)`. So even though the raw plane is a valid JPEG, it does not go through the same metadata-writing pipeline as `saveToFile*()`.

That is why I routed `getFileData()` through the same save pipeline: to make its result match the documented “what would be written to a file” behaviour. I can revise the PR to preserve that behaviour while removing the cache entirely.

Aniketiitk21:
Pushed a follow-up revision removing the native cache.

The PR now only keeps `getFileData()` / `getFileDataAsync()` aligned with the saved-file representation, and I also clarified the public docs/comments so that contract is explicit.

Aniketiitk21:
Also pushed one small non-behavioral cleanup follow-up:
- renamed the helper to make it explicit that it is dealing with encoded JPEG bytes, not raw pixels
- made temp-file cleanup fall back to `deleteOnExit()` if immediate deletion fails

No contract change from the previous revision; `getFileData()` / `getFileDataAsync()` still stay aligned with the saved-file representation.

mrousavy:
> On the second point: wrapping the JPEG plane directly returns the raw encoded JPEG bytes from the in-memory `ImageProxy`, but on Android, that is not necessarily the same representation as VisionCamera's saved-file path. In `HybridPhoto.saveToFile()`, VisionCamera writes the JPEG bytes and then applies EXIF/location/timestamp/orientation metadata via `attachExifData(file)`. So even though the raw plane is a valid JPEG, it does not go through the same metadata-writing pipeline as `saveToFile*()`.
> 
> That is why I routed `getFileData()` through the same save pipeline: to make its result match the documented “what would be written to a file” behaviour. I can revise the PR to preserve that behaviour while removing the cache entirely.

Ahh, this makes sense, thanks for explaining!

I do think that ideally we want to wait for the Google/CameraX team to add support for a better in-memory Photo type that allows us to represent RAW/DNG, HEIC, ULTRA_HDR, and JPEG photos the same way in-memory, with the metadata/EXIF preserving paths.
I raised this as a feature request in the CameraX issue tracker here: https://issuetracker.google.com/u/3/issues/482079661

And ideally we wait until this is implemented on their end, as that makes this entire implementation much simpler. Right now, `Photo` is unfortunately partially broken on Android anyways (it only works for JPEG, not for ULTRA_HDR or RAW).

packages/react-native-vision-camera/android/src/main/java/com/margelo/nitro/camera/hybrids/instances/HybridPhoto.kt
```diff
@@ -125,26 +138,20 @@ class HybridPhoto(
...
-        //       just like on iOS via `AVCapturePhoto.fileDataRepresentation()` - no matter the format.
-        throw Error(
-          "Cannot get File Data for Photos with Image Format \"${image.format}\" " +
-            "until https://issuetracker.google.com/u/3/issues/482079661 is implemented!",
-        )
+    val tempFile = File.createTempFile("VisionCamera_", containerFormat.fileExtension)
+    try {
+      // Reuse the same save pipeline so JS receives the same EXIF/location flags as a saved file.
+      saveToFile(tempFile)
+      return ArrayBuffer.copy(tempFile.readBytes())
```
mrousavy, line 145: This is definitely not what we want anyways. This entirely defeats the purpose of an in-memory type alltogether.

Aniketiitk21:
Closing this for now.

After your feedback and a deeper pass through the Android implementation, I agree the current temp-file roundtrip is the wrong tradeoff for an in-memory `Photo` API. It improves JPEG parity with the saved-file path today, but it does so by breaking the spirit of an in-memory representation, and it still does not solve the broader Android gaps for ULTRA_HDR/RAW.

That makes this feel more like a workaround for a deeper CameraX limitation than a clean VisionCamera fix. Since a proper solution depends on upstream support for metadata-preserving in-memory/file representations across formats, I don't think this implementation belongs in VisionCamera in its current form.

Thanks for the review and for filing the upstream CameraX issue.

mrousavy:
Sounds good! Hopefully CameraX adds that soon

PR #3802, opened 2026-04-26
feat: expose supported photo container formats

Aniketiitk21:
## Summary
- expose `CameraDevice.supportedPhotoContainerFormats` so apps can choose a valid photo container format before capture
- implement platform-aware format discovery on Android and iOS and wire the new property through Nitro-generated bindings
- log the new capability in the simple camera example and document `PhotoOutputOptions.containerFormat` against the new property

## Testing
- `cd packages/react-native-vision-camera && tsc --noEmit false && nitrogen`
- `cd apps/simple-camera/android && ./gradlew.bat :react-native-vision-camera:compileDebugKotlin --no-daemon --console=plain`

## Notes
- iOS build was not run in this environment

### DIFF
--- a/packages/react-native-vision-camera/ios/Hybrid Objects/Inputs/HybridCameraDevice.swift
+++ b/packages/react-native-vision-camera/ios/Hybrid Objects/Inputs/HybridCameraDevice.swift
@@ -24,6 +24,9 @@ final class HybridCameraDevice: HybridCameraDeviceSpec, NativeCameraDevice {
       .map { PixelFormat(mediaSubType: $0.formatDescription.mediaSubType) }
       .withoutDuplicates()
   }()
+  lazy var supportedPhotoContainerFormats: [PhotoContainerFormat] = {
+    return getSupportedPhotoContainerFormats()
+  }()
 
   func getSupportedResolutions(outputStreamType: OutputStreamType) -> [Size] {
     let streamType = streamType(from: outputStreamType)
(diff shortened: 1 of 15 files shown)


packages/react-native-vision-camera/ios/Hybrid Objects/Inputs/HybridCameraDevice.swift
```diff
@@ -45,6 +48,43 @@ final class HybridCameraDevice: HybridCameraDeviceSpec, NativeCameraDevice {
...
+      let rawFormats = output.availableRawPhotoFileTypes
+        .map { PhotoContainerFormat(avFileType: $0) }
+        .filter { $0 == .dng }
+      let formats = (processedFormats + rawFormats).withoutDuplicates()
+      return formats.isEmpty ? [.jpeg] : formats
+    } catch {
+      logger.error("Failed to determine supported photo container formats: \(error)")
+      return [.jpeg]
+    }
+  }
```
mrousavy, line 86: I think this part here is _slow_. 

It is exposed via JS as a property access, but internally it may take many milliseconds, possibly up to a second, to complete. That would be a terrible API for users if they do any kind of filtering. 

We'd have to benchmark this code to see how bad it actually is, but it definitely doesn't look fast at all. It's literally a hardware configuration.

mrousavy (commented):
Thanks for the PR, this is the same code that Claude keeps hallucinating for me all the time as well.

It's imo terrible API design by Apple to have available photo container formats exposed only once you actually _connect_ the output to the camera, that makes the API that we want here kinda impossible to achieve as Android expects the Photo Container Format to be known upfront when creating the output.

That's the reason why it didn't exist before, I wasn't just lazy.

But let's keep this PR open, maybe we can benchmark the code to find out how bad it actually is.

Aniketiitk21:
> Thanks for the PR, this is the same code that Claude keeps hallucinating for me all the time as well.
> 
> It's imo terrible API design by Apple to have available photo container formats exposed only once you actually _connect_ the output to the camera, that makes the API that we want here kinda impossible to achieve as Android expects the Photo Container Format to be known upfront when creating the output.
> 
> That's the reason why it didn't exist before, I wasn't just lazy.
> 
> But let's keep this PR open, maybe we can benchmark the code to find out how bad it actually is.

Thanks, that makes sense, and thanks for clarifying the background here.

I used a property because the caller needs this information before choosing `PhotoOutputOptions.containerFormat`, but I agree the iOS path makes that API shape risky if it requires real session/input/output wiring under the hood. Even if Android can answer this cheaply, the cross-platform contract still has to be safe on iOS.

I definitely do not want to ship a synchronous property that can block JS for a noticeable amount of time just to interrogate hardware. Benchmarking the iOS implementation is the right next step.

If the benchmark shows this is non-trivial, I agree this API should either stay out of the public surface or be redesigned around an explicit async flow rather than a property.
Do you have any Ideas to improve this?

mrousavy:
I think we have two options;

- A) Continue this road, if its super trivial and fast to compute (<3ms) then we can keep this a sync property (`lazy` cache).
- B) Make Photo Container Format a **Constraint** instead.

I feel like because of the restrictions on iOS, a Constraint might be the only solution that is not doing false promises.
We dont know what Apple internally does, maybe it has a reason to only allow RAW/DNG capture if there is no Movie Output attached. Then if we use this assumption to create a session graph, it might fail.

So it kinda sucks because it seems like it'd be related to the Photo Output, but then its actually a session wide configuration....

Aniketiitk21:
Closing this version for now.

After reviewing the iOS path more carefully and considering your feedback, I agree that exposing this as a synchronous `CameraDevice` property is the wrong abstraction unless it can be proven trivially cheap. The more honest long-term direction seems to be either:

- benchmark it and prove that the property shape is safe, or
- redesign this around a session-level `Constraint` instead of a device property

Since that would materially change both the API and the implementation, I'd rather close this property-based proposal than keep pushing the current shape further.

Thanks for the context here - it clarified why this was not already exposed.

PR #3808, opened 2026-04-28
fix(ios): barcode format filter ignored in react-native-vision-camera-barcode-scanner due to discarded union result

gorbypark:
The iOS MLKit barcode format was always empty because OptionSet.union() returns a new value rather than mutating in place, so the result was discarded on every iteration. Replaced with formUnion() so the format filter is actually applied.

### DIFF
--- a/packages/react-native-vision-camera-barcode-scanner/ios/Extensions/ML+BarcodeScannerOptions.swift
+++ b/packages/react-native-vision-camera-barcode-scanner/ios/Extensions/ML+BarcodeScannerOptions.swift
@@ -13,7 +13,7 @@ extension BarcodeScannerOptions {
     // combines the array into an OptionSet/bitmask
     let combinedFormats = formats.reduce(into: MLKitBarcodeScanning.BarcodeFormat()) {
       partial, next in
-      _ = partial.union(next)
+      partial.formUnion(next)
     }
     return MLKitBarcodeScanning.BarcodeScannerOptions(formats: combinedFormats)
   }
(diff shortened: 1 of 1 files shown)


gorbypark:
Here's a patch package patch if it makes testing easier and/or if anyone wants to use this before this is merged.
`react-native-vision-camera-barcode-scanner+5.0.6.patch`
```patch
diff --git a/node_modules/react-native-vision-camera-barcode-scanner/ios/Extensions/ML+BarcodeScannerOptions.swift b/node_modules/react-native-vision-camera-barcode-scanner/ios/Extensions/ML+BarcodeScannerOptions.swift
index e8f7aeb..56b976d 100644
--- a/node_modules/react-native-vision-camera-barcode-scanner/ios/Extensions/ML+BarcodeScannerOptions.swift
+++ b/node_modules/react-native-vision-camera-barcode-scanner/ios/Extensions/ML+BarcodeScannerOptions.swift
@@ -13,7 +13,7 @@ extension BarcodeScannerOptions {
     // combines the array into an OptionSet/bitmask
     let combinedFormats = formats.reduce(into: MLKitBarcodeScanning.BarcodeFormat()) {
       partial, next in
-      _ = partial.union(next)
+      partial.formUnion(next)
     }
     return MLKitBarcodeScanning.BarcodeScannerOptions(formats: combinedFormats)
   }
@@ -25,7 +25,7 @@ extension BarcodeScannerOutputOptions {
     // combines the array into an OptionSet/bitmask
     let combinedFormats = formats.reduce(into: MLKitBarcodeScanning.BarcodeFormat()) {
       partial, next in
-      _ = partial.union(next)
+      partial.formUnion(next)
     }
     return MLKitBarcodeScanning.BarcodeScannerOptions(formats: combinedFormats)
   }
```

mrousavy:
ahh good catch thank you!

PR #3825, opened 2026-04-29
fix(android): annotate classes with @Keep + @DoNotStrip to prevent RC8 stripping them

DaveyEke:


packages/react-native-vision-camera/android/src/main/java/com/margelo/nitro/camera/hybrids/inputs/HybridCameraDevice.kt
```diff
@@ -58,6 +60,8 @@ import com.margelo.nitro.camera.utils.ImageFormatUtils
  * APIs if necessary.
  */
 @SuppressLint("RestrictedApi")
+@Keep
+@DoNotStrip
```
mrousavy, line 64: This is not needed since the object is not autolinked - it would be unnecessary.

packages/react-native-vision-camera/android/src/main/java/com/margelo/nitro/camera/hybrids/inputs/HybridPhysicalCameraDevice.kt
```diff
@@ -35,6 +37,8 @@ import com.margelo.nitro.camera.public.NativeCameraDevice
  * This type has less available information than the combined [CameraInfo].
  */
 @SuppressLint("RestrictedApi")
+@Keep
+@DoNotStrip
```
mrousavy, line 41: same here

packages/react-native-vision-camera/android/src/main/java/com/margelo/nitro/camera/hybrids/inputs/HybridCameraExtension.kt
```diff
@@ -1,14 +1,18 @@
...
 import androidx.camera.extensions.ExtensionMode
 import androidx.camera.extensions.ExtensionsManager
+import com.facebook.proguard.annotations.DoNotStrip
 import com.margelo.nitro.camera.CameraExtensionType
 import com.margelo.nitro.camera.HybridCameraExtensionSpec
 import com.margelo.nitro.camera.extensions.converters.fromExtensionMode
 import com.margelo.nitro.camera.public.NativeCameraExtension
 
+@Keep
+@DoNotStrip
```
mrousavy, line 15: same here

mrousavy:
I think #3835 fixed this now, but let me know if there still is a crash in release

PR #3827, opened 2026-04-30
fix(ios): Crash when taking a photo on iOS 15

Developerius:
This PR fixes a crash that occurs on iOS 15 when attempting to capture a photo.

For iOS versions earlier than 16, AVCapturePhotoSettings.isHighResolutionPhotoEnabled is explicitly set. However, this flag requires isHighResolutionCaptureEnabled = true to be configured on the corresponding AVCapturePhotoOutput. Without this, the app crashes at runtime.


mrousavy:
Thanks for your PR, but should this property really always be `true`? Shouldn't we either leave it `false` by default, or only set it to `true` if a certain `targetSize` is configured? e.g. if its >=4k, or if quality prioritization is `quality` then we can enable this flag? 

Honestly after writing this I feel like it might be best to just leave it untouched by default since iOS <15 is such a small market share nowadays and we can just drop this for a smaller test matrix...

mrousavy:
I merged https://github.com/mrousavy/react-native-vision-camera/pull/3836 now

PR #3848, opened 2026-05-09
ci(ios): fix harness test for ios (permissions should work now)

hannojg:
The new pipeline looks like this: 

Screenshot 2026-05-12 at 14 02 20

Android and iOS now work differently in terms of test execution on AWS, which also means that there is no shared test package between the two platforms (as it was before).

For iOS we now have to build a unsigned XCTest artifact. To AWS we now upload:
- our native app, ipa, unsigned
- the XCTest ipa, unsigned
- Our node project containing the harness tests

On upload AWS will sign the ipa artifacts for us so that they can be installed on AWS's iPhones.
In order to upload and sign an XCTest artifact we had to switch the AWS test type for iOS to `XCTEST_UI`. Luckily though, that changes nothing about the connected host runner so we can still run our nodeJS stuff for harness!

For harness we applied the following PRs as patches:

- PR for building XCTest standalone:
  - https://github.com/callstackincubator/react-native-harness/pull/117
- PR for injecting an external XCTest artifact:
  - https://github.com/callstackincubator/react-native-harness/pull/116


hannojg:
komm merge mein kind

hannojg:
mergieren

mrousavy:
sohn

PR #3878, opened 2026-05-18
Demo of point mapping issue

JimPatterson:
This is a followup to #3871.  I've tested this again with the changes in #3872 and #3874 and I'm still seeing the same problem.

To ensure I'm using the latest code, I moved the demo code into simple-camera app.  I added an icon in the upper left corner to launch the barcode scanner.   It seems to work correctly when the device is held in landscape, but not in portrait.  So maybe is a rotation problem?

An example screenshot of the current problem is Screenshot_20260517-211519

### DIFF
--- a/apps/simple-camera/src/App.tsx
+++ b/apps/simple-camera/src/App.tsx
@@ -5,6 +5,7 @@ import {
 import { createNativeStackNavigator } from '@react-navigation/native-stack'
 import { GestureHandlerRootView } from 'react-native-gesture-handler'
 import { VisionCamera } from 'react-native-vision-camera'
+import { BarcodeScreen } from './screens/BarcodeScreen'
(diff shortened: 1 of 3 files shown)


mrousavy:
Thanks for the repro!

mrousavy:
Does this work on iOS?

JimPatterson:
I don't have a mac, so testing iOS is difficult.

mrousavy:
I managed to reproduce this in our Harness tests via https://github.com/mrousavy/react-native-vision-camera/pull/3896.

Closing this PR now since we need Harness to test such things.

It's currently red on Android and green on iOS.

Maybe quickly take a look at the test and confirm if this is the same issue you are seeing?

I will attempt to fix it soon now that we have it in our tests.

JimPatterson:
Yes, that test looks like it is the right idea.

PR #3887, opened 2026-05-19
perf: Provide a fast path for `Array.withoutDuplicates()` if `Element` is `Hashable`

mrousavy:
This uses `Set` / `Hashable` for faster deduping


mrousavy:
unnecessary over optimization imo

PR #3906, opened 2026-05-21
test(controller): failing harness for setTorchMode supersede race

Ryudo3:


mrousavy:
Thanks for adding the test! See my reply in https://github.com/mrousavy/react-native-vision-camera/pull/3908 and the related issue, this has been somewhat fixed but we need to call `start()` before using the controller.

PR #3908, opened 2026-05-21
test(controller): add failing repro for setZoom-before-start race

Jakef14:
`useZoomUpdater` (src/hooks/internal/useZoomUpdater.ts:16) fires controller.setZoom synchronously on the <Camera> component's first mount, before the consumer's `isActive` toggle has caused the session to start. The returned promise is discarded, so on Android the CameraX `OperationCanceledException: Camera is not active` rejection surfaces as an Unhandled Promise Rejection that consumers cannot intercept. <Camera onError> only receives CameraRuntimeError instances.

This adds a failing harness test that reproduces the timing imperatively: configure() the session, then call setZoom before start() — the moment useZoomUpdater hits today.

### DIFF
--- a/apps/simple-camera/__tests__/visioncamera.controller.harness.ts
+++ b/apps/simple-camera/__tests__/visioncamera.controller.harness.ts
@@ -53,6 +53,50 @@ describe('VisionCamera - Controller', () => {
     }
   })
 
+  it('setZoom called between configure() and start() does not reject', async () => {
+    // Reproduces an unhandled-rejection bug surfaced by `<Camera>`'s
+    // useZoomUpdater hook on Android.
+    //
+    // useZoomUpdater (src/hooks/internal/useZoomUpdater.ts:16) calls
+    // `controller.setZoom(zoom)` synchronously inside a useEffect that runs on
+    // first mount, before the user's `<Camera isActive>` toggle has caused the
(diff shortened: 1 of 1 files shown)


apps/simple-camera/__tests__/visioncamera.controller.harness.ts
```diff
@@ -53,6 +53,50 @@ describe('VisionCamera - Controller', () => {
...
+    ])
+    if (controller == null) throw new Error('no controller')
+
+    // setZoom issued before start(); CameraX has not reached the active
+    // state yet, so this is the moment the bug fires.
+    await expect(
+      controller.setZoom(controller.minZoom),
+    ).resolves.not.toThrow()
+
+    await session.start()
```
mrousavy, line 96: This is what I meant in https://github.com/mrousavy/react-native-vision-camera/issues/3909#issuecomment-4518499578 - the `controller` is not yet usable here. You'd have to call `await session.start()` before doing `controller.setZoom(...)`.

This is just how CameraX works unfortunately - on iOS I think you can already use the controller. But as a general best practice, just use the `CameraController` after `start()` has resolved.

mrousavy:
Thanks for the repro, but closing as mentioned in https://github.com/mrousavy/react-native-vision-camera/issues/3909#issuecomment-4518499578.

PR #3910, opened 2026-05-21
Added MlKit Text Recognition

fracico-tech:
Hi @mrousavy,

I wanted to implement a text recognition frame processor plugin for v5 (similar to what I had for v4). But after seeing your MLKit-based barcode scanner implementation, I realized the most efficient approach would be to build directly on top of that architecture to add MLKit Text Recognition.

I understand you might want to keep the package lean, so I'm completely open to how we ship this. If you're worried about the maintenance burden, we could flag this feature as "community-driven" or experimental.

Current Status:

Android: Tested and working perfectly.

iOS: Tested and working perfectly.

If adding text recognition doesn't align with the roadmap for this repository, please let me know—I'd still love to hear your thoughts on the best way to handle it!

Thanks for the amazing package!


fracico-tech:
Fixed also iOS orientation

mrousavy:
Hey - thanks for the PR, but this should be a third party library - not something that belongs into VisionCamera core.

So feel free to publish this in a separate github repo.
If you still want me to review your code let me know, I can give some suggestions

PR #3918, opened 2026-05-27
fix: Fix coordinate system conversions

JimPatterson:
This addresses the android side of #3871.   This has been tested on a real android device using the code in #3878 and it works correctly for all orientations.  But I don't have a mac or an iPhone and can't test iOS.

This was coded by an AI agent.  It said:

`convertFramePointToCameraPoint` and `convertCameraPointToFramePoint` on Android were using `sensorToBufferTransformMatrix` directly without accounting for the frame's orientation. That matrix handles sensor crop/zoom but carries no rotation — rotation is separate metadata in `rotationDegrees`. As a result, passing oriented-space coordinates (as returned by ML analyzers) produced silently wrong camera coordinates for any non-upright frame.

The fix mirrors iOS's `FrameCoordinateSystemConverter`: before applying the inverted sensor matrix, un-rotate the input point from oriented space to raw buffer space; when converting the other direction, apply the forward rotation to the buffer-space result. The same fix is applied to `HybridDepthFrame`.

The two coordinate harness tests are updated to match the corrected API:
- The round-trip test now passes the oriented-space center `(h/2, w/2)` for rotated frames rather than the raw buffer center `(w/2, h/2)`, since `convertFramePointToCameraPoint` now expects oriented coordinates.
- The oriented-rectangle test is redesigned as a true round-trip check (oriented → camera → oriented) instead of comparing two paths that diverge after the API change.

### DIFF
--- a/apps/simple-camera/__tests__/visioncamera.coordinates.harness.tsx
+++ b/apps/simple-camera/__tests__/visioncamera.coordinates.harness.tsx
@@ -490,7 +490,15 @@ describe('VisionCamera - Coordinates', () => {
     const runtime = workletsProvider.createRuntimeForThread(frameOutput.thread)
     runtime.setOnFrameCallback(frameOutput, (frame) => {
       'worklet'
-      const center = { x: frame.width / 2, y: frame.height / 2 }
+      const w = frame.width
+      const h = frame.height
+      const isRotated =
+        frame.orientation === 'left' || frame.orientation === 'right'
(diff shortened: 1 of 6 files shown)


JimPatterson:
I've updated this with the iOS fixes also.  Using a simple demo app built on both android and iOS real devices, it now correctly draws the box over the barcode on the screen.

Is there anything else I can do to help get this reviewed and merged?

PR #3956, opened 2026-05-27
docs: Hide empty API table of contents

mrousavy:


mrousavy:
there's missing spacing above the "Let's talk" widget - we need some spacing here maybe we can align with the spacing in the center content 

image

PR #3960, opened 2026-05-28
fix: Resolve iOS Metro bundle URL

mrousavy:
## Summary
- persist the Metro port from the iOS build into Info.plist
- default simulator Debug launches to localhost:<RCTMetroPort> so react-native run-ios --port uses the correct Metro server with RNCore prebuilt
- refresh the React-Core-prebuilt CocoaPods checksum generated by pod install

## Verification
- bun example pods
- bun example ios -- --udid 109A1C99-34FC-4CB6-8562-934E6AEA05DC --port 8082
- adb reverse tcp:8082 tcp:8082 && bun example android -- --port 8082
- bun lint-swift


mrousavy:
Closing this. I re-checked latest main with a clean Metro cache and the React Native patch applied; the AppDelegate/Info.plist changes are not needed. The actual reproducible issue was only the generated React-Core-prebuilt checksum in Podfile.lock being out of sync after pod install. Replaced by #3961.

PR #3973, opened 2026-06-01
fix: Add timeout for zoom ramp completion

mrousavy:


mrousavy:
this is fixing a symptom, not a root cause. superseded by #3988

PR #3986, opened 2026-06-02
fix: Add timeout to iOS startZoomAnimation

mrousavy:
## What changed

This PR is scoped to iOS `HybridCameraController.startZoomAnimation()`:

- adds a bounded timeout while waiting for `isRampingVideoZoom` to return to `false`, so a zoom ramp cannot leave the JS Harness test file waiting until the bridge-level timeout
- requests `.old` and `.new` KVO values for `isRampingVideoZoom`, because the completion logic checks for the `true -> false` transition

## Scope

This does not change Harness timeout settings, test expectations, Android CameraX behavior, or Device Farm result handling. If Device Farm or Harness reports a failure, CI should still stay red.

This also is not intended to explain every controller timeout. It is a narrow iOS guard/fix for zoom-ramp completion in `startZoomAnimation()`.


mrousavy:
Combined validation result for run `26828916629`:

- iOS `Test iOS` is green even though raw Device Farm reported `FAILED`, so the collector fix is doing the intended thing: Harness output wins when the Harness summary is green.
- iOS Harness summary: `11 passed, 11 total`; `122 passed`; controller passed. Controller timings: exposure-bias `791 ms`, focus/reset `1745 ms`.
- Android `visioncamera.controller.harness.ts` also passed. Controller timings: exposure-bias `2224 ms`, focus/reset `5330 ms`.
- Android check remains red because Harness reported real failures in `photo`, `coordinates`, and `video`; the collector correctly failed on those real Harness/JUnit failures.

So the controller-file-not-running symptom is fixed on this combined main-based state. Remaining Android failures are separate test/product assertions, not Harness/controller startup or bridge timeout.

mrousavy:
this is fixing a symptom, not a root cause. superseded by #3988

PR #3988, opened 2026-06-02
fix: Don't start zoom animation if we are already on that exact `zoom` value

mrousavy:
Maybe a better fix than https://github.com/mrousavy/react-native-vision-camera/pull/3986 ?


packages/react-native-vision-camera/ios/Hybrid Objects/HybridCameraController.swift
```diff
@@ -321,9 +321,16 @@ final class HybridCameraController: HybridCameraControllerSpec, NativeCameraCont
...
+      if self.captureDevice.videoZoomFactor == zoom {
+        // We are already exactly on our target value
+        completion()
+        return
+      }
+
       var observation: NSKeyValueObservation?
       observation = self.captureDevice.observe(
         \.isRampingVideoZoom,
+        options: [.old, .new],
```
mrousavy, line 333: i think this is actually the real fix and was a bad oversight from my end

PR #3992, opened 2026-06-03
fix: wait for Android exposure capture readiness

mrousavy:
## Summary

This keeps Android `CameraSession.start()` scoped to CameraX camera-state readiness, and moves capture-result readiness to the controller operation that actually needs it: supported, in-range, state-changing `setExposureBias(...)` calls.

It also keeps the Device Farm/Harness CI hardening from earlier commits:

- package Bun into Device Farm test packages instead of relying on network install inside Device Farm
- print Device Farm diagnostics when Harness artifacts are missing

## Why

The Harness diagnostic run for `visioncamera.controller.harness.ts` showed the Android bridge timeout happened while the controller test was running:

```text
Last started test before the device stopped responding: VisionCamera - Controller sets exposure bias to min/max when the device supports it
```

CameraX exposure compensation futures complete from capture results. Waiting only for `CameraState.OPEN` is not always strong enough: the camera can be open while the capture-result pipeline has not yet produced the metadata CameraX uses to complete `setExposureCompensationIndex(...)`.

A previous attempt waited for the first capture result in every `CameraSession.start()`. That made the controller suite pass once, but it was too broad: valid multi-output starts can have no consumer attached yet and should not fail just because no capture result is produced at start. This revision removes that global start wait and only fences the exposure-bias path that the logs identified.

If the camera is open but no capture result arrives before applying exposure bias, the controller now fails after 5s with a specific native error instead of letting the Harness file-level RPC timeout burn roughly 120s. Invalid or no-op exposure-bias calls skip the readiness wait and still go through CameraX directly.

## Verification

- `bun run lint-kotlin -- packages/react-native-vision-camera/android/src/main/java/com/margelo/nitro/camera/hybrids/HybridCameraController.kt packages/react-native-vision-camera/android/src/main/java/com/margelo/nitro/camera/hybrids/HybridCameraSession.kt`
- `cd apps/simple-camera/android && JAVA_HOME=/opt/homebrew/opt/openjdk@17 ./gradlew :react-native-vision-camera:compileDebugKotlin --no-daemon --console=plain`

Device Farm is running again on `34b52f0c` to verify that the Android controller suite still runs and the multi-output regression from the global start wait is gone.

### DIFF
--- a/.github/actions/collect-devicefarm-results/action.yml
+++ b/.github/actions/collect-devicefarm-results/action.yml
@@ -5,9 +5,18 @@ inputs:
   artifact-folder:
     description: Device Farm artifact folder produced by the test action
     required: true
+  aws-region:
+    description: AWS region where the Device Farm run was scheduled
+    required: false
+  console-url:
+    description: AWS Device Farm console URL for the run
+    required: false
   platform:
     description: Platform label used in filenames and log output
     required: true
+  run-arn:
+    description: AWS Device Farm run ARN
+    required: false
   test-step-outcome:
     description: Outcome of the upstream Device Farm execution step
     required: false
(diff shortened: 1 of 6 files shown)


mrousavy:
i think this is BS, CameraX should allow us operating with the Camera after `start()` completed I think

PR #3993, opened 2026-06-03
fix(android): align device orientation degrees

ludwig-pro:
## What

Align Android device-orientation degree mapping with the physical `CameraOrientation` semantics used by iOS.

| 0 | -90° | 90° | 180 |
|--------|--------|--------|--------|
| Portrait | Left | Right | Down |

`OrientationEventListener` reports physical device rotation degrees, but `HybridDeviceOrientationManager` previously reused `CameraOrientation.fromDegrees(...)`, which maps 90°/270° for buffer/image orientation semantics.

This caused Android `createOrientationManager('device')` to report `left`/`right` mirrored compared to iOS for the same physical device pose.

This PR now applies the existing image/frame orientation mapper and counter-rotates it directly at the Android device-orientation callsite with `fromDegrees(degrees).counterRotated()`, leaving the existing `fromDegrees(...)` behavior unchanged for image/frame paths.

## Test plan

- Installed and launched the example app on a physical Pixel 9a; confirmed `useOrientation('device')` reports the right orientation
- Temporarily validated the mapping with the local JUnit guard before removing JUnit from the final PR per review feedback
- Verified no JUnit leftovers with `rg -n "junit|CameraOrientationDeviceRotationDegreesTest" packages/react-native-vision-camera/android`
- `cd apps/simple-camera/android && ./gradlew :react-native-vision-camera:compileDebugKotlin --console=plain`

### Before
android-before-small

### After
android-after-small


packages/react-native-vision-camera/android/src/test/java/com/margelo/nitro/camera/extensions/CameraOrientationDeviceRotationDegreesTest.kt
```diff
@@ -0,0 +1,33 @@
+package com.margelo.nitro.camera.extensions
+
+import com.margelo.nitro.camera.CameraOrientation
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+class CameraOrientationDeviceRotationDegreesTest {
+  @Test
```
mrousavy: We don't have any JUnit tests - please remove those.

packages/react-native-vision-camera/android/build.gradle
```diff
@@ -169,6 +169,8 @@ dependencies {
 
   // Add a dependency on NitroImage
   implementation project(":react-native-nitro-image")
+
+  testImplementation "junit:junit:4.13.2"
```
mrousavy: Nope. We use Harness for testing, mentioned in the README.md in __tests__.

packages/react-native-vision-camera/android/src/main/java/com/margelo/nitro/camera/extensions/CameraOrientation+degrees.kt
```diff
@@ -22,6 +22,16 @@ fun CameraOrientation.Companion.fromDegrees(degrees: Int): CameraOrientation {
...
 
+fun CameraOrientation.Companion.fromDeviceRotationDegrees(degrees: Int): CameraOrientation {
+  val normalizedDegrees = normalizeDegrees(degrees)
+  return when (normalizedDegrees) {
+    in 45..135 -> CameraOrientation.RIGHT
+    in 135..225 -> CameraOrientation.DOWN
+    in 225..315 -> CameraOrientation.LEFT
+    else -> CameraOrientation.UP
+  }
+}
```
mrousavy: Ok so it's the same as `fromDegrees` but inverted for left & right

mrousavy (changes requested):
Hey - thanks for your PR!

This is much appreciated if it works, but we do testing differently. We don't test with JUnit (and adding the dependency here is not the way to go) - we test via full E2E testing on a real device via react-native-harness.

Please tell your agent to remove the JUnit tests.
Then, please add react-native-harness tests (vitest syntax). We can probably not add tests for device orientation changes as the device orientation is fixed in AWS, so if that cannot be tested with Harness then I guess there's no point in adding tests for that - then just the code is fine too.

packages/react-native-vision-camera/android/src/main/java/com/margelo/nitro/camera/extensions/CameraOrientation+degrees.kt
ludwig-pro: Yep, simplified to reuse `fromDegrees(degrees).counterRotated()` 🫡

packages/react-native-vision-camera/android/src/test/java/com/margelo/nitro/camera/extensions/CameraOrientationDeviceRotationDegreesTest.kt
ludwig-pro: Removed 🫡

packages/react-native-vision-camera/android/build.gradle
ludwig-pro: Removed 🫡

packages/react-native-vision-camera/android/src/main/java/com/margelo/nitro/camera/extensions/CameraOrientation+degrees.kt
```diff
@@ -22,6 +22,10 @@ fun CameraOrientation.Companion.fromDegrees(degrees: Int): CameraOrientation {
   }
 }
 
+fun CameraOrientation.Companion.fromDeviceRotationDegrees(degrees: Int): CameraOrientation {
+  return CameraOrientation.fromDegrees(degrees).counterRotated()
```
mrousavy: But is it even correct for interface orientation then? Do we also counterRotate that or no

packages/react-native-vision-camera/android/src/main/java/com/margelo/nitro/camera/extensions/CameraOrientation+degrees.kt
ludwig-pro: Inlined it now to keep the diff smaller: `fromDegrees(rotationDegrees).counterRotated()` 🫡

packages/react-native-vision-camera/android/src/main/java/com/margelo/nitro/camera/extensions/CameraOrientation+degrees.kt
ludwig-pro: Yeah, `device` and `interface` are a bit counter-intuitive because they don’t describe the same thing:
- `device` describes how the phone is physically rotated, based on `OrientationEventListener`.
- `interface` describes how the UI/display is rotated, based on `Display.rotation` / `Surface.ROTATION_*`.

So when the phone is physically rotated -90°, `device` reports `left`, but the UI needs to rotate +90° to stay readable, so `interface` reports `right`.

That’s why I think we should only apply `counterRotated()` on the device-orientation path, and leave interface orientation as-is.

ludwig-pro: Do you think it’s worth adding a small note to the docs?

Basically: `device` is the physical rotation from `OrientationEventListener`, while `interface` is the UI/display rotation from `Display.rotation` / `Surface.ROTATION_*`, so they can be opposite when the UI counter-rotates to stay readable ?

ludwig-pro: screen-20260604-144429-1780577057012-under-10mb

PR #3998, opened 2026-06-05
[codex] stabilize iOS Harness Device Farm CI

mrousavy:
## What changed

- Package the GitHub runner's Bun binary into the iOS XCTest UI test package so Device Farm does not depend on downloading Bun during the install phase.
- Export Harness XCTest logs from the iOS Device Farm run and teach the result collector to copy `xcodebuild.log` alongside Harness JUnit/log artifacts.
- Print Device Farm job/suite/test diagnostics when the run step fails or expected Harness artifacts are missing.
- Treat iOS Harness JUnit results as authoritative only when JUnit publication succeeds and the XCTest log contains `** TEST EXECUTE SUCCEEDED **`. Android remains strict on the Device Farm step outcome.
- Include `.github/actions/collect-devicefarm-results/**` in the workflow trigger/change detector so collector changes run the workflow they affect.

## Why

The latest `main` iOS Device Farm run failed with `ERRORED`, `Total: 0`, and no Harness artifacts. Existing diagnostic runs showed that packaging Bun gets Device Farm past that setup failure and all Harness iOS tests then pass, but Device Farm still reports its generic XCTEST_UI `Tests` lifecycle as failed even while Harness JUnit and the XCTest log report success.

The pinned Harness Apple package (`@react-native-harness/platform-apple@1.4.0-rc.1`) tears down the XCTest agent process with signals during disposal. That can poison AWS Device Farm's wrapper-level result even when the actual Harness/Jest tests and XCTest runner succeeded. Until Harness exposes a clean upstream shutdown/exit contract, CI should fail on the actual test artifacts rather than on AWS's wrapper lifecycle false negative.

## Validation

- `git diff --check`
- Ruby YAML load for `.github/workflows/harness-aws-device.yml`, `.github/actions/collect-devicefarm-results/action.yml`, and `apps/simple-camera/device-farm-tests/AwsTestSpecIOS.yml`
- Local artifact lookup simulation for direct and zipped Device Farm artifacts (`harness-results.junit.xml`, `harness-output.log`, `xcodebuild.log`)
- Inspected the latest failing `main` iOS run and previous diagnostic iOS runs: Harness reported all tests passing while Device Farm reported wrapper failure.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+patches/*.patch -whitespace
--- a/.github/actions/collect-devicefarm-results/action.yml
+++ b/.github/actions/collect-devicefarm-results/action.yml
@@ -5,9 +5,18 @@ inputs:
   artifact-folder:
     description: Device Farm artifact folder produced by the test action
     required: true
+  aws-region:
+    description: AWS region where the Device Farm run was scheduled
+    required: false
+  console-url:
+    description: AWS Device Farm console URL for the run
+    required: false
   platform:
     description: Platform label used in filenames and log output
     required: true
+  run-arn:
+    description: AWS Device Farm run ARN
+    required: false
   test-step-outcome:
     description: Outcome of the upstream Device Farm execution step
     required: false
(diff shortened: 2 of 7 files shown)


mrousavy:
Closing this draft. The CI-side workaround is the wrong direction; the evidence points to fixing Harness upstream so the XCTest agent exits cleanly instead of teaching this repo more Device Farm result heuristics.

PR #4015, opened 2026-06-10
test: Add failing harness tests for AE resetting to max ISO on session start

stevengoldberg:
## What

Adds two failing harness tests to `visioncamera.controller.harness.ts` reproducing a v5 regression: every `CameraSession.start()` resets auto-exposure to the format's **max ISO**, so the first ~600–900 ms of preview are massively overexposed while AE visibly ramps back down. v4 and the native camera app show correctly exposed frames immediately.

## The two tests

1. **`keeps an exposure lock applied between configure() and start()`** — deterministic and scene-independent. Locks exposure to a mid-range ISO / short duration after `configure()` but before `start()`, then asserts the lock survives `start()`. Today `start()` clobbers the lock: `exposureMode` resets to continuous auto-exposure and `iso` jumps to the format's max.

2. **`resumes auto-exposure near previously converged values on a fresh session`** — the user-facing symptom. Lets AE converge in a first session, stops it, starts a second session on the same device, and asserts the first readable ISO is within 4x of the converged value. Today the second session restarts at max ISO (~50x above a typical indoor converged value). The test skips itself when the scene is so dark that converged ISO is near max (e.g. taped-over Device Farm cameras), since the regression would be indistinguishable from correct convergence there.

## Observed telemetry (iPhone, indoor scene, RNVC 5.0.11)

```
[AE] session started
[AE] iso=5184 shutter=1/24 bias=0 mode=continuousAutoExposure   <- max ISO for the format
[AE] iso=1280 shutter=1/30 bias=0 mode=continuousAutoExposure
[AE] iso=920  shutter=1/30 bias=0 mode=continuousAutoExposure
[AE] iso=482  shutter=1/30 bias=0 mode=continuousAutoExposure
[AE] iso=296  shutter=1/35 bias=0 mode=continuousAutoExposure
[AE] iso=108  shutter=1/91 bias=0 mode=continuousAutoExposure   <- converged after ~700ms
```

This happens on every `start()`, including when the `AVCaptureDevice` still holds the previously converged exposure values, and even when an explicit `setExposureLocked(...)` was applied between `configure()` and `start()`.

## Expected

`start()` should leave the device's exposure state alone (or at minimum honor an explicit lock), matching v4 and `AVCaptureSession` behavior in a plain native app, where the first rendered frames are already correctly exposed.

### DIFF
--- a/apps/simple-camera/__tests__/visioncamera.controller.harness.ts
+++ b/apps/simple-camera/__tests__/visioncamera.controller.harness.ts
@@ -1,4 +1,10 @@
-import { beforeAll, describe, expect, it } from 'react-native-harness'
+import {
+  beforeAll,
+  describe,
+  expect,
+  it,
+  waitUntil,
+} from 'react-native-harness'
 import type {
   CameraDevice,
   CameraDeviceFactory,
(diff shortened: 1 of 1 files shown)


mrousavy:
Thanks! PR triggered

PR #4019, opened 2026-06-11
fix: Simplify `isMicrophone` accessor to inspect ports

mrousavy:
This diff avoids accessing the `AVCaptureDevice` just to inspect if it `isMicrophone`, and instead just checks if any of the `ports` are `audio` ports.


mrousavy:
I'll get https://github.com/mrousavy/react-native-vision-camera/pull/4044 on main first then update this PR with those new tests to ensure nothing breaks.

PR #4031, opened 2026-06-20
fix: catch rejections in torch/zoom/exposure control updaters

ssahil97:
## What

The internal control-updater hooks fire the nitro controller's promise-returning setters without a `.catch()`:

- `useTorchModeUpdater` → `controller.setTorchMode(...)`
- `useZoomUpdater` → `controller.setZoom(...)`
- `useExposureUpdater` → `controller.setExposureBias(...)`

This PR catches those rejections and `console.warn`s instead, matching the existing logging convention in `useCamera` / `useFrameOutput`.

## Why

These setters reject under recoverable, real-world conditions, and the un-caught promise bubbles to the JS global unhandled-rejection handler — polluting crash reporters (Sentry etc.) even though the app keeps running fine:

- **`setTorchMode`** → `java.lang.IllegalStateException: No flash unit` when the active device/lens has no torch (front camera, many emulators). Reproduce: render `<Camera torchMode="on" />` on a device whose active lens has no flash.
- **`setZoom` / `setExposureBias`** → `androidx.camera.core.CameraControl$OperationCanceledException: Camera is not active.` when the `zoom`/`exposure` prop updates while the session is inactive or tearing down (common race when a camera screen unmounts).

Example unhandled rejection seen in production (v5.0.11, Android):

```
Error: androidx.camera.core.CameraControl$OperationCanceledException: Camera is not active.
    at androidx.camera.camera2.impl.ZoomControl.applyZoomState(ZoomControl.kt:160)
    ...
    at com.margelo.nitro.camera.hybrids.HybridCameraController$setZoom$1.invokeSuspend(HybridCameraController.kt:134)

Error: java.lang.IllegalStateException: No flash unit
    at androidx.camera.camera2.impl.TorchControl.setTorchAsync(TorchControl.kt:128)
    ...
    at com.margelo.nitro.camera.hybrids.HybridCameraController$setTorchMode$1.invokeSuspend(HybridCameraController.kt:205)
```

## Changes

Each updater catches the setter rejection and logs a warning. No behavior change beyond not leaking the rejection — the failures are transient/recoverable and the existing `controller == null` / value-`null` guards are untouched.

## Notes

`setTorchMode`'s own doc already states `@throws If the device does not have a torch` — so a thrown/rejected promise here is expected, and the hook should own it rather than letting it escape.

### DIFF
--- a/packages/react-native-vision-camera/src/hooks/internal/useExposureUpdater.ts
+++ b/packages/react-native-vision-camera/src/hooks/internal/useExposureUpdater.ts
@@ -11,13 +11,20 @@ export function useExposureUpdater(
     if (controller == null) return
     if (exposure == null) return
 
+    // `setExposureBias` rejects if the session became inactive before the
+    // update applied (e.g. during teardown). This is recoverable, so catch the
+    // rejection here — otherwise it surfaces as an unhandled promise rejection.
+    const onExposureError = (error: unknown): void => {
+      console.warn('Failed to set exposure bias:', error)
+    }
+
     if (typeof exposure === 'number') {
       // number
-      controller.setExposureBias(exposure)
+      controller.setExposureBias(exposure).catch(onExposureError)
       return
     } else {
       // SharedValue<number>
-      controller.setExposureBias(exposure.get())
+      controller.setExposureBias(exposure.get()).catch(onExposureError)
       const listener = VisionCameraWorkletsProxy.bindUIUpdatesToController(
         exposure,
         controller,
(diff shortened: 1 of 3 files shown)


mrousavy:
Thanks - but console logging is not the solution. We should have an explicit `onError` parameter, which is gonna be the one we received from the `useCamera` hook.

ssahil97:
> Thanks - but console logging is not the solution. We should have an explicit `onError` parameter, which is gonna be the one we received from the `useCamera` hook.

I see, thanks for taking a look.

PR #4032, opened 2026-06-22
fix: Change `'yuv'` from `-full` to `-video` range

mrousavy:
Based on the discussion with @wcandillon in https://github.com/mrousavy/react-native-vision-camera/pull/4022, I discovered that the de-facto default on iOS is actually `kCVPixelFormatType_420YpCbCr8BiPlanarVideoRange`, not `kCVPixelFormatType_420YpCbCr8BiPlanarFullRange`.

So in this PR we change that - if you pass `'yuv'` to your `FrameOutput` it now represents video range, not full range 8-bit YUV 4:2:0 buffers.

Technically a breaking change, although pretty much all libraries handle both full and video range. So I doubt anybody will notice anything.


wcandillon:
I like it! Would be happy if it merges!

PR #4044, opened 2026-06-25
chore: Add a test to disable audio in Session and record again

mrousavy:
Adds a test to start a session with `enableAudio: true`, then re-configure to `enableAudio: false` and then record a video.

This tests if the microphone is properly removed again if it was previously added. Verified by ensuring the captured Video has no audio try (via react-native-video).


mrousavy:
We need to run #4019 against this test before merging #4019.

mrousavy:
react-native-video does not expose this metadata properly... so we cant test it

PR #4071, opened 2026-07-09
chore(ci): change android harness AWS tests to be unmetered

hannojg:
Since we now purchased a permanent device slot we have to change the billing method to `unmetered` to make use of it


hannojg:
Worked on AWS and was not metered (but used the slot)!

PR #4072, opened 2026-07-13
chore: Run iOS Harness tests with SimCam

kmagiera:
Add a GitHub Actions workflow that runs the SimpleCamera iOS React Native Harness suite against a controllable camera source on a GitHub-hosted simulator. This brings simulator camera coverage into regular CI without requiring a physical Device Farm run.

### Workflow setup:

- use the standard macos-26 runner with an iPhone 17 Pro simulator on iOS 26.5
- install the floating latest SimCam Homebrew cask, activate it with SIMCAM_LICENSE_TOKEN, and select a deterministic static image source
- keep simulator selection and lifecycle inside React Native Harness instead of manually booting or recording the simulator
- build the app, run the harness, and retain harness and Xcode logs for diagnosis

### Harness compatibility:

- rn-harness.config.mjs previously treated every CI environment as a physical-device run. HARNESS_IOS_DEVICE_MODE adds an explicit simulator override for this workflow while preserving physical-device behavior for existing CI jobs and simulator behavior for local runs.
- SimCam supplies camera video frames but does not emulate microphone input. HARNESS_SIMCAM is scoped to this workflow and skips only the records-with-audio-enabled case; all other camera and video coverage remains enabled, and non-SimCam runs continue to execute the audio case unchanged.

### Verification:

- 13 test suites passed
- 120 tests passed and 31 unsupported or conditional cases were skipped
- public sanitized logs: https://gist.github.com/kmagiera/faf73ad38085234ff51ab3aa5752f905

### Discussion:

The workflow is based on the harness-aws-device.yml and in particular uses the same paths filters for dispatch. I'm not familiar with the repository structure but given the simcam-based tests should be much faster and less expensive to run we may want to use a different set of patterns and run them more frequently if it makes sense.

### Future work

Currently simcam runs the same suite of tests that also runs on the device farm. However, these tests barely test camera related capabilities and mostly focus on configuring the camera APIs properly. Simcam allows for controlling what the camera actually sees, so it opens new possibilities for tests that can also verify the output of camera related functionalities rather than just checking whether the camera component showed up on screen.


apps/simple-camera/src/assets/simcam-harness-source.png
mrousavy, line 1: 🤣

PR #4081, opened 2026-07-19
fix: Fix uncatchable NSException crash in `setOutputSettings()` on iOS

justrustc:
## What

Any call to `CameraVideoOutput.setOutputSettings()` on iOS — even with an empty settings object — can hard-crash the app with:

```
Swift runtime failure: unhandled C++ / Objective-C exception
Thread: com.margelo.camera.video (serial)
closure #1 in HybridCameraVideoOutput.setOutputSettings(settings:)
```

## Why

`HybridCameraVideoOutput.setOutputSettings()` currently does:

```swift
var currentSettings = self.output.outputSettings(for: connection) // fully-resolved dict
currentSettings[AVVideoCodecKey] = try codec.toAVVideoCodecType()
self.output.setOutputSettings(currentSettings, for: connection)   // 💥
```

The problem: `outputSettings(for:)` returns a **fully-resolved** dictionary that includes `AVVideoWidthKey` / `AVVideoHeightKey`, but `supportedOutputSettingsKeys(for:)` does not necessarily allow those keys to be *set*. On my device (iPhone 14 Pro, iOS 18.7.8, 1080p@60 session) the whitelist is only:

```
supportedOutputSettingsKeys = [AVVideoCodecKey, AVVideoCompressionPropertiesKey]
outputSettings(for:)        = [AVVideoCodecKey, AVVideoCompressionPropertiesKey,
                               AVVideoWidthKey, AVVideoHeightKey]  // ← width/height are illegal to set back
```

Passing the resolved dict back therefore throws an `NSInvalidArgumentException` inside AVFoundation. Swift cannot catch Objective-C exceptions, so instead of a rejected Promise the whole app crashes. Since the final `setOutputSettings` call sits *outside* the `if let codec` block, this crashes even for `setOutputSettings({})`.

The allowed-keys whitelist varies by device / OS / session configuration (older devices reportedly allow *only* `AVVideoCodecKey`, see react-native-camera#2773), which is probably why this survived testing — and most users never hit this API because modern iPhones already default to HEVC.

## How

Follow the same approach the existing `setBitRate()` helper already uses: build a minimal dictionary with only the keys we actually want to change (codec, plus `AVVideoAverageBitRateKey` when `targetBitRate` is set, since `setOutputSettings` replaces the previous settings wholesale) and let AVFoundation fill in the rest with defaults valid for the new codec.

Also converts two more crash paths into catchable `RuntimeError` Promise rejections instead of NSExceptions:

- calling while the video connection is not active yet (`connection.isActive == false`)
- requesting a codec that is not in `availableVideoCodecTypes` (the error message lists the supported codecs)

## Tested

Reproduced and verified on a real device (iPhone 14 Pro, iOS 18.7.8, 1080p@60, `useVideoOutput` + `setOutputSettings({ codec: 'h265' })`):

- before: 100% crash on every call
- after: resolves normally, recording works, codec/bit-rate are applied (verified via `outputSettings(for:)` logging and the produced files)
- unsupported codec / inactive connection now reject the Promise with a descriptive message instead of crashing

### DIFF
--- a/apps/simple-camera/__tests__/visioncamera.video.harness.ts
+++ b/apps/simple-camera/__tests__/visioncamera.video.harness.ts
@@ -684,4 +684,34 @@ describe('VisionCamera - Video', () => {
     console.log(`supported video codecs: ${codecs.join(', ')}`)
     await session.stop()
   })
+
+  it('applies video output settings on iOS after the output is attached', async (context) => {
+    if (Platform.OS !== 'ios') {
+      return context.skip('setOutputSettings: iOS only')
+    }
+    const session = await VisionCamera.createCameraSession(false)
+    const videoOutput = VisionCamera.createVideoOutput({
+      targetResolution: CommonResolutions.HD_16_9,
+      enableAudio: false,
+    })
+    await session.configure([
+      {
+        input: backDevice,
+        outputs: [{ output: videoOutput, mirrorMode: 'auto' }],
+        constraints: [],
+      },
+    ])
+
+    try {
+      await videoOutput.setOutputSettings({})
+
+      const codecs = videoOutput.getSupportedVideoCodecs()
+      expect(codecs.length).toBeGreaterThan(0)
+      for (const codec of codecs) {
+        await videoOutput.setOutputSettings({ codec })
+      }
+    } finally {
+      await session.stop()
+    }
+  })
 })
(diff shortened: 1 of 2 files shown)


mrousavy:
Thanks for your PR. I fixed the code, refactored it, and also added a Harness test to test this behaviour on CI.

mrousavy (approved):

PR #4084, opened 2026-07-23
fix: route configure failures and controller setter rejections to onError

jslok:
## What

Two classes of camera errors are currently lost, both fixed by routing them to the existing `onError` from `useCamera` — following the direction given in #4031 (*"We should have an explicit `onError` parameter, which is gonna be the one we received from the `useCamera` hook."*).

### 1. `session.configure()` rejections are silently swallowed → permanent black preview

In `useCameraController`, the `await session.configure(...)` inside the effect's async `load()` has no catch. If the device can't build the session (e.g. it rejects the use-case combination — we hit this in production with a `PRIVATE`-format analysis use-case on Samsung S22 Ultra and others, confirmed via Crashlytics), the rejection disappears: **no controller, no `onError`, and the preview stays black forever** with nothing surfaced anywhere.

`configure()` failures are now caught and routed to a new internal `onConfigureError`, which `useCamera` wires to its `onError`. The connection object is built *before* the try so a throw from a user-supplied getter (a caller bug) isn't misreported as a configure failure.

### 2. Controller setter rejections surface as unhandled Promise rejections

`useZoomUpdater` / `useExposureUpdater` / `useTorchModeUpdater` fire-and-forget `setZoom` / `setExposureBias` / `setTorchMode`. These reject under recoverable, real-world conditions:

- CameraX `OperationCanceledException` ("Camera is not active") when the effect re-runs against a freshly re-configured, not-yet-streaming session (any `constraints`/fps change while `zoom` is set)
- `IllegalStateException: No flash unit` for torch on front cameras/emulators

The rejection bubbles to the global unhandled-rejection handler — full-screen RedBox in dev, crash-reporter noise in prod. The hooks now take an optional `onError` (passed from the `Camera` view's props); callers without one fall back to `console.error`, matching `useCamera`'s `defaultOnErrorHandler`.

### Worklets listener path

The same setters are also called from the `SharedValue` listener worklet in `react-native-vision-camera-worklets` (`bindUIUpdatesToController`), on the UI runtime where no `onError` is reachable without changing the `RuntimeThreadProvider` interface. Rejections there are caught and logged via `console.error`, matching the existing error convention for frame/depth callbacks in that same file. Happy to extend `bindUIUpdatesToController` with an error callback instead if you'd prefer that.

## Testing

Both fixes have been running in production in our scanner app (5.1.1 + patch-package): configure failures now surface to `onError` (we use it to fall back to a different pixel format), and pinch-zoom during session re-configuration no longer RedBoxes in dev.

### DIFF
--- a/packages/react-native-vision-camera-worklets/src/createRuntimeThreadProvider.ts
+++ b/packages/react-native-vision-camera-worklets/src/createRuntimeThreadProvider.ts
@@ -73,7 +73,17 @@ export function createRuntimeThreadProvider(): RuntimeThreadProvider {
       scheduleOnUI(() => {
         'worklet'
         value.addListener(id, (v) => {
-          controller[funcName](v)
+          // Controller setters return Promises; a rejection (e.g. CameraX
+          // OperationCanceledException while the session is re-configuring,
+          // like a pinch-zoom during an fps/constraints change) must not
+          // surface as an unhandled rejection on the UI runtime.
+          controller[funcName](v).catch((e) => {
+            const message =
+              typeof e === 'object' && e != null && 'message' in e
+                ? String(e.message)
+                : `${e}`
+            console.error(message, e)
+          })
         })
       })
       return {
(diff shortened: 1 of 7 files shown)


mrousavy:
I fixed this in #4101 ! Thanks though

PR #4085, opened 2026-07-23
feat(resizer): add 'stretch' ScaleMode

jslok:
## What

Adds a third `ScaleMode` to `react-native-vision-camera-resizer`: **`'stretch'`** fills the whole output by stretching the input independently on each axis — no black bars, no cropping, aspect ratio **not** preserved.

## Why

ML models with fixed square inputs (e.g. 640×640 detectors) are typically trained on images stretched to square. With only `cover`/`contain` available:

- `cover` center-crops, cutting off detection targets near the frame edges
- `contain` letterboxes, wasting ~44% of a square input tensor on black bars when fed 16:9 camera frames (and shrinking the effective object resolution accordingly)

`stretch` matches the training-time preprocessing exactly and uses every input pixel. We use it in production to feed 16:9 frames into a square keypoint-detection model — accuracy at frame edges improved measurably vs `cover`.

## Implementation

- **Vulkan (Android)** `Resizer.comp` + `VulkanResizerShaderConfig.cpp`: ordinal `2u`; the output coordinate maps 1:1 onto the source `[0,1]²` per axis. The `cover`/`contain` math is unchanged, just moved into per-case blocks since `stretch` has no scale factor or centering offset.
- **Metal (iOS)** `ResizerKernels.metal` + `OutputFormat+Resizer.swift`: same mapping, ordinal `2`.
- **TS spec** `ResizerFactory.nitro.ts`: `ScaleMode = 'cover' | 'contain' | 'stretch'` + docs.
- **`nitrogen/generated`** `ScaleMode.swift` / `ScaleMode.hpp`: updated to match what `nitrogen` emits for the new union member (happy to re-run `bun specs` if you'd rather the CI regenerate).

Mirroring and rotation are applied after the coordinate mapping, so they compose with `stretch` the same way they do with the existing modes.

## Testing

Running in production on both platforms (Android Vulkan path across Adreno/Mali/Xclipse, iOS Metal path) as a patch-package addition — output tensors verified against CPU-reference stretched resizes.

### DIFF
--- a/packages/react-native-vision-camera-resizer/android/src/main/cpp/vulkan/VulkanResizerShaderConfig.cpp
+++ b/packages/react-native-vision-camera-resizer/android/src/main/cpp/vulkan/VulkanResizerShaderConfig.cpp
@@ -107,6 +107,8 @@ uint32_t ShaderSpecializationData::getScaleModeOrdinal(ScaleMode scaleMode) {
       return 0u;
     case ScaleMode::CONTAIN:
       return 1u;
+    case ScaleMode::STRETCH:
+      return 2u;
   }
 
   throw std::runtime_error("Unsupported Resizer ScaleMode.");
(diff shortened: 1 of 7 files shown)


mrousavy:
> ML models with fixed square inputs (e.g. 640×640 detectors) are typically trained on images stretched to square

Wait really?

jslok:
> > ML models with fixed square inputs (e.g. 640×640 detectors) are typically trained on images stretched to square
> 
> Wait really?

Claude put that and I didn't check it. 😬

But stretch is still useful in many cases. Thanks for the quick merge!

Issue #3773, opened 2026-04-23
Race-condition crash in HybridCameraSession.start() — attach-side sibling of #3730 (iPhone 11, iOS 18.7.7, v5.0.4)


qutrek:
### What's happening?

App crashes with `EXC_CRASH (SIGABRT)` on iPhone 11 / iOS 18.7.7 during normal camera usage. The crashed thread is always a CoreMedia `FigCaptureSessionNotificationQueue` worker processing `_handleConfigurationCommittedNotificationWithPayload` and asserting in `[AVCaptureOutput attachToFigCaptureSession:]_block_invoke.cold.1` at `AVCaptureOutput.m:364`. A *separate* thread is in `HybridCameraSession.start()` → `AVCaptureSession.startRunning()` → `_buildAndRunGraph` at the moment of the assert.

This is the **attach-side** sibling of #3730. #3730 (closed as fixed in v5.0.0) was the detach-side of the same race; the attach side appears not to have been addressed and is reproducible in v5.0.4.

We have **6 byte-for-byte identical** crash reports from a single user from real TestFlight usage. All triggered by ordinary actions: photo retake, zoom switching, video record start, and general camera usage with no specific repro action reported. Each triggered through a different user-visible action but produced the *same* stack frames in the crashed thread and the same parallel `HybridCameraSession.start()` thread.

### Hypothesis

`HybridCameraSession.start()` calls `session.startRunning()` immediately after a previous `configure()`'s `commitConfiguration()` returned, with no barrier waiting for CoreMedia to finish propagating the asynchronous "configuration committed" notification on `FigCaptureSessionNotificationQueue`. On multi-physical-cam devices (iPhone 11 has wide + ultra-wide), the CoreMedia XPC commit takes longer to propagate than on single-cam devices, widening the race window. When the notification handler runs `_makeConfigurationLive:` and iterates outputs to call `attachToFigCaptureSession:`, an output's `_outputInternal->figCaptureSession` is non-NULL (already attached by the racing `startRunning`), tripping the internal assert.

The serial `Self.queue` in `HybridCameraSession` correctly serializes JS-driven calls (`configure`, `start`, `stop`), but does not synchronize with CoreMedia's async commit notification.

### Reproduceable Code

Standard usage. The `<Camera>` is mounted with stable photo/video outputs; `isActive` is derived from focus + foreground + a `pendingMedia` flag (off during a per-shot review screen).

```tsx
[...]

PR #4086, opened 2026-07-23
fix: Don't remove from `session` collections while enumerating them

mrousavy:
Two fixes for `CameraSession` reconfigure/teardown crashes:

1. Reworks `updateInputs`/`updateOutputs`/`updateConnections` to collect-then-remove in two separate steps instead of removing from `session.inputs`/`outputs`/`connections` while enumerating them.
2. Fixes the #3773/#3730/#4026 SIGABRT family: an `AVCaptureOutput` is attached/detached to the underlying `FigCaptureSession` asynchronously, so an output removed from one session (e.g. hook-owned outputs surviving a `<Camera>` re-mount) can still be attached at the CoreMedia level - adding it to a second session then crashes in `attachToFigCaptureSession:`. The only point where the detach is guaranteed is the previous session's `deinit`, which detaches synchronously. `useCameraSession` now `dispose()`s the session on unmount (deterministic, ordered on the session queue before any re-mount work), and outputs track their session via a `weak` reference so `configure()` fails with a clear error if an output is moved to a new session while the previous one is still alive - instead of racing CoreMedia. Inner `AVCaptureOutput` instances stay stable for the output's whole lifetime.

Adds Harness tests for the previously uncovered removal paths (device switch while running, multi-output removal in one reconfigure, camera + microphone input removal in one pass, `configure([])` teardown + reconfigure), a regression test re-using the same outputs across 9 re-created sessions (#3773), and a deterministic test of the new output re-use guard.

Note: a deterministic crash repro for the Fig race is not possible from JS - the attach/detach ordering inversion happens inside mediaserverd's per-session queues (verified by disassembly), which is also why the guard exists: the safe re-use point is only provable at session deinit.

### DIFF
--- a/apps/simple-camera/__tests__/visioncamera.session.harness.ts
+++ b/apps/simple-camera/__tests__/visioncamera.session.harness.ts
@@ -5,6 +5,7 @@ import {
   it,
   waitUntil,
 } from 'react-native-harness'
+import { Platform } from 'react-native'
 import type {
   CameraDeviceFactory,
   TargetCameraPosition,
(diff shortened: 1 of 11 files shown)


mrousavy:
Bridged ObjC arrays like these are copy anyways, so this PR is effectively no-op. AI hallucinated "fix". The Harness tests however are real tests and can be merged i think

PR #4090, opened 2026-07-24
feat: Add `zoom`, `exposure`, `torchMode` props and `controller` ref to `<SkiaCamera />`

mrousavy:
## What

`<SkiaCamera />` now supports the declarative `zoom`, `exposure` and `torchMode` props (matching `<Camera />`), and exposes the `CameraController` on its ref for imperative control.

## Why

`<SkiaCamera />` extends `CameraProps`, but `zoom` / `exposure` / `torchMode` are **not** part of that shared base - they live on the view components (`CameraViewProps`) as declarative convenience wrappers over `CameraController.setZoom` / `setExposureBias` / `setTorchMode`. `<Camera />` wires them up via `useZoomUpdater` / `useExposureUpdater` / `useTorchModeUpdater`; `<SkiaCamera />` simply never did.

These are `CameraController` capabilities, **not** preview features, so nothing about swapping the native Preview for a Skia Canvas justified their absence. And because `SkiaCameraRef` didn't expose `controller` either, there was previously **no** declarative *or* imperative path to torch/zoom/exposure from a `<SkiaCamera />` at all.

## Changes

- **Core:** Move `useZoomUpdater`, `useExposureUpdater`, `useTorchModeUpdater` from `hooks/internal/` to `hooks/` and export them from the package root, so the separate `-skia` package can reuse the identical wiring (including animated `SharedValue` support) instead of duplicating the binding logic. `<Camera />` now imports them from the new location - no behavior change.
- **Skia:** Add `zoom`, `exposure`, `torchMode` props (plus the matching `getInitialZoom` / `getInitialExposureBias` getters) and wire them through the same three updater hooks. Expose `controller: CameraController | undefined` on `SkiaCameraRef`.

## Not included (intentionally)

Preview-specific props stay excluded because the native Preview view is replaced by a Skia Canvas: `resizeMode`, `implementationMode`, `onPreviewStarted` / `onPreviewStopped`, and the native `enableNativeZoomGesture` / `enableNativeTapToFocusGesture` (use RN Gesture Handler + the `controller` ref / `focusTo` instead).

## Testing

- `bun camera typecheck`, `bun skia typecheck` and `biome check` all pass.
- The underlying controller methods these props delegate to (`setZoom`, `setTorchMode`, `setExposureBias`) are already covered by the on-device harness in `visioncamera.controller.harness.ts`.
- Note: there is currently **no** harness coverage for `<SkiaCamera />` - the `-skia` package isn't wired into the `simple-camera` harness app, and even the declarative props on `<Camera />` aren't tested at the view level (only the controller methods are). Adding a `<SkiaCamera />` harness would mean pulling `@shopify/react-native-skia` into the test app plus a native rebuild; happy to do that as a follow-up if wanted.


mrousavy:
Closing in favor of #4101

PR #4092, opened 2026-07-24
fix: Simplify microphone detection + adding

### LINKED ISSUE #4091: Camera flip crashes on iOS 16 (`EXC_BAD_ACCESS` in `isMicrophone`)
On iOS 16, flipping the camera (any session reconfigure with inputs attached) crashes immediately with `EXC_BAD_ACCESS (SIGSEGV)`, `KERN_INVALID_ADDRESS at 0x0`. Only happens in Release builds, only on iOS < 17. Verified on an iPhone 8 (iOS 16.7.16), VisionCamera 5.1.1.

Crash is on the camera session queue, top frame is `HybridCameraSession.updateInputs` → `AVCaptureInput.isMicrophone`:

```
Thread com.margelo.camera.session:
0  HybridCameraSession.updateInputs(_:)  (HybridCameraSession.swift:233)
1  closure #1 in HybridCameraSession.configure
```

**Root cause:** `isMicrophone` compares `deviceType == .microphone`. `AVCaptureDeviceTypeMicrophone` is iOS 17+, so it's a *weak* import - the GOT slot is null on iOS 16. Even though the comparison sits behind `#available(iOS 17.0, *)`, the `-O` optimizer hoists the load of the constant out of the loop in `updateInputs`, above the availability check, so we dereference the null slot before the guard runs → segfault at `0x0`. Cold start is fine (no inputs, loop body never runs); the first flip is the first time the loop executes with inputs, which is why it always hits there. iOS 17+ and simulators are immune because the symbol binds.

(This is actually a Swift compiler bug - the optimizer shouldn't speculate that load past `#available`. Reported upstream at swiftlang/swift#90916 / swiftlang/swift#90931 - but we should still fix it in VisionCamera since it affects anything built with current toolchains.)

**Fix:** don't reference the weak symbol at all. `isMicrophone` can just check the media type, which is what it actually means and works on every iOS version:

```swift
var isMicrophone: Bool {
  guard let self = self as? AVCaptureDeviceInput else { return false }
  return self.device.hasMediaType(.audio)
}
```

Verified on-device: with the `.microphone` comparison the app crashes on the first flip; with `hasMediaType(.audio)` it survives repeated flips.

**Same hazard elsewhere** (gated constants compared under `#available`, could crash the same way if the optimizer hoists them): `AV+DeviceType.swift` (`.builtInLiDARDepthCamera`, `.continuityCamera`, `.external`), `HybridCameraDevice.swift`, `AV+VideoCodec.swift` (ProRes). Worth wrapping those comparisons in an `@available @inline(never)` helper so the load can't be speculated.

mrousavy:
Simplify the microphone detection and removal code.

Previously we checked against `deviceType` being `.microphone` or `.builtInMicrophone`, which under a specific Swift compiler bug was inlined, and on iOS 16 accidentally resolved to a `nil` value at runtime causing a crash.

This was reported in the Swift repo (https://github.com/swiftlang/swift/issues/90916) and a PR has already been created to fix this (https://github.com/swiftlang/swift/pull/90945) - but that will take a while before it lands in an Xcode release.

- Fixes https://github.com/mrousavy/react-native-vision-camera/issues/4091

### DIFF
--- a/packages/react-native-vision-camera/ios/Extensions/AVFoundation/AVCaptureInput+isMicrophone.swift
+++ b/packages/react-native-vision-camera/ios/Extensions/AVFoundation/AVCaptureInput+isMicrophone.swift
@@ -13,10 +13,6 @@ extension AVCaptureInput {
     guard let self = self as? AVCaptureDeviceInput else {
       return false
     }
-    if #available(iOS 17.0, *) {
-      return self.device.deviceType == .microphone
-    } else {
-      return self.device.deviceType == .builtInMicrophone
-    }
+    return self.device.hasMediaType(.audio)
   }
 }
--- a/packages/react-native-vision-camera/ios/Extensions/AVFoundation/AVCaptureSession+addAudioInput.swift
+++ b/packages/react-native-vision-camera/ios/Extensions/AVFoundation/AVCaptureSession+addAudioInput.swift
@@ -10,7 +10,9 @@ import NitroModules
 
 extension AVCaptureSession {
   func addAudioInput() throws {
-    let microphone = try getDefaultMicrophone()
+    guard let microphone = AVCaptureDevice.default(for: .audio) else {
+      throw RuntimeError.error(withMessage: "This device does not have a microphone!")
+    }
     let input = try AVCaptureDeviceInput(device: microphone)
     logger.info("Adding microphone \(input)...")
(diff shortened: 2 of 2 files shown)


den1a:
Tested on ios 16. Everything is fine! Thanks!

PR #4115, opened 2026-07-31
feat: Add `onUIRotationChanged` event

mrousavy:


mrousavy:
made it better in #4147

PR #4117, opened 2026-08-01
fix(resizer): bound the AHardwareBuffer import cache and guard dispose() against in-flight resize()

jslok:
Two independent lifetime bugs in the Resizer, both found while profiling a long-running Frame Processor that resizes every frame.

## 1. The AHardwareBuffer import cache is unbounded (Android)

`VulkanHardwareBufferInterop::_cachedImages` only ever grows. Importing an `AHardwareBuffer` into Vulkan **acquires a reference on it** (per the `VK_ANDROID_external_memory_android_hardware_buffer` spec), so every cached entry pins a full camera buffer alive for the lifetime of the `Resizer`.

That is fine while one camera session streams — CameraX recycles a small set of buffers, which is exactly what the cache is for. The problem is session restarts: an fps/constraints change, a pause/resume, a format change, etc. make CameraX allocate a **fresh** buffer set, and the old set stays pinned by the cache forever. Nothing ever evicts it, not even when the camera fully disconnects.

Measured on a Pixel 10 over ~49 min of ordinary use with periodic camera restarts:

| | start | end |
|---|---|---|
| retained 1920x1080 YUV buffers | 39 | 137 |
| gralloc held by them | ~120 MB | ~418 MB |
| total PSS | 588 MB | 1.21 GB |

Fix: the cache is now LRU-ordered (`splice` to front on a hit) and bounded by `kMaxCachedImages = 12`, evicting from the back before an insert. A single streaming session's working set is ~4-6 buffers (`ImageAnalysis` with `maxImages=4`), so 12 leaves generous slack while still converging onto the current session's buffers within a few frames of a restart. The cache is a pure optimization anyway — `VulkanResizerPipeline::run(..)` submits and waits synchronously, so a miss costs one import.

Storage changed from `std::vector` to `std::list` because `importImage(..)` hands out a reference *into* the container and the LRU now reorders on hits and evicts on misses; list nodes have stable addresses, so `splice(..)` and `pop_back()` provably never move the entry being returned.

Eviction is safe at that point: `run(..)` holds `_stateMutex` across import -> dispatch -> submit-and-wait, so no other imported image is in flight on the GPU.

## 2. `dispose()` races an in-flight `resize()` (both platforms)

`resize()` read `_pipeline`, null-checked it, then dereferenced it with no synchronization, while `dispose()` could concurrently reset the same `unique_ptr` — which runs `~VulkanResizerPipeline()` and tears down the `VkDevice` (and the pipeline's own `_stateMutex`) out from under an in-flight `run()`. That is a native use-after-free, not the catchable `"This Resizer has already been disposed!"` error callers are written against.

iOS has the same shape: a Swift stored-property load/store of a class reference is not atomic, so an unguarded `pipeline = nil` racing the `guard let` load is a data race on the possibly-last reference.

This is reachable whenever a consumer disposes eagerly rather than leaving the instance to the GC — `dispose()` then runs on the JS thread while `resize()` can still be running on the Frame Processor thread.

Fix: a `std::mutex` (Android) / `NSLock` (iOS) held across `resize()`, `dispose()` and the `memorySize` / `getExternalMemorySize()` getter. A dispose landing mid-frame now blocks until that resize returns, and every later call throws the catchable error.

`getExternalMemorySize()` needs the guard too — Nitro calls it from `HybridObject::toObject(..)`, i.e. on whichever thread converts the Resizer into a Runtime, including a Frame Processor's `unbox()` on the camera thread. It cannot self-deadlock: Nitro never calls it from inside `resize()`/`dispose()`.

A `GPUFrame` that is still checked out is unaffected by `dispose()` — on iOS its `MetalBufferView`'s `onRelease` closure strongly captures the `MetalReusableBuffer`, so ARC keeps the output buffer alive independently of the pipeline.

## Testing

Built and run on-device (Android, Vulkan path) with a per-frame Frame Processor. The buffer bound was A/B verified: with the LRU cap the retained-buffer count stays flat across repeated camera restarts instead of ratcheting, and PSS no longer climbs. The dispose guard was exercised by disposing from a React effect cleanup while frames were still being processed.

I have not been able to test the iOS Metal path on device — the iOS change is the mechanical equivalent of the Android one, so please review it with that in mind.

### DIFF
--- a/packages/react-native-vision-camera-resizer/android/src/main/cpp/HybridResizer.cpp
+++ b/packages/react-native-vision-camera-resizer/android/src/main/cpp/HybridResizer.cpp
@@ -38,6 +38,10 @@ namespace {
 HybridResizer::HybridResizer(const ResizerOptions& options) : HybridObject(TAG), _pipeline(std::make_unique<vulkan::VulkanResizerPipeline>(options)) {}
 
 std::shared_ptr<HybridGPUFrameSpec> HybridResizer::resize(const std::shared_ptr<camera::HybridFrameSpec>& frame) {
+  // Held for the whole call - `dispose()` destroys `_pipeline`, so releasing the lock before `run()`
+  // returns would let the pipeline be freed underneath us.
+  std::lock_guard<std::mutex> lock(_lifecycleMutex);
+
   if (_pipeline == nullptr) [[unlikely]] {
     throw std::runtime_error("This Resizer has already been disposed!");
   }
@@ -59,6 +63,9 @@ std::shared_ptr<HybridGPUFrameSpec> HybridResizer::resize(const std::shared_ptr<
 }
 
 void HybridResizer::dispose() {
+  // Blocks until any in-flight `resize()` has returned, so the pipeline is never destroyed while it is running.
+  std::lock_guard<std::mutex> lock(_lifecycleMutex);
+
   if (_pipeline == nullptr) {
     return;
   }
@@ -71,6 +78,10 @@ void HybridResizer::dispose() {
 }
 
 size_t HybridResizer::getExternalMemorySize() noexcept {
+  // Nitro calls this from `HybridObject::toObject(...)`, i.e. on whichever Thread converts the Resizer
+  // into a Runtime - including a Frame Processor's `unbox()` on the camera Thread.
+  std::lock_guard<std::mutex> lock(_lifecycleMutex);
+
   return _pipeline != nullptr ? _pipeline->getOutputBufferAllocationSize() : 0;
 }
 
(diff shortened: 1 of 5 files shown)


mrousavy:
Those are two unrelated changes, please split into two PRs.

mrousavy:
Also, why did you not simply add the eviction code for >=12 images and kept the `std::vector` and `back()` code as is? Why change it to a `std::list`?

jslok:
> Also, why did you not simply add the eviction code for >=12 images and kept the `std::vector` and `back()` code as is? Why change it to a `std::list`?

Valid point. I will resubmit as you suggest and as separate PRs later today.

mrousavy:
Thanks!

PR #4129, opened 2026-08-05
fix(resizer): bound the AHardwareBuffer import cache (Android)

jslok:
Split out of #4117 ([as requested](https://github.com/mrousavy/react-native-vision-camera/pull/4117#issuecomment-5164805114)), reworked to [keep the `std::vector` and `back()` code as suggested](https://github.com/mrousavy/react-native-vision-camera/pull/4117#issuecomment-5164824629) — the container and hit path are unchanged, the bound is plain oldest-first eviction with no LRU reordering.

Importing an AHardwareBuffer into Vulkan acquires a reference on it, so every entry in the Resizer's unbounded import cache pins a full camera buffer. Each camera session restart (fps/constraints change, pause/resume) makes CameraX allocate a fresh buffer set, so the pinned set grows by a whole set per restart and is never released. Measured on a Pixel 10 over ~49 min with periodic restarts: 39 → 137 retained 1920x1080 buffers (~418 MB of gralloc), total PSS 588 MB → 1.21 GB.

Fix: cap the cache at 12 entries, evicting the oldest-inserted entry before inserting. FIFO is enough — eviction only fires on inserts, and a steady streaming session (~4–6 recycled buffers) is all cache hits, so the worst case is a few one-off re-imports around a restart. Eviction is safe at that point because `run()` holds `_stateMutex` across import → dispatch → submit-and-wait, so nothing is in flight on the GPU.

Tested on device (Pixel 10, Vulkan path): retained-buffer count stays flat across repeated camera restarts instead of ratcheting, and PSS no longer climbs.

### DIFF
--- a/packages/react-native-vision-camera-resizer/android/src/main/cpp/vulkan/VulkanHardwareBufferInterop.cpp
+++ b/packages/react-native-vision-camera-resizer/android/src/main/cpp/vulkan/VulkanHardwareBufferInterop.cpp
@@ -73,6 +73,14 @@ const VulkanHardwareBufferInterop::ImportedImage& VulkanHardwareBufferInterop::i
     iterator = _cachedImages.erase(iterator);
   }
 
+  // Evict the oldest entries before inserting so old camera sessions' buffers get released.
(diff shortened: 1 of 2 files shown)


mrousavy:
Thanks!

Issue #4026, opened 2026-06-14
🐛 [iOS 26] SIGABRT on camera teardown — AVCaptureSession dealloc → detachFromFigCaptureSession assert (V5 detach-side of #3730/#3773)


duc-beluga:
### What's happening?

On iOS 26, the app crashes with `EXC_CRASH (SIGABRT)` when the `<Camera>` view is torn down (unmounting the screen after capturing a photo). The crash is an AVFoundation assertion fired from `-[AVCaptureSession dealloc]` running **on the main thread inside a `CA::Transaction::commit()`** — the `AVCaptureVideoPreviewLayer`'s deferred release cascades into a session dealloc that asserts in `detachFromFigCaptureSession`.

This is the **detach/teardown side** of the race in #3773 (attach side, open). #3730 was the detach side too and was closed as "fixed in V5" — this reproduces on **V5 (5.0.10)**, so the teardown side does not appear fixed. Setting `isActive={false}` before unmount and delaying the unmount ~240ms both reduce but do **not** eliminate it.

### Reproduceable Code

`<Camera>` is mounted inside a bottom-sheet/overlay. After `capturePhotoToFile` resolves, the photo-preview branch renders (so `<Camera>` unmounts), then the overlay itself unmounts on submit.

```tsx
const device = useCameraDevice('back');
const photoOutput = usePhotoOutput({ qualityPrioritization: 'speed', containerFormat: 'jpeg' });

<Camera
  style={StyleSheet.absoluteFill}
  device={device}
  outputs={[photoOutput]}     // photo only
  isActive={appForeground && hostOpen}
  onError={...}
/>
```

### Relevant log output

Symbolicated, byte-for-byte identical across 2 TestFlight crash reports:

```
2   libsystem_c.dylib   abort
3   libsystem_c.dylib   __assert_rtn
4   AVFCapture  -[AVCaptureOutput detachFromFigCaptureSession:]_block_invoke.cold.1 (AVCaptureOutput.m:329)
5   AVFCapture  -[AVCaptureOutput detachFromFigCaptureSession:]_block_invoke (AVCaptureOutput.m:329)
8   AVFCapture  -[AVCaptureOutput detachFromFigCaptureSession:] (AVCaptureOutput.m:325)
9   CoreFoundation  -[NSSet makeObjectsPerformSelector:withObject:]
10  AVFCapture  -[AVCaptureSession _makeConfigurationLive:] (AVCaptureSession.m:4862)
[...]

PR #4131, opened 2026-08-05
fix: detach preview layers before camera teardown

caiobarroso:
Detaches unwanted iOS preview layers before AVCaptureSession removes inputs or outputs.

Adds a Harness regression test covering photo capture followed by Camera unmount and remount.

Validated on a physical iPhone 17 Pro running iOS 26.5.

### DIFF
--- a/apps/simple-camera/__tests__/visioncamera.camera-view.harness.tsx
+++ b/apps/simple-camera/__tests__/visioncamera.camera-view.harness.tsx
@@ -325,6 +325,79 @@ describe('VisionCamera - Camera View', () => {
     }
   })
 
+  it('captures a photo before unmounting and remounting the Camera view', async () => {
+    const photoOutput = VisionCamera.createPhotoOutput({
+      targetResolution: CommonResolutions.HD_4_3,
+      containerFormat: 'jpeg',
+      quality: 0.8,
+      qualityPrioritization: 'balanced',
+    })
+
+    for (let attempt = 1; attempt <= 2; attempt++) {
+      const cameraRef = createRef<CameraRef>()
+      const layout = deferred<Layout>()
+      const started = deferred()
+      const previewStarted = deferred()
+      const onError = fn((error: Error) => {
+        layout.reject(error)
+        started.reject(error)
+        previewStarted.reject(error)
+      })
+
+      const { unmount } = await render(
+        <Camera
+          ref={cameraRef}
+          device={backDevice}
+          isActive={true}
+          outputs={[photoOutput]}
+          style={StyleSheet.absoluteFill}
+          onLayout={(event) => {
+            layout.resolve(toLayout(event))
+          }}
+          onStarted={started.resolve}
+          onPreviewStarted={previewStarted.resolve}
+          onError={onError}
+        />,
+      )
+
+      const cameraLayout = await withTimeout(
+        layout.promise,
+        10_000,
+        `photo remount Camera onLayout attempt ${attempt}`,
+      )
+      await withTimeout(
+        started.promise,
+        15_000,
+        `photo remount Camera onStarted attempt ${attempt}`,
+      )
+      await withTimeout(
+        previewStarted.promise,
+        15_000,
+        `photo remount Camera onPreviewStarted attempt ${attempt}`,
+      )
+      expect(onError).not.toHaveBeenCalled()
+
+      const camera = cameraRef.current
+      if (camera == null) throw new Error('no Camera ref')
+      expectPreviewGeometry(camera, cameraLayout)
+
+      const photo = await photoOutput.capturePhoto(
+        { flashMode: 'off', enableShutterSound: false },
+        {},
+      )
+      try {
+        expect(photo.width).toBeGreaterThan(0)
+        expect(photo.height).toBeGreaterThan(0)
+      } finally {
+        photo.dispose()
+      }
+
+      // Do not wait for a timer here. This intentionally exercises the
+      // unmount/remount boundary that can race with native session teardown.
+      unmount()
+    }
+  })
+
   it('replaces an active Camera view with another active Camera view in one update', async () => {
     const firstRef = createRef<CameraRef>()
     const secondRef = createRef<CameraRef>()
--- a/packages/react-native-vision-camera/ios/Hybrid Objects/HybridCameraSession.swift
+++ b/packages/react-native-vision-camera/ios/Hybrid Objects/HybridCameraSession.swift
@@ -66,6 +66,12 @@ final class HybridCameraSession: HybridCameraSessionSpec {
         )
       }
 
+      // Removing an input or output can implicitly remove its AVCaptureConnection
+      // before updateConnections(...) gets a chance to inspect it. Detach preview
+      // layers first so they do not keep a stale reference to this session while
+      // the Camera view is being unmounted.
+      self.detachUnwantedPreviewLayers(connections)
+
       // Remove all unwanted inputs and add all new inputs
       try self.updateInputs(connections)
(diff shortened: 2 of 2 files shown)


mrousavy:
I don't get the issue - what happens without your PR?

`updateOutputs(..)` may remove a Preview Layer if it is no longer part of the target connections, which implicitly also removes its `AVCaptureConnection` - your comment correctly highlights that - so where is the problem? Why do you have to remove it before it's being implicitly removed?

It all runs under a `beginConfiguration()`/`endConfiguration()` batch, so where is the problem? Both execute at the same time?

"Fixes" for race conditions are often a sign of code smell or flakyness, often LLMs hallucinate fixes and cannot reproduce it anymore once and call it a day- but logically there shouldn't be any difference other than more work and more code.

caiobarroso:
> I don't get the issue - what happens without your PR?
> 
> `updateOutputs(..)` may remove a Preview Layer if it is no longer part of the target connections, which implicitly also removes its `AVCaptureConnection` - your comment correctly highlights that - so where is the problem? Why do you have to remove it before it's being implicitly removed?
> 
> It all runs under a `beginConfiguration()`/`endConfiguration()` batch, so where is the problem? Both execute at the same time?
> 
> "Fixes" for race conditions are often a sign of code smell or flakyness, often LLMs hallucinate fixes and cannot reproduce it anymore once and call it a day- but logically there shouldn't be any difference other than more work and more code.


https://github.com/user-attachments/assets/3e869189-a0d8-4c8f-a276-9c406ed7018f
https://github.com/user-attachments/assets/d5368546-a8f5-4f13-98b6-e7d099f03ee3

This happened only on iPhone

In the app that I work on, one Camera modal stayed open while users took up to 10 photos. The photos were shown in an app preview list while the native Camera preview remained active

When the user tapped Send, the progress bar started and the batch upload began. During the send and camera closing flow, the app would frequently crash. I attached two videos showing real production cases

The connection and the preview layer are separate. Removing the output may remove the connection automatically, but the preview layer could still be attached at that point. Our change detaches the preview first, then removes the connection and output

This is not about two calls running at the same time. It is about the teardown order. `beginConfiguration()` groups the session changes, but it does not detach the preview layer

I could not reproduce this in the normal Simple Camera flow because its Camera is always mounted. It does not exercise the same capture, send, and teardown flow. We only observed this problem on iPhone

The production patch included a broader shutdown flow for our modal. This PR extracts the relevant iOS teardown change as a smaller upstream fix. After applying the full patch in our app, the same flow stopped crashing and remained stable.

This is also consistent with [issue #4026](https://github.com/mrousavy/react-native-vision-camera/issues/4026).

mrousavy:
> The connection and the preview layer are separate. Removing the output may remove the connection automatically, but the preview layer could still be attached at that point. Our change detaches the preview first, then removes the connection and output

I don't understand this part - my understanding is that if I remove the preview layer, it also removes the connection - is that not the case?

mrousavy:
Ohh, is this the bug?

1. User has one Preview attached, configured just fine
2. User removes his Preview but wants to attach a new preview
3. `updateOutputs(..)` adds the new preview but apparently doesnt remove the first because we rely on `updateConnections(..)` for that
4. It crashes because we have two previews?

mrousavy:
Wait actually I might be wrong, because we test exactly that in `starts a replacement preview and stops the removed preview while running`

caiobarroso:
> Ohh, is this the bug?
> 
> 1. User has one Preview attached, configured just fine
> 2. User removes his Preview but wants to attach a new preview
> 3. `updateOutputs(..)` adds the new preview but apparently doesnt remove the first because we rely on `updateConnections(..)` for that
> 4. It crashes because we have two previews?

I don’t think that is quite the flow we had

Yes, explicitly removing the Preview also removes its connection. But the user is not removing one Preview and attaching another one

There is only one native Preview. The user simply taps Send. In one video the user took four photos, and in the other only one. In both cases, the app crashes when the send flow reaches the Camera teardown. I attached both videos from real production cases. So this is not a “two previews” problem

The fix is to detach the existing Preview before the Camera session removes its connections during teardown. `beginConfiguration()` groups the changes, but it does not change this order

I could not reproduce it in Simple Camera because its Camera is always mounted, so it does not go through this teardown flow. We only observed this only on iPhone, android works perfectly fine

If helpful, I can also share the complete patch we have been relying on since before version 5.1.0. It has worked reliably for us, which is why I have not moved to newer versions yet. The camera is a critical part of our operation, so I cannot risk the crash coming back

caiobarroso:
> > Ohh, is this the bug?
> > 
> > 1. User has one Preview attached, configured just fine
> > 2. User removes his Preview but wants to attach a new preview
> > 3. `updateOutputs(..)` adds the new preview but apparently doesnt remove the first because we rely on `updateConnections(..)` for that
> > 4. It crashes because we have two previews?
> 
> I don’t think that is quite the flow we had
> 
> Yes, explicitly removing the Preview also removes its connection. But the user is not removing one Preview and attaching another one
> 
> There is only one native Preview. The user simply taps Send. In one video the user took four photos, and in the other only one. In both cases, the app crashes when the send flow reaches the Camera teardown. I attached both videos from real production cases. So this is not a “two previews” problem
> 
> The fix is to detach the existing Preview before the Camera session removes its connections during teardown. `beginConfiguration()` groups the changes, but it does not change this order
> 
> I could not reproduce it in Simple Camera because its Camera is always mounted, so it does not go through this teardown flow. We only observed this only on iPhone, android works perfectly fine
> 
> If helpful, I can also share the complete patch we have been relying on since before version 5.1.0. It has worked reliably for us, which is why I have not moved to newer versions yet. The camera is a critical part of our operation, so I cannot risk the crash coming back

The actual flow is:

1. The user mounts the Camera inside the modal.
2. The user takes one or more photos. In the videos, one user takes four photos and another takes only one.
3. The native Camera Preview remains mounted while the photos are shown in the app.
4. The user taps Send.
5. The progress bar starts and the photos are processed.
6. The modal then closes and the Camera session is torn down.
7. Without the patch, the app could crash during this flow.
8. The patch detaches the existing Preview before the session removes its connections.

The user never removes one Preview and attaches another one. There is only one native Preview

mrousavy:
I think I fixed this in https://github.com/mrousavy/react-native-vision-camera/pull/4134

mrousavy:
Actually hmm based on your flow I dont think this is related

caiobarroso:
> I think I fixed this in #4134

Thanks, I see what #4134 fixes. It addresses the case where one native Preview is replaced by another, and that is a real issue

But I think that is not the flow in our app. We have one native Preview. The user taps Send, and the Camera session is later torn down. Since `updateInputs()` runs before `updateOutputs()`, the connection may already be removed before the Preview layer is handled by #4134

This is why #4131 detaches the Preview before the input and output teardown starts.

I can share the complete patch we have been using since before version 5.1.0, as well as the relevant code if u need to reproduce the same error

mrousavy:
I changed my PR to do almost the same as this PR with slightly different code and E2E verified tests. Lmk if that works for you!

caiobarroso:
> I changed my PR to do almost the same as this PR with slightly different code and E2E verified tests. Lmk if that works for you!

Thank you so much! I"ll test and will share some feedback

caiobarroso:
> I changed my PR to do almost the same as this PR with slightly different code and E2E verified tests. Lmk if that works for you!

I tested the final change in our app and the crashes are fully gone

Thank you very much for all the effort you put into VisionCamera and the community. I’m a big fan of your work

Since the fix follows the main idea from #4131, would you mind mentioning @caiobarroso or linking #4131 in the PR description for attribution? I would really appreciate it

mrousavy:
sure

Issue #3356, opened 2025-01-07
🐛 Camera Preview sometimes hangs in landscape view on iPadOS


marcinnexudus:
### What's happening?

Some users have reported crashes when they land on the screen where the Camera preview is. This is a tablet app. Looking at the Bugsnag reports it happens only in the landscape orientation and only on iPadOS . This seems to be an issue:

0  libsystem_kernel.dylib +0x1598   ___ulock_wait2
1  libsystem_platform.dylib +0x1d98 __os_unfair_lock_lock_slow
2  libobjc.A.dylib +0xb5d4          _objc_sync_enter
3  AVFCapture +0x20d40              -[AVCaptureVideoPreviewLayer layerDidBecomeVisible:]

The full stack trace is on pastebin: https://pastebin.com/YUwXzp4U

I have not been able to reproduce the hang but each time I land on the Camera screen I've got this warning about unsupported layout (it only appears in landscape mode), which may or may not be related. 

```
14:05:54.727: [info] 📸 VisionCamera.didSetProps(_:): Updating 25 props: [onInitialized, cameraId, codeScannerOptions, enableBufferCompression, onOutputOrientationChanged, onStarted, onCodeScanned, position, preview, top, right, isActive, isMirrored, resizeMode, onError, onStopped, onViewReady, onPreviewOrientationChanged, onPreviewStopped, enableFrameProcessor, onPreviewStarted, left, bottom, outputOrientation, onShutter]
14:05:54.729: [info] 📸 VisionCamera.configurePreviewOrientation(_:): Updating Preview rotation: landscapeRight...
14:05:54.730: [info] 📸 VisionCamera.configureOutputOrientation(_:): Updating Outputs rotation: landscapeRight...
14:05:54.730: [info] 📸 VisionCamera.configure(_:): configure { ... }: Waiting for lock...
14:05:54.736: [info] 📸 VisionCamera.configure(_:): configure { ... }: Updating CameraSession Configuration... Difference(inputChanged: true, outputsChanged: true, videoStabilizationChanged: true, orientationChanged: true, formatChanged: true, sidePropsChanged: true, torchChanged: true, zoomChanged: true, exposureChanged: true, audioSessionChanged: true, locationChanged: true)
14:05:54.736: [info] 📸 VisionCamera.configureDevice(configuration:): Configuring Input Device...
[...]

PR #4141, opened 2026-08-07
fix(ios): serialize preview layer attach/detach on the session queue

### LINKED ISSUE #4098: 🐛 iOS: permanent main-thread deadlock in HybridPreviewView.updatePreviewLayer (addSublayer vs AVCaptureSession lock, follow-up to #3356, still in 5.1.1)
### Prerequisites

- I searched existing issues: #3356 is the same signature (bottom 4 frames identical) on v4.5.2, closed 2026-04-21 without a fix. This report shows the deadlock is still present in v5.1.1 and adds the second thread of the deadlock pair plus a lock-ordering analysis, which #3356 lacked.

### Reproduction

Not deterministically reproducible: it is a race on every `<Camera>` mount, observed in production at roughly 0.1% of mounts on a fleet of kiosk iPads (8 fatal hangs in 5 days across 3 devices, captured by Sentry's fatal-app-hang tracking with full thread dumps). Mount/unmount stress loops of a Camera screen should eventually reproduce it.

### Steps to reproduce

1. Mount a `<Camera>` with a preview while the session is already configured and starting (any remount of a camera screen exercises this: `usePreviewOutput` is memoized per mount, and `HybridPreviewView.previewOutput didSet` calls `updatePreviewLayer()`).
2. `updatePreviewLayer()` runs `view.layer.addSublayer(previewLayer)` on the main thread with the layer's `session` already attached.
3. Race window: if AVFoundation is concurrently delivering KVO for that preview layer on the session queue, the app deadlocks permanently.

### What did you expect to happen?

Preview attaches; no main-thread block.

### What actually happened?

Permanent main-thread freeze (kiosk dead until force-kill). ABBA deadlock:

- Main thread holds the CoreAnimation transaction/layer lock (inside `addSublayer`) and waits on the capture session's `@synchronized` lock in `-[AVCaptureVideoPreviewLayer layerDidBecomeVisible:]`.
- An AVFCapture KVO thread holds the session lock and commits a CoreAnimation transaction (`CA::Transaction::commit` from the background queue), waiting on the CA lock.

Each holds what the other needs.

### Affected platforms

iOS (iPadOS 17.6, 18.7, 26.0; iPad7,11 and iPad13,18; landscape-mounted devices, matching the landscape-only observation in #3356)

### Device(s) affected

iPad7,11 (7 events), iPad13,18 (1 event)

### VisionCamera version

5.1.1 (note: `HybridPreviewView.swift` on current `main` is byte-identical, so main is affected too)

### React Native version

0.86 (Expo SDK 57)

### React Native architecture

New Architecture (Fabric + Nitro)

### Features being used

Preview, Photo capture, Frame processor (face detection)

### Relevant logs / stack trace

Thread 0 (main), identical in all 8 events:

```
ClockSmart :: closure in HybridPreviewView.updatePreviewLayer
QuartzCore :: -[CALayer addSublayer:]
QuartzCore :: CA::Layer::insert_sublayer / update_added_sublayer / mark_visible
AVFCapture :: -[AVCaptureVideoPreviewLayer layerDidBecomeVisible:]
libobjc    :: objc_sync_enter
libsystem_platform :: _os_unfair_lock_lock_slow
libsystem_kernel   :: __ulock_wait2
```

Thread 13 (same event):

```
Foundation :: NSKeyValueDidChange / NSKeyValueNotifyObserver
AVFCapture :: -[AVCaptureVideoPreviewLayer observeValueForKeyPath:ofObject:change:...]
QuartzCore :: CA::Transaction::commit
QuartzCore :: CA::Context::commit_transaction
QuartzCore :: CA::Layer::layout_and_display_if_needed / layout_if_needed
libsystem_platform :: _os_unfair_lock_lock_slow
libsystem_kernel   :: __ulock_wait2
```

### Additional information

Root cause is the lock ordering in `HybridPreviewView.updatePreviewLayer()`: `addSublayer` is called while the preview layer is session-connected, so `layerDidBecomeVisible:` must take the session lock under the CA lock, inverting the order used by AVF's own KVO path (session lock -> CA commit).

Workaround we are shipping via patch-package (happy to PR if you want this shape): when `previewLayer.session != nil && previewLayer.connection != nil` (i.e. `configure()` has finished; guards against racing the session-queue configure batch), snapshot `connection.inputPorts.first`, set `previewLayer.session = nil`, do the CA work (addSublayer/frame/videoGravity), then re-attach with `setSessionWithNoConnection(_:)` and rebuild the connection with `AVCaptureConnection(inputPort:videoPreviewLayer:)` inside `beginConfiguration`/`commitConfiguration`, mirroring `AVCaptureSession.addConnection(input:output:)`. The session lock is then never awaited while the CA lock is held. A plain `.session =` restore is NOT safe: it destroys the manually built preview connection and forms an implicit one (or none).

m00kid:
Fixes #4098 (same cycle as #3356).

### The deadlock

`addSublayer()` takes the CoreAnimation transaction lock, and AVFoundation then waits on the capture session lock inside `layerDidBecomeVisible()`. Meanwhile the AVFCapture KVO thread holds the session lock and commits a CA transaction, waiting on the CA lock. ABBA, and the main thread hangs permanently.

### Why not just detach on the main thread

That breaks the cycle but races `CameraSession`'s serial queue. `AVCaptureSession` is not thread-safe, so a main-thread `beginConfiguration`/`commitConfiguration` pair interleaving with `configure()`/`start()` gave us `startRunning may not be called between calls to beginConfiguration and commitConfiguration` traps plus `attach/detachFromFigCaptureSession` assertions.

### What this does

All session mutation moves onto `HybridCameraSession.queue`, the same serial queue `configure()` and `start()` already use. Only pure CoreAnimation work stays on the main thread, so the main thread never waits for the session lock and no two threads ever configure the session concurrently.

The preview connection is rebuilt the way `CameraSession` does it (`setSessionWithNoConnection` plus an explicit `AVCaptureConnection`), since the plain `.session` setter forms an implicit connection to the wrong port. The re-attach is a no-op if a `configure()` ran while the layer was detached and re-attached it already.

`HybridCameraSession.queue` becomes internal so the preview view can share it.

### Testing

Running in production since late July on a fleet of unattended iOS kiosks (iOS 17.6 through 26.0, iPad and iPhone). The app-hang cluster this was causing has been at zero since.

Note #4134 does not cover this one: `HybridPreviewView.swift` is byte-identical between 5.1.1 and 5.2.2, so `updatePreviewLayer` and the lock ordering are unchanged.

### DIFF
--- a/packages/react-native-vision-camera/ios/Hybrid Objects/HybridCameraSession.swift
+++ b/packages/react-native-vision-camera/ios/Hybrid Objects/HybridCameraSession.swift
@@ -10,7 +10,11 @@ import NitroModules
 
 final class HybridCameraSession: HybridCameraSessionSpec {
   let session: AVCaptureSession
-  private static let queue: DispatchQueue = DispatchQueue(
+  // Internal (was private) so HybridPreviewView can serialize its
+  // preview-layer attach/detach with configure()/start(). AVCaptureSession is
+  // not thread-safe; every session mutation in this module must go through
+  // this one serial queue.
+  static let queue: DispatchQueue = DispatchQueue(
     label: "com.margelo.camera.session",
     qos: .userInteractive,
     attributes: [],
--- a/packages/react-native-vision-camera/ios/Hybrid Objects/Views/HybridPreviewView.swift
+++ b/packages/react-native-vision-camera/ios/Hybrid Objects/Views/HybridPreviewView.swift
@@ -126,43 +126,142 @@ final class HybridPreviewView: HybridPreviewViewSpec {
   private func updatePreviewLayer() {
     DispatchQueue.main.async {
       if let previewLayer = self.previewLayer {
-        // Remove all sublayers that are not our AVCaptureVideoPreviewLayer
-        self.view.layer.sublayers?.removeAll { $0 != previewLayer }
-        if self.view.layer.sublayers?.contains(previewLayer) != true {
-          // If we don't have the preview layer, we add it here
-          self.view.layer.addSublayer(previewLayer)
-        }
-        previewLayer.frame = self.view.bounds
-
-        // Update resizeMode
-        let resizeMode = self.resizeMode ?? .cover
-        previewLayer.videoGravity = resizeMode.toAVLayerVideoGravity()
-
-        // Add listener to `isPreviewing`
-        self.isPreviewingObserver = previewLayer.observe(
-          \.isPreviewing,
-          options: [.initial, .new, .old],
-          changeHandler: { [weak self] _, change in
-            guard let self else { return }
-            guard let wasPreviewing = change.oldValue,
-              let isPreviewing = change.newValue
-            else { return }
-            guard wasPreviewing != isPreviewing else { return }
-            if isPreviewing {
-              logger.info("PreviewView started!")
-              self.onPreviewStarted?()
-            } else {
-              logger.info("PreviewView stopped!")
-              self.onPreviewStopped?()
+        // Never touch the AVCaptureSession from the main thread.
+        //
+        // The deadlock this avoids (#4098, and #3356): addSublayer() takes
+        // the CoreAnimation transaction lock and AVFoundation then waits on
+        // the capture session lock inside layerDidBecomeVisible:, while the
+        // AVFCapture KVO thread holds the session lock and waits on the CA
+        // lock. Classic ABBA, and it hangs the main thread permanently.
+        //
+        // Detaching and re-attaching the session on the MAIN thread does fix
+        // that cycle, but it then races CameraSession's serial queue:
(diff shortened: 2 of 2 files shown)


mrousavy:
Thanks for your efforts to make a PR, but this is not a good fix at all and violates the encapsulation of each Camera Output, which now accesses the Capture Session directly - using private APIs (which are now public; e.g. queue).

This is a sloppy AI workaround, not a fix.

m00kid:
Got it @mrousavy, fair enough. The deadlock in #4098 is real and reproducible on our side regardless, so hopefully that part is still useful for a proper fix.

mrousavy:
Yea for sure I think if we have a repro I'd love to keep this open and find a proper fix, but as long as we don't I can't really do anything

PR #4143, opened 2026-08-10
fix(android): compose Photo.toImage() rotation before mirroring

dazakdev:
## What

`HybridPhoto.toImage()` builds its transform with `preScale(-1f, 1f)` for the mirror and `postRotate(...)` for the orientation. `preScale` right-multiplies and `postRotate` left-multiplies, so the composition is mirror-first. A reflection conjugates a rotation into its inverse, so the two possible orderings differ by twice the rotation — a half turn at the quarter-turn orientations a portrait capture produces, and nothing at all at `up` or `down`.

The effect is that `toImage()` disagrees by 180° with `saveToTemporaryFile()`, with `capturePhotoToFile()` and with the live preview, for one and the same capture. `orientation` and `isMirrored` are both reported correctly — only the composition is wrong.

CameraX composes it the other way round: `Exif.flipHorizontally()` maps orientation 6→5 and 8→7, which is a flip in display space, and `FileUtil.updateFileExif` calls `Exif.rotate(...)` before `Exif.flipHorizontally()`. `HybridPhoto.attachExifData` follows that same order, so before this change the two ways out of one `Photo` contradicted each other.

The same composition appears a second time in `ImageProxy.toBitmap(orientation, isMirrored)`, which `HybridFrameConverter.convertFrameToImage` and `convertDepthToImage` both go through. Fixed there too, since it is the same defect reached from a different entry point.

## Fix

Apply both transforms with `post*`, so they run in written order: rotate upright, then mirror.

## Test

`visioncamera.photo.harness.ts` gains one case. It captures a mirrored front-camera photo, saves it, and reads the stored frame back with `Images.loadFromFileAsync`. For a grid of sample points it maps each rendered point back into the stored frame through the rotation the `Photo` reports, applied before the mirror, and averages the absolute difference.

Both images decode the same JPEG, so a correct rendering is pixel identical and the mean difference is exactly `0` — the scene in front of the camera does not matter, which should keep it stable on a taped Device Farm sensor. Composing the other way round lands every sample a half turn away, on unrelated pixels.

Verified on an emulator, following the procedure in the tests README:

- on `main`: `expected 49.90518518518518 to be close to +0`
- with the fix: passes

The test skips, with a reason, in the three cases where both orderings are the same matrix and there is nothing to observe: when the reported orientation is not a quarter turn, when the photo is not mirrored, and when the platform's decoder has already applied the orientation to the stored frame.

## Notes

**This is device-dependent.** The rotation branch only runs when `photo.orientation` is not `up`, and that value comes from `ImageProxy.imageInfo.rotationDegrees`. CameraX can honour the requested orientation either by rotating the pixels in the pipeline — reporting `0` — or by leaving them and reporting the degrees; only the second path reaches this code. A Huawei P40 (Android 10) always reports `up` while the photo dimensions swap, so the defect cannot be produced on it at all. An emulator with a host webcam as the front camera reports `left` and reproduces it.

Setting `photoOutput.outputOrientation` is not a way to force the case: CameraX then takes the bake-the-rotation path and the reported orientation drops back to `up`. The test therefore uses the orientation as the pipeline reports it, and skips rather than pretending to cover the case.

**No platform guard.** The behaviour should hold on both platforms and iOS composes it correctly today. On an iOS simulator the test currently skips, because the connection orientation defaults to portrait and the reported orientation is `up`.

**A question, not a claim, about the rotation direction.** `ImageProxy.toBitmap` rotates by `orientation.degrees` while `HybridPhoto.toImage` rotates by `orientation.counterRotated().degrees`. Both are handed the same value — `HybridFrameOutput` passes `image.orientation` and `HybridPhoto.orientation` is `image.orientation` — so for a frame reported as `left` one path turns 90° and the other 270°. I only measured the photo path, where `counterRotated()` is what agrees with the EXIF file and the preview; I have not run the frame converter, and the comment in `HybridFrameOutput` about orientation being "relative to current target orientation" leaves room for `ImageAnalysis` following a different convention. Left untouched here — happy to follow up separately once you say which is intended.

An image from the emulator:

Screenshot 2026-08-10 at 12 44 20

### DIFF
--- a/apps/simple-camera/__tests__/visioncamera.photo.harness.ts
+++ b/apps/simple-camera/__tests__/visioncamera.photo.harness.ts
@@ -9,6 +9,7 @@ import {
   waitUntil,
 } from 'react-native-harness'
 import type { Image } from 'react-native-nitro-image'
+import { Images } from 'react-native-nitro-image'
 import type {
   CameraDevice,
   CameraDeviceFactory,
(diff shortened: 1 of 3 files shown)


mrousavy:
Thanks!

PR #4164, opened 2026-08-20
feat: Upgrade Nitro to 0.37 for View improvements

mrousavy:


package.json
```diff
@@ -125,6 +125,9 @@
       }
     }
   },
+  "overrides": {
+    "react-native-nitro-modules": "0.37.0-beta.1"
+  },
```
mrousavy: This has to go before we merge.

PR #4168, opened 2026-08-21
fix(android): Resolve every permission request when they run in parallel

### LINKED ISSUE #3834: Android: parallel requestPermission calls leak coroutines and surface as 'JPromise was destroyed'
Opening the camera in our app fires camera, microphone, and location permission requests from the same `useEffect`. On Android the JS console gets these almost immediately:

```
ERROR  [Error: Uncaught (in promise, id: 0) Error: java.lang.RuntimeException: Timeouted: JPromise was destroyed!
        at com.facebook.jni.HybridData$Destructor.deleteNative(Native Method)
        at com.facebook.jni.HybridData$Destructor.destruct(HybridData.java:82)
        at com.facebook.jni.DestructorThread$1.run(DestructorThread.java:78)
]
ERROR  [Error: Uncaught (in promise, id: 1) Error: java.lang.RuntimeException: Timeouted: JPromise was destroyed!
... (same)
ERROR  [Error: Uncaught (in promise, id: 2) Error: java.lang.RuntimeException: Timeouted: JPromise was destroyed!
... (same)
```

The dialogs themselves show, user grants/denies, and the hook selectors (`hasPermission` / `status`) update correctly. Only the `requestPermission()` promises hang.

### Setup

```ts
const camera = useCameraPermission();
const mic = useMicrophonePermission();
const location = useLocation(); // from react-native-vision-camera-location

useEffect(() => {
  if (!visible) return;
  if (camera.status === 'not-determined') void camera.requestPermission();
  if (mic.status === 'not-determined') void mic.requestPermission();
  if (!location.hasPermission) void location.requestPermission();
}, [visible, camera, mic, location]);
```

| Package | Version |
| --- | --- |
| react-native-vision-camera | 5.0.8 |
| react-native-vision-camera-location | 5.0.8 |
| react-native-nitro-modules | 0.35.6 |
| react-native | 0.83.6 |
| Tested on | Android 14 + Android 15 |

### Diagnosis

`packages/react-native-vision-camera/android/src/main/java/com/margelo/nitro/camera/extensions/ReactApplicationContext+permissions.kt`:

```kotlin
suspend fun ReactApplicationContext.requestPermission(permission: String): Boolean {
  return suspendCoroutine { continuation ->
    val activity = currentActivity ?: throw Error("No Activity!")
    if (activity is PermissionAwareActivity) {
      ...
      val currentRequestCode = permissionRequestCode++
      val listener = PermissionListener { requestCode, _, grantResults ->
        if (requestCode == currentRequestCode) {
          ...
          continuation.resume(hasPermission)
          return@PermissionListener true
        }
        return@PermissionListener false
      }
      activity.requestPermissions(arrayOf(permission), currentRequestCode, listener)
    }
  }
}
```

Each call passes its own `PermissionListener`, but RN's `PermissionAwareActivity` only stores one in-flight `PermissionListener` field. Calling `requestPermissions(...)` again before the first result comes back overwrites the previous listener. With three parallel calls, listener A is replaced by B, B is replaced by C. Only listener C's continuation resumes when the user finishes the dialogs; A and B's coroutines suspend forever, the JPromises stay pending until GC, and Nitro's `JPromise.hpp` destructor produces the rejection above:

```cpp
~JPromise() override {
  if (isPending()) [[unlikely]] {
    std::runtime_error error("Timeouted: JPromise was destroyed!");
    this->reject(...);
  }
}
```

The "Timeouted" wording is misleading; there's no actual timeout, it's the destructor catching a never-resolved promise.

### Suggested fix

Keep a `Map<Int, Continuation<Boolean>>` keyed by `requestCode` and a single dispatcher `PermissionListener` that resumes whichever continuation matches the incoming code. Sketch:

```kotlin
private val pendingContinuations = ConcurrentHashMap<Int, Continuation<Boolean>>()
private val dispatcher = PermissionListener { requestCode, _, grantResults ->
  val cont = pendingContinuations.remove(requestCode) ?: return@PermissionListener false
  cont.resume(grantResults.firstOrNull() == PackageManager.PERMISSION_GRANTED)
  true
}
```

`requestPermission(...)` then registers its continuation in the map and passes the same dispatcher every time, so `PermissionAwareActivity`'s single-listener slot only ever holds the dispatcher.

Alternatively serialize with a `Mutex` so requests run one at a time. Simpler but the dialogs show sequentially which is worse UX.

### Workaround for now

Catching the rejection silences the noise, since the hook's `hasPermission` / `status` is the actual source of truth:

```ts
if (camera.status === 'not-determined') void camera.requestPermission().catch(() => {});
if (mic.status === 'not-determined')    void mic.requestPermission().catch(() => {});
if (!location.hasPermission)            void location.requestPermission().catch(() => {});
```

Happy to send a PR with the dispatcher approach if it's the direction you want to take.

dennytosp:
Fixes #3834

## What

Requesting more than one permission at a time on Android — e.g. camera, microphone and location from the same `useEffect` — leaves every `requestPermission()` Promise hanging. Nitro eventually destroys them, and the app sees:

```
ERROR  [Error: Uncaught (in promise, id: 0) Error: java.lang.RuntimeException: Timeouted: JPromise was destroyed!]
```

The dialogs themselves work and `hasPermission` / `status` end up correct — only the Promises never settle.

## Root cause

Two single-slot fields, one in React Native and one in Android itself. Routing results by request code alone does not fix it — the requests also have to actually run one at a time.

**1. `PermissionAwareActivity` only remembers one `PermissionListener`.**

[`ReactActivityDelegate.requestPermissions`](https://github.com/facebook/react-native/blob/v0.85.3/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/ReactActivityDelegate.java#L244-L248) overwrites the field on every call:

```java
public void requestPermissions(String[] permissions, int requestCode, @Nullable PermissionListener listener) {
    mPermissionListener = listener;
    getPlainActivity().requestPermissions(permissions, requestCode);
}
```

VisionCamera created a fresh `PermissionListener` per request, so listener A was replaced by B and B by C. Only the last one ever ran; the earlier continuations stayed suspended forever.

The same delegate also drops the listener as soon as one returns `true`, and buffers the result in a single `mPermissionsCallback` field while a dialog is up — so an overlapping request can lose a result there too.

**2. `Activity.requestPermissions(...)` refuses concurrent requests.**

[AOSP `Activity.java`](https://github.com/aosp-mirror/platform_frameworks_base/blob/android15-release/core/java/android/app/Activity.java#L5711-L5716):

```java
if (mHasCurrentPermissionsRequest) {
    Log.w(TAG, "Can request only one set of permissions at a time");
    // Dispatch the callback with empty arrays which means a cancellation.
    onRequestPermissionsResult(requestCode, new String[0], new int[0], deviceId);
    return;
}
```

The second request is cancelled with empty grant results before the user ever sees it. The old code mapped that empty array to "denied" — and since `shouldShowRequestPermissionRationale(...)` is `false` for a permission that was never asked, `PermissionStateStore` would then persist it as *permanently denied*.

## Fix

A new `PermissionRequestDispatcher` (`android/.../camera/extensions/PermissionRequestDispatcher.kt`) owns all of the plumbing:

- **one shared `PermissionListener`**, with the per-request state kept here in a `ConcurrentHashMap<Int, CancellableContinuation<IntArray>>` keyed by request code, so no request can clobber another one's listener,
- **a `Mutex`**, so Android only ever sees one in-flight request and never hits the "only one set of permissions at a time" cancellation,
- the listener only returns `true` — which makes React Native drop the shared listener again — once nothing is in flight anymore,
- cancelling the coroutine removes its pending entry, and a `requestPermissions(...)` that throws resumes its caller with that error instead of leaking it.

`ReactApplicationContext.requestPermission(...)` keeps the exact same signature and still owns the `PermissionStateStore` bookkeeping, so `HybridCameraFactory` and `HybridLocationManager` are untouched — none of this leaks into surface-level code.

One small behavior change worth calling out: an empty (cancelled) grant result no longer records a *permanently denied* state, since nothing was actually asked. Happy to drop that if you'd rather keep the diff strictly to the hang.

## Tests

New `apps/simple-camera/__tests__/visioncamera.permissions.harness.ts`, plus its row in the `__tests__` README layout table:

- `resolves camera and microphone requests that are started in parallel`
- `resolves every request when the same permission is requested multiple times at once`

Both are shared tests with no platform guard — the behavior should hold everywhere, and iOS passes them today. On Android before this change they hang and fail on `withTimeout(...)`. The harness grants permissions on install (`permissions: true` in `rn-harness.config.mjs`), so both requests are expected to resolve `true`, same as the existing `expect(cameraPermissionStatus).toBe('authorized')` in the hooks suite.

No JUnit tests / no new Gradle test dependencies, per the Harness-only convention.

## Verification

- `bun run lint-kotlin` — clean, no reformatting
- `biome check` — clean on the new test file
- `tsc --noEmit` — no errors in the new test file
- `./gradlew :react-native-vision-camera:compileDebugKotlin` — compiles
- No Android device on hand for a local Harness run, so the on-device pass is up to CI.


packages/react-native-vision-camera/android/src/main/java/com/margelo/nitro/camera/extensions/PermissionRequestDispatcher.kt
```diff
@@ -0,0 +1,63 @@
...
+
+  private val listener =
+    PermissionListener { requestCode: Int, _: Array<String>, grantResults: IntArray ->
+      val continuation = pendingRequests.remove(requestCode) ?: return@PermissionListener false
+      if (continuation.isActive) {
+        continuation.resume(grantResults)
+      }
+      // Returning `true` makes React Native drop the shared listener, so only give the slot up once
+      // there is no request left that still needs its result delivered.
+      return@PermissionListener pendingRequests.isEmpty()
```
mrousavy: Returning `true` tells React Native to clear the listener, correct, but what if a second call to ask for Permissions arrives in the meantime? Then this one here clears the listener and the second call never gets resolved, ultimately ending up at the same bug that this PR attempted to fix, no? (not tested so not verified, just assumption)

mrousavy: Codex suggested a possible fix to complete/resume on the `mainHandler`, but I think that might be a bit of an internal assumption and could end up being flakey too, unless we can absolutely narrow this down in a clear Harness test (that isnt affected by flakey race conditions).

Maybe we can find a better/solid solution :)

packages/react-native-vision-camera/android/src/main/java/com/margelo/nitro/camera/extensions/ReactApplicationContext+permissions.kt
```diff
@@ -43,34 +40,27 @@ fun ReactApplicationContext.getPermissionStatus(permission: String): PermissionS
...
+    throw Error("Activity is not a PermissionAwareActivity!")
+  }
+
+  PermissionStateStore.setHasRequestedPermission(this, permission, true)
+  val grantResults = PermissionRequestDispatcher.request(activity, permission)
+  if (grantResults.isEmpty()) {
+    // Android cancelled the request without ever asking the user, so we did not learn anything
+    // new about this permission - don't remember it as denied.
+    return false
+  }
```
mrousavy, line 60: This is also problematic because later we ask `shouldShowRequestPermissionRationale` which will be `false` even though this has never been presented to the user.

What kind of state is that? Is that a bug?

packages/react-native-vision-camera/android/src/main/java/com/margelo/nitro/camera/extensions/ReactApplicationContext+permissions.kt
```diff
@@ -43,34 +40,27 @@ fun ReactApplicationContext.getPermissionStatus(permission: String): PermissionS
...
+
+  val hasPermission = grantResults.first() == PackageManager.PERMISSION_GRANTED
```
mrousavy: Code-style nit-pick, but we should probably use `grantResults.singleOrNull()`, then if it's `null` we can see that it was canceled or never shown (throw here? or reject? it is technically a cancelation or a bug or whatever), and if it's not null it has to be a single value, no? 

Because what if there are multiple `grantResults` in here? That'd also be a bug we want to protect against - hence `singleOrNull()`.

mrousavy (changes requested):
Thanks for attempting to fix this! A few nit picks before we can merge

PR #4169, opened 2026-08-22
fix: Restore CameraDevice neutralZoom

### LINKED ISSUE #3845: Neutral zoom?
### Prerequisites

- [x] I have searched [existing issues](https://github.com/mrousavy/react-native-vision-camera/issues?q=is%3Aissue) for a similar request.

### What problem are you trying to solve?

This isn't a feature request, but it didn't fit the bug template. Not sure whether this is a bug, a docs issue, or user error.

On an iOS device with an ultra-wide-angle camera:
```
const device = useCameraDevice('back', { physicalDevices: ['ultra-wide-angle', 'wide-angle', 'telephoto'] })
console.log(device.minZoom)
console.log(device.zoomLensSwitchFactors)
```
### Proposed API / solution

The [docs on zooming](https://visioncamera.margelo.com/docs/zooming#default-zoom) say:

> If a virtual Camera contains an ['ultra-wide-angle'](https://visioncamera.margelo.com/api/react-native-vision-camera/type-aliases/DeviceType) Camera (the "0.5x Camera"), the user may zoom out to less than 1 (here 0.5) instead.
> 
> [minZoom](https://visioncamera.margelo.com/api/react-native-vision-camera/hybrid-objects/CameraDevice#minzoom) <= 1 (e.g. 0.5x)
> natural zoom = 1
> [maxZoom](https://visioncamera.margelo.com/api/react-native-vision-camera/hybrid-objects/CameraDevice#maxzoom) >= 1 (e.g. 3x)

And the section on [zoomLensSwitchFactors](https://visioncamera.margelo.com/api/react-native-vision-camera/hybrid-objects/CameraDevice#zoomlensswitchfactors) says:

> const camera = ...           // Triple-[Camera](https://visioncamera.margelo.com/api/react-native-vision-camera/views/Camera) (0.5x, 1x, 3x)
> camera.zoomLensSwitchFactors // [1, 3]

But the actual output is:

```
console.log(device.minZoom) // 1
console.log(device.zoomLensSwitchFactors) // [2, 8]
```

Is this the expected output?

In vision camera v4, there was a [neutralZoom property](https://visioncamera4.margelo.com/docs/api/interfaces/CameraDevice#neutralzoom), but that doesn't exist in v5. If `minZoom` is correctly returned as 1, how is the application meant to determine that `neutralZoom` should be 2?

### Alternatives considered

_No response_

### Platforms this should target

iOS

### Would you be willing to contribute this?

No, I'm requesting only.

### Additional context

_No response_

### Submission

- [x] I described the problem, not just the solution.
- [x] I wrote this request in my own words. I did not paste AI-generated proposals.

ngocdevv:
## What

VisionCamera v5 no longer exposes `CameraDevice.neutralZoom`. On iOS virtual Cameras with an ultra-wide constituent, AVFoundation uses an internal zoom scale where `minZoom` can be `1` while the standard wide-angle lens is reached at `2`. Consumers therefore cannot initialize at the natural user-facing `1x` value without recreating native lens-selection logic.

This restores `neutralZoom` so consumers can start on the standard wide-angle lens and still allow zooming out to ultra-wide.

Closes #3845

## How

- Adds `CameraDevice.neutralZoom` to the Nitro spec and regenerates the Swift, Kotlin, and C++ bindings.
- On iOS 18+, converts AVFoundation display zoom back to its internal factor using `displayVideoZoomFactorMultiplier`.
- On older iOS versions, uses the virtual-device switchover factor immediately before the standard wide-angle constituent, matching the v4 behavior.
- Returns `1.0` on Android and for physical Cameras.
- Updates the zoom guide and adds Harness capability invariants for the restored property.

## Test

- `bun camera specs`
- `bun run build`
- `bunx tsc --noEmit -p apps/simple-camera/tsconfig.json`
- `bun example build:android` (Kotlin, JNI/C++, four ABIs, debug APK)
- `xcodebuild -workspace apps/simple-camera/ios/SimpleCamera.xcworkspace -scheme SimpleCamera -configuration Debug -destination generic/platform=iOS\ Simulator CODE_SIGNING_ALLOWED=NO build`
- `bun lint-swift`
- `bun lint-kotlin`
- `bun lint-js`
- `bun docs types:check`
- `bun docs check:links`

### DIFF
--- a/packages/react-native-vision-camera/ios/Extensions/AVFoundation/AVCaptureDevice+neutralZoom.swift
+++ b/packages/react-native-vision-camera/ios/Extensions/AVFoundation/AVCaptureDevice+neutralZoom.swift
@@ -1,0 +1,40 @@
+///
+/// AVCaptureDevice+neutralZoom.swift
+/// VisionCamera
+/// Copyright © 2026 Marc Rousavy @ Margelo
+///
+
+import AVFoundation
+import Foundation
+
+extension AVCaptureDevice {
+  /**
+   * The internal video zoom factor that AVFoundation displays as `1x`.
+   *
+   * On older iOS versions, derive it from the switchover factor immediately
+   * before the standard wide-angle constituent Camera.
+   */
+  var neutralZoomFactor: Double {
+    if #available(iOS 18.0, *) {
+      let multiplier = displayVideoZoomFactorMultiplier
+      if multiplier > 0 {
+        return 1 / multiplier
+      }
+    }
+
+    guard
+      let wideAngleIndex = constituentDevices.firstIndex(where: {
+        $0.deviceType == .builtInWideAngleCamera
+      }),
+      wideAngleIndex > 0
+    else {
+      return 1
+    }
+
+    let switchoverIndex = wideAngleIndex - 1
+    guard virtualDeviceSwitchOverVideoZoomFactors.indices.contains(switchoverIndex) else {
+      return 1
+    }
+    return virtualDeviceSwitchOverVideoZoomFactors[switchoverIndex].doubleValue
+  }
+}
--- a/apps/simple-camera/__tests__/visioncamera.devices.harness.ts
+++ b/apps/simple-camera/__tests__/visioncamera.devices.harness.ts
@@ -94,6 +94,8 @@ describe('VisionCamera - Devices', () => {
   it('reports sane capability invariants for each device', () => {
     for (const device of factory.cameraDevices) {
       expect(device.minZoom).toBeLessThanOrEqual(device.maxZoom)
+      expect(device.neutralZoom).toBeGreaterThanOrEqual(device.minZoom)
+      expect(device.neutralZoom).toBeLessThanOrEqual(device.maxZoom)
 
       if (device.supportsExposureBias) {
         expect(device.minExposureBias).toBeLessThanOrEqual(
--- a/docs/content/docs/zooming.mdx
+++ b/docs/content/docs/zooming.mdx
@@ -13,7 +13,7 @@ The [`<Camera />`](/api/react-native-vision-camera/views/Camera) view allows adj
 ```tsx
 function App() {
   // [!code ++]
-  const zoom = useSharedValue(device.minZoom)
+  const zoom = useSharedValue(device.neutralZoom)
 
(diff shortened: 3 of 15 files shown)


packages/react-native-vision-camera/ios/Extensions/AVFoundation/AVCaptureDevice+neutralZoom.swift
```diff
@@ -0,0 +1,40 @@
...
+extension AVCaptureDevice {
+  /**
+   * The internal video zoom factor that AVFoundation displays as `1x`.
+   *
+   * On older iOS versions, derive it from the switchover factor immediately
+   * before the standard wide-angle constituent Camera.
+   */
+  var neutralZoomFactor: Double {
+    if #available(iOS 18.0, *) {
+      let multiplier = displayVideoZoomFactorMultiplier
```
mrousavy, line 19: This value changes as the device starts zooming. So if you zoom in, and then ask `device.neutralZoomFactor`, it is no longer `1` (or `2`) but might be any arbitrary number.
_I think_.

packages/react-native-vision-camera/ios/Extensions/AVFoundation/AVCaptureDevice+neutralZoom.swift
```diff
@@ -0,0 +1,40 @@
...
+      wideAngleIndex > 0
+    else {
+      return 1
+    }
+
+    let switchoverIndex = wideAngleIndex - 1
+    guard virtualDeviceSwitchOverVideoZoomFactors.indices.contains(switchoverIndex) else {
+      return 1
+    }
+    return virtualDeviceSwitchOverVideoZoomFactors[switchoverIndex].doubleValue
```
mrousavy, line 38: Also this Plan B and Plan C approach are kinda ugly. I'm not sure if there is a better way, but as it is, its a bit ugly.

mrousavy (changes requested):
Thanks for this PR.

I still haven't decided yet if we should have a `.neutralZoom`, or if we should just normalize zoom on iOS to also go below `1` so that it is like on Android..? Then zoom start is always 1 and we dont need to worry about anything else

Issue #4114, opened 2026-07-30
🐛 convertCameraPointToFramePoint ignores normalising extents on orientation swap (iOS) — square region maps to a 3.16:1 strip


florianpreusner:
### Prerequisites

- [x] I have read and followed every applicable step in the Nitro Modules Troubleshooting guide.
- [x] I have read the VisionCamera Troubleshooting guide.
- [x] I have searched existing issues and found nothing matching (no hits for `convertCameraPointToFramePoint`).
- [x] I am on the latest version, **or** have a specific reason I cannot upgrade — see "VisionCamera version" below.

### Reproduction

https://github.com/mrousavy/react-native-vision-camera/pull/4113 — adds one failing Harness test, `maps a square Frame region onto a square View region`, to `apps/simple-camera/__tests__/visioncamera.coordinates.harness.tsx`.

### Steps to reproduce

1. Check out the branch in #4113.
2. Run the coordinates harness on an iOS device: `bun run test:harness:ios -- --testPathPatterns=coordinates`.
3. Watch `maps a square Frame region onto a square View region`.
4. Run the same file on an Android device — it passes there.

### What did you expect to happen?

Composing the two public coordinate conversions should preserve aspect. A square region in Frame space should map to a square region in View space:

```
PreviewView.convertViewPointToCameraPoint()  ->  Frame.convertCameraPointToFramePoint()
```

Every transform in between — sensor scale, orientation counter-rotation, the preview's aspect-preserving crop — preserves aspect. Camera space itself is allowed to be anisotropic (iOS normalises per-axis to `[0, 1]`), but converting out of it should undo the same per-axis scaling it applied.

### What actually happened?

On iOS the square comes back as a wide, short strip. Measured on iPhone 15 Pro Max, frame 3840×2160, `orientation: "left"`, using a square guide cell of identical ±19.7 dp in both view axes:

```
sensor Δx 0.0562  ->  frame Δy 121.4 px
sensor Δy 0.0999  ->  frame Δx 383.6 px
```

**383.6 × 121.4 px — a ratio of 3.159**, where a correct mapping measures **1.000**. Pixel 8 (1280×720) composes correctly at 1.000.

`convertViewPointToCameraPoint` looks correct on its own: for the square patch it returns Δx/Δy = 0.5625 = exactly 9/16, consistent with per-axis normalisation over a 16:9 image.

[...]

PR #4175, opened 2026-08-26
fix: align iOS frame coordinates with camera sensor

huytdps13400:
## Summary\n\n- align iOS Frame and Depth coordinate conversion with the unrotated AVFoundation sensor space used by PreviewView\n- retain absolute buffer orientation and mirroring separately from the existing public output-relative metadata\n- preserve the legacy transform for Photo depth data, where no AVCaptureConnection is available\n\nThe regression test was already merged in #4113 and is red on main for iOS while Android passes.\n\n## Verification\n\n- bun camera typecheck\n- swift format lint on all six changed Swift files (exit 0; two pre-existing warnings)\n- xcodebuild -project Pods/Pods.xcodeproj -scheme VisionCamera -sdk iphonesimulator -destination generic/platform=iOS Simulator (arm64 and x86_64)\n- git diff --check\n- affine proof for all four buffer orientations, mirrored and unmirrored, including inverse round-trips\n\nThe full SimpleCamera build is not available locally because the installed Xcode lacks the optional Metal Toolchain required by the unrelated VisionCameraResizer shader target. The existing AWS Device Farm coordinate harness remains the end-to-end oracle for this PR.\n\nFixes #4114

### DIFF
--- a/packages/react-native-vision-camera/ios/Utils/FrameCoordinateSystemConverter.swift
+++ b/packages/react-native-vision-camera/ios/Utils/FrameCoordinateSystemConverter.swift
@@ -13,18 +13,22 @@ enum FrameCoordinateSystemConverter {
    * Get a Matrix that can convert a point in the
    * given `pixelBuffer` to normalized Camera coordinates
    * (`(cx, cy) ∈ [0, 1]²`).
-   * The `orientation` and `isMirrored` flags affect the
-   * Matrix if the `Frame` needs those to be adjusted.
+   * The buffer's physical orientation and mirroring affect
+   * the Matrix if the `Frame` needs those to be adjusted.
    */
   static func getFrameToCameraMatrix(
     pixelBuffer: CVPixelBuffer,
-    orientation: CameraOrientation,
-    isMirrored: Bool
+    bufferOrientation: CameraOrientation,
+    isBufferMirrored: Bool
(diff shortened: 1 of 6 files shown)


packages/react-native-vision-camera/ios/Utils/FrameCoordinateSystemConverter.swift
```diff
@@ -13,18 +13,22 @@ enum FrameCoordinateSystemConverter {
...
+    bufferOrientation: CameraOrientation,
+    isBufferMirrored: Bool
   ) -> CGAffineTransform {
     var matrix = CGAffineTransform.identity
 
-    // 1. Counter-rotate by the orientation to get it up-right
-    switch orientation {
+    // AVFoundation Camera coordinates use the unrotated sensor image, which is always
+    // landscape-right (home button on the right). In CameraOrientation, this is `.left`.
+    let sensorRelativeOrientation = bufferOrientation.relativeTo(.left)
```
mrousavy, line 28: This is simply not true.

mrousavy (changes requested):
Thanks for your PR, but this cannot be accepted as it contains a bunch of unnecessary refactors (naming is better that way, that's correct, but making it non-breaking causes lots of extra code), and one hardcoded orientation modification **which is simply not true in production**. This will break on tons of devices.

PR #4188, opened 2026-09-09
feat(barcode-scanner): expose enableAllPotentialBarcodes

AdarshJais:
<h5 dir="ltr">The gap</h5><p dir="ltr">From JS there is currently no way to tell whether the camera is looking at something barcode-shaped at all. A frame with no Barcode and a frame containing a Barcode that MLKit could not read both return an empty array.</p><p dir="ltr">That makes scan failures unmeasurable. When a user reports "it just doesn't scan", there is nothing to inspect — you cannot distinguish:</p><ul dir="ltr"><li><strong>The detector never found the symbol</strong> — low contrast, poor lighting, out of focus</li><li><strong>The detector found it and could not decode it</strong> — too far away, too few pixels per module</li></ul><p dir="ltr">Those need opposite corrections. More light versus more zoom. Without the distinction, any fix is a guess.</p><h5 dir="ltr">MLKit already provides this — it just isn't reachable</h5><p dir="ltr">From <a href="https://developers.google.com/ml-kit/vision/barcode-scanning/android" target="_blank" rel="noreferrer">Google's Android barcode-scanning guide</a>:</p><blockquote dir="ltr"><p dir="ltr">Starting from bundled model 17.1.0 and unbundled model 18.2.0, you can also call <code>enableAllPotentialBarcodes()</code> to return all potential barcodes <strong>even if they cannot be decoded</strong>.</p></blockquote><p dir="ltr">And from the <a href="https://developers.google.com/android/reference/com/google/mlkit/vision/barcode/BarcodeScannerOptions.Builder" target="_blank" rel="noreferrer"><code>BarcodeScannerOptions.Builder</code> reference</a>:</p><blockquote dir="ltr"><p dir="ltr">Returns all potential barcodes when enabled, even if they cannot be decoded. <code>Barcode.getRawBytes()</code> and <code>Barcode.getRawValue()</code> will return null for any undecoded barcodes, but <code>Barcode.getBoundingBox()</code> will return the area potentially containing a barcode. This can be used to facilitate further detection, e.g., by zooming in the camera to get a clearer image of any barcode in the returned bounding box.</p></blockquote><p dir="ltr">The package already ships MLKit <code>17.3.0</code>, so the capability is present. <code>toMLBarcodeScannerOptions()</code> just builds <code>BarcodeScannerOptions.Builder().setBarcodeFormats(...).build()</code> and nothing else, so it cannot be turned on.</p><h5 dir="ltr">API</h5><div><div><div><div style=""><pre>const scanner = useBarcodeScanner({
  barcodeFormats: ['qr-code'],
  enableAllPotentialBarcodes: true,
})</pre></div><div><div></div></div></div></div></div><p dir="ltr">Also available on <code>useBarcodeScannerOutput</code> / <code>&lt;CodeScanner /&gt;</code>.</p><h5 dir="ltr">What it enables</h5><p dir="ltr">Two things that are impossible today:</p><ol dir="ltr" start="1"><li><strong>Diagnosis.</strong> Every frame can be classified — saw nothing / saw but could not read / read. A failure becomes a measurement instead of an anecdote.</li><li><strong>Auto-zoom.</strong> A code too small to decode still reports where it is, which is exactly the use Google names in the quote above.</li></ol><p dir="ltr"><span><a href="https://github.com/margelo/react-native-vision-camera/issues/3957" target="_blank" rel="noreferrer"><span><span></span>margelo/react-native-vision-camera#3957</span><span></span></a></span> and <span><a href="https://github.com/margelo/react-native-vision-camera/issues/4142" target="_blank" rel="noreferrer"><span><span></span>margelo/react-native-vision-camera#4142</span><span></span></a></span> both report "the scanner misses barcodes that are clearly visible on screen", and both were closed for lack of a reproduction. This flag is what would let a reporter produce one.</p><h5 dir="ltr">Verified against a real failing label</h5><p dir="ltr">A QR printed directly onto kraft paper (common low-contrast delivery packaging), at 1280×720 / ~30fps:</p><div><div>
  |  
-- | --
Frames MLKit located the code on | ~13%
Continuous time in view | 30.1s
Successful decodes | 0

</div></div><p dir="ltr">Before this flag that session produced no signal whatsoever. With it, the 13% detection rate says the problem is <strong>detection</strong>, not decoding — which points at contrast and focus rather than zoom, and would have led to the wrong fix without it.</p><p dir="ltr">Also verified on a deliberately masked QR: located continuously, decoded only once uncovered.</p><h5 dir="ltr">Platform support</h5><p dir="ltr"><strong>Android only.</strong> <code>MLKitBarcodeScanning.BarcodeScannerOptions</code> on iOS exposes only <code>formats</code> — there is no equivalent. Documented with <code>@note</code>, matching the existing pattern on <code>CameraExtension</code> ("only supported on Android") and <code>CameraDevice.supportsFocus</code> ("On non iOS platforms this is always <code>false</code>").</p><h5 dir="ltr">Notes</h5><ul dir="ltr"><li>Defaults to <code>false</code> — no behaviour change for existing users.</li><li>Generated Nitro files updated with <code>bun scanner specs</code>; the diff is purely additive (<code>std::optional&lt;bool&gt;</code> / <code>Boolean?</code>).</li><li><code>bun lint-kotlin</code> and Biome both clean.</li><li>Docs section added to <code>barcode-scanner.mdx</code>.</li></ul><h5 dir="ltr">Follow-up (not in this PR)</h5><p dir="ltr"><code>setZoomSuggestionOptions</code> is the natural companion — it consumes these bounding boxes to auto-zoom. Kept separate to keep this reviewable.</p>The gap


docs/content/docs/barcode-scanner.mdx
```diff
@@ -129,6 +129,33 @@ async function scanBarcodeImage() {
...
+is `undefined` while [`boundingBox`](/api/react-native-vision-camera-barcode-scanner/hybrid-objects/Barcode#boundingbox)
+describes where the potential Barcode is. That tells apart two failures which need
+opposite fixes - a Barcode that cannot be *seen* (lighting, contrast, focus) from one
+that cannot be *read* (too far away, or too few pixels per module) - and lets you act
+on the second by zooming towards the reported region.
+
+> [!NOTE]
+> This is an Android-only option. The iOS MLKit SDK exposes no equivalent, and setting
+> it there has no effect.
+
```
mrousavy: This AI-generated docs section is bad. It will have to be re-written.

As for iOS support, why doesn't MLKit provide that?

AdarshJais: Fair, rewrote it, was way too wordy for the rest of the page.

On iOS: MLKit doesn't expose it there. The [iOS BarcodeScannerOptions](https://developers.google.com/ml-kit/reference/swift/mlkitbarcodescanning/api/reference/Classes/BarcodeScannerOptions) is only formats  no enableAllPotentialBarcodes, and no setZoomSuggestionOptions either, so the iOS SDK looks generally behind on options rather than this being a deliberate omission. Couldn't find anything from Google explaining why

Lemme know, If I might be missing any direction here.

PR #4193, opened 2026-09-17
docs: add Silicon simulator limitation using barcode-scanner

Minishlink:
Hello,

I stumbled upon this problem yesterday: Google doesn't ship with a arm64 xcframework so you can't use your regular simulators on a Silicon Mac. I feel like it's worth adding this information in the README because it can be a no-go straightaway. https://github.com/margelo/react-native-nitro-zxing was a welcome surprise :) (and it makes much more sense than using ML for this use-case)

Best,
Louis


Minishlink:
Please note that current version of nitro-zxing doesn't seem fully compatible with current version of vision-camera, I needed to patch the podspec to make every header public as a quick workaround

```ruby
s.public_header_files = [
    "cpp/**/*.{h,hpp}",
    "ios/**/*.{h,hpp}",
]
```

klabusta:
Any other suggestion on how to get back the simulators without switching to https://github.com/margelo/react-native-nitro-zxing ?

